PR #45, opened 2016-11-05
[DNM] Use Bintray repos

dkoshkin:


alexbrand:
Should we squash down the commits to a more meaningful commit history?

alexbrand:
Should we close this?

dkoshkin:
Not switching at this point

Issue #55, opened 2016-11-08
Integration test: Timeout (12s) waiting for privilege escalation prompt


dkoshkin:
```
15/16..  Configure Kubernetes DNS                                               
- Running task: verify kubernetes-dns service is running
  ip-10-0-3-225 Timeout (12s) waiting for privilege escalation prompt:          [ERROR]
error installing: error running playbook: error running ansible: exit status 2


• Failure [2451.920 seconds]
Happy Path Installation Tests
/var/snap-ci/src/github.com/apprenda/kismatic/integration/install_test.go:194
  Installing with private Docker registry
  /var/snap-ci/src/github.com/apprenda/kismatic/integration/install_test.go:193
    Using a 1/1/1 CentOS 7 layout
    /var/snap-ci/src/github.com/apprenda/kismatic/integration/install_test.go:192
      Using a custom registry provided by the user
      /var/snap-ci/src/github.com/apprenda/kismatic/integration/install_test.go:191
        should result in a working cluster [It]
        /var/snap-ci/src/github.com/apprenda/kismatic/integration/framework.go:16
```

dkoshkin:
This looks like a reported Anisble issue https://github.com/ansible/ansible/issues/13278. the suggested fix is to increase the `timeout` option in Ansible. This is an SSH timeout not a task timeout.
I don't know if this will fix the issue, it could jut be an AWS thing.

dkoshkin:
On a related note do we use https://github.com/apprenda/kismatic/blob/master/ansible.cfg or https://github.com/apprenda/kismatic/blob/master/ansible/ansible.cfg anywhere, I can't find any mentions.

Looks like the cfg vars are being set in https://github.com/apprenda/kismatic/blob/master/pkg/ansible/runner.go#L132

PR #59, opened 2016-11-08
Increase Ansible timeout for SSH connections

dkoshkin:
Fix for #55 

Can we remove `ansible.cfg` files?

### DIFF
--- a/pkg/ansible/runner.go
+++ b/pkg/ansible/runner.go
@@ -133,6 +133,7 @@ func (r *runner) startPlaybook(playbookFile string, inv Inventory, vars ExtraVar
 	os.Setenv("ANSIBLE_HOST_KEY_CHECKING", "False")
 	os.Setenv("ANSIBLE_CALLBACK_PLUGINS", filepath.Join(r.ansibleDir, "playbooks", "callback"))
 	os.Setenv("ANSIBLE_CALLBACK_WHITELIST", "json_lines")
+	os.Setenv("TIMEOUT", "30")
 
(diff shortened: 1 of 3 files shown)


alexbrand:
Sounds like we need to set the `ANSIBLE_SSH_CONTROL_PATH` env var as well to the value in the ansible.cfg to be able to get rid of the config file.

pkg/ansible/runner.go
```diff
@@ -133,6 +133,8 @@ func (r *runner) startPlaybook(playbookFile string, inv Inventory, vars ExtraVar
 	os.Setenv("ANSIBLE_HOST_KEY_CHECKING", "False")
 	os.Setenv("ANSIBLE_CALLBACK_PLUGINS", filepath.Join(r.ansibleDir, "playbooks", "callback"))
 	os.Setenv("ANSIBLE_CALLBACK_WHITELIST", "json_lines")
+	os.Setenv("TIMEOUT", "30")
+	os.Setenv("ANSIBLE_SSH_ARGS", "-C -o ControlMaster=auto -o ControlPersist=1800s -o ControlPath=/tmp/ssh-%r-%h-%p")
```
dkoshkin: `ANSIBLE_SSH_ARGS` is seems to be enough

dkoshkin:
Ansible command that gets exectued, all variables are still present
```
<192.168.205.13> SSH: EXEC ssh -C -vvv -C -o ControlMaster=auto -o ControlPersist=1800s -o ControlPath=/tmp/ssh-%r-%h-%p -o StrictHostKeyChecking=no
```

PR #61, opened 2016-11-09
Use systemd to manage kismatic-inspector

dkoshkin:
The Kismatic Inspector didn't always start correctly and caused many integration test failures

Minor fixes:
* Print Ansible run command in a new line 
* Always respect `LEAVE_ARTIFACTS`

### DIFF
--- a/ansible/roles/preflight/tasks/main.yaml
+++ b/ansible/roles/preflight/tasks/main.yaml
@@ -15,6 +15,8 @@
     template:
       src: kismatic-inspector.service.j2
       dest: "{{ init_system_dir }}/kismatic-inspector.service"
+    notify:
+      - reload services
 
   - name: start kismatic-inspector service
     service:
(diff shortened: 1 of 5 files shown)


ansible/roles/preflight/tasks/main.yaml
```diff
@@ -3,32 +3,31 @@
...
       mode: 0744
-  - name: start the Kismatic Inspector in server mode
-    command: "{{ bin_dir }}/kismatic-inspector server --node-roles {{ \",\".join(group_names) }} --port 8888 {{ (allow_package_installation == 'true') | ternary('', '-e') }}"
-    async: 300
-    register: start_inspector
-  - name: ensure Kismatic Inspector server is running
-    async_status: jid={{ start_inspector.ansible_job_id }}
-    register: job_result
+  - debug: msg="{{ bin_dir }}/kismatic-inspector server --node-roles {{ \",\".join(group_names) }} --port 8888 {{ (allow_package_installation == 'true') | ternary('', '-e') }}"
+  - name: copy calico-node.service to remote
```
alexbrand: Name is wrong here

dkoshkin:
I though about adding it, you only need that when the service is already running and the file gets changed.
I'll add it because that could happen, ie. when a node changes roles

alexbrand:
Sounds good. Other than that LGTM

dasmb:
LGTM.

PR #63, opened 2016-11-10
Increase retry count for AWS client

alexbrand:
We continue getting the 400 errors when creating instances on AWS. I have increased the number of retries to 5.


dkoshkin:
LGTM, lets get a run in

PR #64, opened 2016-11-10
Actually removed ansible.cfg

dkoshkin:


alexbrand:
Should we close this if we are unable to remove the ansible.cfg file?

dkoshkin:
I added `ANSIBLE_SSH_ARGS` to match the `ansible.cfg` since Im pretty sure that file only gets read during `make itnegration` and not when the kismatic CIL is used

dkoshkin:
Closing as this breaks snap-ci

Issue #56, opened 2016-11-08
Integration test: kubelet fails to start


dkoshkin:
```
13/16..  Install Kubernetes Kubelet on Worker Nodes                             
- Running task: kubelet : verify kubelet is running
  ip-10-0-3-174                                                                 [ERROR]
---- STDOUT ----
● kubelet.service - Kubernetes Kubelet
   Loaded: loaded (/etc/systemd/system/kubelet.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Mon 2016-11-07 05:44:26 UTC; 758ms ago
     Docs: https://github.com/GoogleCloudPlatform/kubernetes
  Process: 10988 ExecStart=/usr/bin/kubelet --allow-privileged=true --api-servers=${MASTER_LIST} --cloud-provider= --cluster-dns=${CLUSTER_DNS} --cluster-domain=cluster.local --configure-cbr0=true --container-runtime=docker --docker=unix:///var/run/docker.sock --hostname-override=${HOSTNAME} --kubeconfig=${KUBECONFIG} --pod-infra-container-image=${INFRA_IMG} --register-schedulable=${REGISTER_SCHEDULABLE} --serialize-image-pulls=false --tls-cert-file=${CERT_FILE} --tls-private-key-file=${KEY_FILE} --network-plugin=cni --network-plugin-dir=${NETWORK_PLUGIN_DIR} --v=2 (code=killed, signal=TERM)
 Main PID: 10988 (code=killed, signal=TERM)

Nov 07 05:44:26 ip-10-0-3-174 kubelet[10988]: I1107 05:44:26.220342   10988 kubelet_node_status.go:356] Recording NodeHasSufficientDisk event message for node ip-10-0-3-174
Nov 07 05:44:26 ip-10-0-3-174 kubelet[10988]: I1107 05:44:26.220361   10988 kubelet_node_status.go:356] Recording NodeHasSufficientMemory event message for node ip-10-0-3-174
Nov 07 05:44:26 ip-10-0-3-174 kubelet[10988]: I1107 05:44:26.220393   10988 kubelet_node_status.go:356] Recording NodeHasNoDiskPressure event message for node ip-10-0-3-174
Nov 07 05:44:26 ip-10-0-3-174 kubelet[10988]: I1107 05:44:26.220405   10988 kubelet_node_status.go:73] Attempting to register node ip-10-0-3-174
[...]

PR #65, opened 2016-11-10
Actually retry when kubelet fails to start

alexbrand:
Addresses #56


dkoshkin:
Should the `delay` be longer than `RetrySec` ie 6 seconds?

I think all services have this problem, can you add the status check to all other systemd verify tasks.

alexbrand:
You were right. The service running checks did not have the until condition set. 

Also updated the interval between systemd restarts to 3 seconds instead of 10.

dkoshkin:
LGTM

PR #77, opened 2016-11-14
Implement kismatic install step command

alexbrand:
The step action can be used to run a single ansible play(book) located in ansible/playbooks. The ansible play will be run exactly the same way as if it were being run as part of `kismatic install apply`.

Sample usage:
```
out ./kismatic install step _apiserver.yaml

Running Task========================================================================
1/1....  Install Kubernetes API Server                                          [OK]

Task completed successfully
```


dkoshkin:
Can you regenerate the docs by first deleting `kismatic-cli-docs` folder and running `make kismatic-cli-docs` 

Would it be beneficial to be able to rerun the tasks based on their printed name, ie
`./kismatic install step Install Kubernetes API Server`. Maybe using [tags](http://docs.ansible.com/ansible/playbooks_tags.html ) or just maintain a map of names to yaml file

dasmb:
```Would it be beneficial to be able to rerun the tasks based on their printed name, ie
./kismatic install step Install Kubernetes API Server. Maybe using tags or just maintain a map of names to yaml file```

Yes. How much extra work is this, though?

Also: we don't want to leak too many Ansible details to a cluster operator, but in this case the PR is intended to make it easier for an operator who understands Ansible to run it the way `kismatic apply` does. I think it would be okay without building a playbook map.

alexbrand:
My main reason for not adding that extra UX optimization was that this is mainly a debug feature.

alexbrand:
Updated docs.

dkoshkin:
LGTM

PR #81, opened 2016-11-14
Don't set cbr0 flag on api-server (deprecated)

### LINKED ISSUE #80: Don't set --configure-cbr0 flag as it has been deprecated
Looking at kubelet logs I found the following message:
```
Nov 14 21:24:54 master01 kubelet[15905]: Flag --configure-cbr0 has been deprecated, 
Will be removed in a future version. Please use kubenet or other network plugins.
```

Given that we are using calico, my assumption is that we can kill that flag altogether.

More info:
https://github.com/kubernetes/kubernetes/pull/31197
https://github.com/kubernetes/kubernetes/issues/31937

alexbrand:
Fixes #80


alexbrand:
Failure was flake. Re-running test.

dkoshkin:
LGTM, this flag is actually removed in 1.5

PR #83, opened 2016-11-14
Remove --api-servers flag from kubelet unit file

### LINKED ISSUE #82: The --api-servers flag on the kubelet is been deprecated
The `--api-servers` flag that we set on the kubelet is deprecated:

```
Nov 14 21:24:54 master01 kubelet[15905]: Flag --api-servers has been deprecated, 
Use --kubeconfig instead. Will be removed in a future version.
```

We are already using `--kubeconfig`, but we should also add the following flag:
```
--require-kubeconfig                             If true the Kubelet will exit if there are configuration errors, 
and will ignore the value of --api-servers in favor of the server defined in the kubeconfig file.
```

alexbrand:
Fixes #82


dkoshkin:
LGTM

dasmb:
LGTM.

PR #96, opened 2016-11-16
Add null checks in aws client

### LINKED ISSUE #94: Panic during test due to unchecked pointer dereferencing
The following error was logged during one of our test runs:
```
Full Stack Trace
        	/var/snap-ci/go/src/runtime/panic.go:458 +0x243
        github.com/apprenda/kismatic/integration/aws.Client.GetNode(0xab85f9, 0x9, 0xabc7c9, 0xf, 0xac7a65, 0x1c, 0xab9a74, 0xb, 0xc420018102, 0x14, ...)
        	/var/snap-ci/src/github.com/apprenda/kismatic/integration/aws/client.go:190 +0x2da
```

The problem is that the AWS API might return null for both private IP and dns name. We need to perform the same null check we perform with the public IP address.

alexbrand:
Fixes #94

### DIFF
--- a/integration/aws/client.go
+++ b/integration/aws/client.go
@@ -181,13 +181,21 @@ func (c Client) GetNode(id string) (*Node, error) {
 	}
 	instance := resp.Reservations[0].Instances[0]
 
+	var privateDNSName string
+	if instance.PrivateDnsName != nil {
+		privateDNSName = *instance.PrivateDnsName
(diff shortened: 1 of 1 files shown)


alexbrand:
Test failure was flake. Re-running.

dkoshkin:
LGTM

PR #98, opened 2016-11-16
Create key/cert for signing service account tokens

### LINKED ISSUE #95: Master nodes are using different keys for validating service accounts
Service account tokens are signed using the private key passed to the controller manager (`--service-account-private-key-file`). Each API server in an HA setup must be able to validate this signature using the corresponding public key. (https://github.com/kubernetes/kubernetes/issues/22351#issuecomment-260824100)

Currently, we are using different keys on each master node for both the controller manager and the API server service account key flags. This means that requests to the kubernetes service (which is backed by all API Servers) using service account tokens will fail intermittently because the API server will not be able to validate them.

We most likely need to generate a separate public/private key pair that shall be used for signing service account tokens, and shall be distributed to all master nodes.

Relevant flags
--service-account-private-key-file (controller manager)
--service-account-key-file (api server)

alexbrand:
Fixes #95


dkoshkin:
LGTM

Issue #99, opened 2016-11-17
Workers are configured to speak to a single manager in a HA setup


DavidHuie:
I've configured Kismatic to configure a multi-master Kubernetes cluster. After inspecting the kubeconfig configuration on each worker, I noticed that the workers are configured to only speak to the apiserver on first manager node I have listed in kismatic-cluster.yml. I did provide `load_balanced_fqdn` and `load_balanced_short_name` in kismatic-cluster.yml for my manangers, so that's what I expect them to talk to.

DavidHuie:
I did some more digging around. kube-proxy on the managers & workers also points to a single manager. On the masters, kube-controller-manager & kube-scheduler point towards a single master as well.

PR #100, opened 2016-11-17
Use apiserver load balancer in all k8s components

DavidHuie:
Addresses #99 

Without using apiserver master load balancer, the cluster isn't
actually high availability since all components have a hard dependency
on the only just the first configured master. This change makes it so
that kubelets, kube-schedulers, kube-proxies, &
kube-controller-managers talk to the apiserver load balancer.

This change has been verified on the Kismatic provisioned cluster that I manage.


dkoshkin:
Thanks for the PR, I've used what you had and added an integration test to cover this issue.

PR #101, opened 2016-11-18
Use load_balanced_fqdn when setting up the cluster

dkoshkin:
* Propogate `load_balanced_fqdn` to Ansible
* Run an integration test with DNS records pointing to API server with AWS Route53

### DIFF
--- a/integration/aws/client.go
+++ b/integration/aws/client.go
@@ -65,31 +65,23 @@ type Credentials struct {
 
 // Client for provisioning machines on AWS
 type Client struct {
-	Config        ClientConfig
-	Credentials   Credentials
-	session       *session.Session
-	ec2Client     *ec2.EC2
-	route53Client *route53.Route53
+	Config      ClientConfig
+	Credentials Credentials
+	session     *session.Session
 }
 
 func (c *Client) getEC2APIClient() (*ec2.EC2, error) {
-	if c.ec2Client == nil {
-		if err := c.prepareSession(); err != nil {
-			return nil, err
-		}
-		c.ec2Client = ec2.New(c.session)
+	if err := c.prepareSession(); err != nil {
+		return nil, err
 	}
-	return c.ec2Client, nil
+	return ec2.New(c.session), nil
 }
 
 func (c *Client) getRoute53APIClient() (*route53.Route53, error) {
-	if c.ec2Client == nil {
-		if err := c.prepareSession(); err != nil {
-			return nil, err
-		}
-		c.route53Client = route53.New(c.session)
+	if err := c.prepareSession(); err != nil {
+		return nil, err
 	}
-	return c.route53Client, nil
+	return route53.New(c.session), nil
 }
 
 func (c *Client) prepareSession() error {
@@ -331,9 +323,9 @@ func modifyHostedZone(record *DNSRecord, action string, hostedZoneID string, api
 	if changeID == nil || *changeID == "" {
 		return fmt.Errorf("Something went wrong, DNS change ID is nil")
 	}
-	// TODO verify propagated
+
 	changeReq := &route53.GetChangeInput{
-		Id: aws.String(*changeID), // Required
+		Id: aws.String(*changeID),
 	}
 
 	err = retryWithBackoff(func() error {
--- a/integration/aws/client.go
+++ b/integration/aws/client.go
@@ -3,12 +3,14 @@ package aws
 import (
 	"fmt"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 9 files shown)


integration/aws/client.go
```diff
@@ -55,28 +65,50 @@ type Credentials struct {
...
 // Client for provisioning machines on AWS
 type Client struct {
-	Config      ClientConfig
-	Credentials Credentials
-	ec2Client   *ec2.EC2
+	Config        ClientConfig
+	Credentials   Credentials
+	session       *session.Session
+	ec2Client     *ec2.EC2
+	route53Client *route53.Route53
```
alexbrand: Do we need to stash both these clients, or is stashing the session enough?

integration/aws/client.go
```diff
@@ -55,28 +65,50 @@ type Credentials struct {
...
+		if err := c.prepareSession(); err != nil {
+			return nil, err
+		}
+		c.ec2Client = ec2.New(c.session)
+	}
+	return c.ec2Client, nil
+}
+
+func (c *Client) getRoute53APIClient() (*route53.Route53, error) {
+	if c.ec2Client == nil {
```
alexbrand: Bad check here

integration/aws/client.go
```diff
@@ -217,15 +249,119 @@ func (c Client) DestroyNodes(nodeIDs []string) error {
...
+
+	resp, err := api.ChangeResourceRecordSets(req)
+	if err != nil {
+		return err
+	}
+	changeID := resp.ChangeInfo.Id
+	if changeID == nil || *changeID == "" {
+		return fmt.Errorf("Something went wrong, DNS change ID is nil")
+	}
+	// TODO verify propagated
```
alexbrand: Are lines 340-346 verifying that it propagated?

integration/install_test.go
```diff
@@ -81,10 +81,10 @@ var _ = Describe("Happy Path Installation Tests", func() {
 					})
 				})
 			})
-			Context("using a 3/2/3 layout with CentOS 7", func() {
+			Context("using a 3/2/3 layout with CentOS 7, with DNS", func() {
```
alexbrand: Could we enhance this test to kill one of the master nodes (shutdown all master services on master[0]), and make sure the cluster survives? If doable, I'd suggest moving it to it's own "Describe"

integration/provision.go
```diff
@@ -22,12 +22,15 @@ const (
...
 	AMIUbuntu1604USEAST = "ami-29f96d3e"
 	AMICentos7UsEast    = "ami-6d1c2007"
+	AWSHostedZoneID     = "Z1LNBHSE28OF08"
 )
 
 type infrastructureProvisioner interface {
 	ProvisionNodes(NodeCount, linuxDistro) (provisionedNodes, error)
 	TerminateNodes(provisionedNodes) error
 	SSHKey() string
+	ConfgiureDNS(masterIPs []string) (*DNSRecord, error)
```
alexbrand: Typo

integration/provision.go
```diff
@@ -295,6 +328,16 @@ func (p packetProvisioner) TerminateNodes(nodes provisionedNodes) error {
 	return nil
 }
 
+func (p packetProvisioner) ConfgiureDNS(masterIPs []string) (*DNSRecord, error) {
+	// TODO
+	return nil, fmt.Errorf("ConfgiureDNS not imeplemented")
```
alexbrand: Typos

integration/provision.go
```diff
@@ -295,6 +328,16 @@ func (p packetProvisioner) TerminateNodes(nodes provisionedNodes) error {
...
+}
+
+func (p packetProvisioner) RemoveDNS(dnsRecord *DNSRecord) error {
+	// TODO
+	return fmt.Errorf("RemoveDNS not imeplemented")
```
alexbrand: Typo

integration/aws/client.go
```diff
@@ -33,6 +36,12 @@ type Node struct {
 	SSHUser        string
 }
 
+// DNSRecord in Router53 on AWS
+type DNSRecord struct {
+	Name   string
+	Values []string
```
dkoshkin: It doesn't have to be, it can be fqdn or hostnames. `value` is consistent with whats its called in DNS

integration/aws/client.go
dkoshkin: Good point, probably safer to recreate the client too, I'll remove it.

integration/aws/client.go
alexbrand: Aaah gotcha! thanks

dkoshkin:
I will squash before merging

PR #106, opened 2016-11-21
Create the calico pool before starting calico on the nodes

alexbrand:
We have seen intermittent issues when trying to access workloads from nodes that are part of the cluster. After investigation, we discovered that the issue is being caused by an open Calico issue: https://github.com/projectcalico/calico-containers/issues/716

The fix proposed by the Calico team is to create the pool before starting calico on the nodes.

### DIFF
--- a/ansible/roles/calico/tasks/main.yaml
+++ b/ansible/roles/calico/tasks/main.yaml
@@ -1,46 +1,7 @@
 ---
-  # install Calico network plugin
-  # setup directories
-  - name: create {{ network_plugin_dir }} directory
-    file:
-      path: "{{ network_plugin_dir }}"
-      state: directory
-  - name: create {{ network_plugin_bin_dir }} directory
-    file:
-      path: "{{ network_plugin_bin_dir }}"
-      state: directory
-
-  # install calicoctl
-  # start calico-node service
-  - name: copy calico-node.service to remote
-    template:
-      src: calico-node.service.j2
-      dest: "{{ kubernetes_service_dir }}/calico-node.service"
-      owner: "{{ kubernetes_owner }}"
-      group: "{{ kubernetes_group }}"
-      mode: "{{ kubernetes_service_mode }}"
-    notify:
-      - reload services
-      - enable calico-node
-      - restart calico-node service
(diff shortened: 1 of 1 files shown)


dkoshkin:
LGTM contingent on passing integration tests

alexbrand:
Investigating CI failures as they are networking related.

matthewdupre:
Please do let me know if this workaround fails so we can appropriately prioritise a real fix.

alexbrand:
The failures we observed in the runs before seem to be related to kuberang. This is evidenced by:
1. The error message says that the busybox container does not exist when trying to exec into it.
```
Accessed Nginx service at 172.17.191.187 from BusyBox                           %!(EXTRA string=error: Internal error occurred: error executing command in container: container not found ("kuberang-busybox")
, string=kuberang-busybox-957457981-0pe7k)[ERROR]
Accessed Nginx service via DNS kuberang-nginx-1479754990254793918 from BusyBox  [ERROR]
Accessed Nginx pod at 172.16.126.71 from BusyBox                                [ERROR]
Accessed Nginx pod at 172.16.126.72 from BusyBox                                [ERROR]
Accessed Nginx pod at 172.16.126.70 from BusyBox                                [ERROR]
```
2. The pods themselves were accessible via the network
```
Accessed Nginx service at 172.17.159.69 from BusyBox                            %!(EXTRA string=wget: can't connect to remote host (172.17.159.69): No route to host
, string=kuberang-busybox-957457981-oc9pa)[ERROR]
Accessed Nginx service via DNS kuberang-nginx-1479746472958610379 from BusyBox  [OK]
Accessed Nginx pod at 172.16.113.132 from BusyBox                               [OK]
Accessed Nginx pod at 172.16.113.131 from BusyBox                               [OK]
```

With this in mind, I think we should be OK merging this in. What does everyone else think?

dkoshkin:
Yeah looks like you are right, I just noticed the same behavior running locally even though the pods worked correctly when I tried manually.

alexbrand:
@dkoshkin Sounds good. I will merge this then. Thanks.

PR #136, opened 2016-11-29
Repo fix

dkoshkin:
* Split up DEB package installation 
* Added integration tests for running previous versions of Kismatic
* Update to k8s v1.4.6 and use new repos with CF
* Updated kismatic links to use CF

### DIFF
--- a/ansible/roles/packages/tasks/main.yaml
+++ b/ansible/roles/packages/tasks/main.yaml
@@ -18,8 +18,8 @@
     uri: url="{{ deb_gpg_key_url }}"
     notify:
       - install apt-transport-https package
-      - add Kismatic deb repository
       - add Kismatic deb key
+      - add Kismatic deb repository
       - apt-get update
     when: ansible_distribution == 'Ubuntu'
     ignore_errors: yes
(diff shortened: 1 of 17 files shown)


ansible/roles/packages/tasks/main.yaml
```diff
@@ -32,26 +32,20 @@
...
-    register: result
-    until: result|success
-    retries: 3
-    delay: 3
+  - name: install yum packages
+    command: /bin/true
+    notify:
+      - install etcd yum package
+      - install kubernetes master yum package
+      - install kubernetes worker yum package
```
alexbrand: Is this essentially a noop? Given that master installs worker as well?

integration/install_released_test.go
```diff
@@ -0,0 +1,48 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+// Test a specific released version of Kismatic
+var _ = Describe("Installing with previous version of Kismatic", func() {
+	var path string
+	BeforeEach(func() {
+		// setup previous version of Kismatic
+		kisReleasedPath, err := DownloadKismaticRelease(previousKismaticVersion)
```
alexbrand: Any reason for defining this constant in another file? Took me a while to find where it was defined.

integration/install_released_test.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+// Test a specific released version of Kismatic
+var _ = Describe("Installing with previous version of Kismatic", func() {
+	var path string
```
alexbrand: could this be `kisReleasedPath` instead of having an extra var (`path`)?

ansible/roles/packages/tasks/main.yaml
dkoshkin: There is the `when` condition, the worker package still needs to be installed on nodes that are only workers

alexbrand: Got it! Thanks

alexbrand:
This looks good to me.

PR #143, opened 2016-12-02
Validate ssh

### LINKED ISSUE #138: Kismatic fails with non-useful error when SSH key has wrong permissions
I have run into various SSH-related issues during the validation phase, and the error message returned by kismatic is just "UNREACHABLE".

This makes me think that the node is unreachable on the network. My first instinct is to ping the machines, and once that passes I am out of ideas.

After some investigation, I found that the issue was related to the SSH key having wrong permissions.

dkoshkin:
Added validation for SSH connection

Closes #138 

```
Reading installation plan file "kismatic-testing.yaml"                          [OK]
- Etcd nodes: error parsing SSH key: Parse SSH key error: ssh: no key found
- Master nodes: error parsing SSH key: Parse SSH key error: ssh: no key found
- Worker nodes: error parsing SSH key: Parse SSH key error: ssh: no key found
validation error prevents installation from proceeding
```
```
- Worker nodes: error SSH into node: node3, dial tcp 192.168.205.120:22: i/o timeout
validation error prevents installation from proceeding
```
```
- Etcd nodes: error SSH into node: node1, ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
- Master nodes: error SSH into node: node2, ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
- Worker nodes: error SSH into node: node3, ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
validation error prevents installation from proceeding
```
```
- Etcd nodes: error parsing SSH key: Encrypted SSH key is not permitted
- Master nodes: error parsing SSH key: Encrypted SSH key is not permitted
- Worker nodes: error parsing SSH key: Encrypted SSH key is not permitted
validation error prevents installation from proceeding
```


pkg/install/validate.go
```diff
@@ -149,6 +166,48 @@ func (s *SSHConfig) validate() (bool, []error) {
...
+	}
+
+	return v.valid()
+}
+
+func verifySSH(node *Node, sshConfig *SSHConfig, sshClientConfig *ssh.ClientConfig) error {
+	server := node.IP + ":" + strconv.Itoa(sshConfig.Port)
+	conn, err := net.DialTimeout("tcp", server, time.Second*5)
+	if err != nil {
+		return fmt.Errorf("cannot ping node")
```
dasmb: "Cannot connect to node %v at port %v"; want to avoid confusion with ICMP ping and want to be clear which IP is in use (having the potential for multiple IP addresses in play per node, it's valuable to have error messages call out exactly what failed)

pkg/install/validate.go
```diff
@@ -149,6 +166,48 @@ func (s *SSHConfig) validate() (bool, []error) {
...
+			Auth: []ssh.AuthMethod{
+				auth,
+			},
+		}
+		var wg sync.WaitGroup
+		wg.Add(len(s.nodes))
+		for _, node := range s.nodes {
+			go func(node Node) {
+				if err := verifySSH(&node, s.sshConfig, sshClientConfig); err != nil {
+					v.addError(fmt.Errorf("error SSH into node: %s, %v", node.Host, err))
```
alexbrand: This is problematic as we are performing concurrent writes to the underlying slice. We'll need to use a channel if we want to do this concurrently.

dkoshkin:
Please take a look at the newest changes

pkg/install/validate.go
```diff
@@ -149,6 +191,48 @@ func (s *SSHConfig) validate() (bool, []error) {
...
+
+func verifySSH(node *Node, sshConfig *SSHConfig, sshClientConfig *ssh.ClientConfig) error {
+	server := node.IP + ":" + strconv.Itoa(sshConfig.Port)
+	conn, err := net.DialTimeout("tcp", server, time.Second*5)
+	if err != nil {
+		return fmt.Errorf("cannot connect to node: %v", server)
+	}
+
+	// Try to connect with a timeout
+	_, _, _, err = ssh.NewClientConn(conn, server, sshClientConfig)
```
alexbrand: Do we need to close the connection?

pkg/install/validate.go
```diff
@@ -29,6 +35,37 @@ func ValidateNode(node *Node) (bool, []error) {
 	return v.valid()
 }
 
+// ValidatePlanSSHConnection tries to establish SSH connections to all nodes in the cluster
+func ValidatePlanSSHConnection(p *Plan) (bool, []error) {
+	v := newValidator()
+
+	ok, errs := ValidateSSHConnection(&SSHConnection{&p.Cluster.SSH, p.Etcd.Nodes}, "Etcd nodes")
```
alexbrand: Can we simplify lines 42-55 to:
```
v.validateWithErrPrefix("Etcd nodes", &SSHConnection{&p.Cluster.SSH, p.Etcd.Nodes})
v.validateWithErrPrefix("Master nodes", &SSHConnection{&p.Cluster.SSH, p.Master.Nodes})
v.validateWithErrPrefix("Worker nodes", &SSHConnection{&p.Cluster.SSH, p.Worker.Nodes})
``` 

Or am I missing something? Also seems like we would have needed `if !ok`?

pkg/install/validate.go
```diff
@@ -149,6 +191,48 @@ func (s *SSHConfig) validate() (bool, []error) {
...
+			Auth: []ssh.AuthMethod{
+				auth,
+			},
+		}
+		var wg sync.WaitGroup
+		wg.Add(len(s.Nodes))
+		for _, node := range s.Nodes {
+			go func(node Node) {
+				if err := verifySSH(&node, s.SSHConfig, sshClientConfig); err != nil {
+					v.addError(fmt.Errorf("error SSH into node: %s, %v", node.InternalIP, err))
```
alexbrand: This is problematic because we are doing concurrent updates to the underlying slice. We'll need a channel here if we want to concurrently verify SSH connectivity.

pkg/install/validate.go
dkoshkin: Yeah I think you are right, this is messier than jut doing above, Ill change it.

pkg/install/validate.go
dkoshkin: Yep

pkg/install/validate.go
```diff
@@ -149,6 +180,67 @@ func (s *SSHConfig) validate() (bool, []error) {
...
+					errQueue <- nil
+				}
+			}(node)
+		}
+		go func() {
+			for err := range errQueue {
+				if err != nil {
+					v.addError(err)
+				}
+				wg.Done()
```
alexbrand: I believe this means that we'll increment the waitgroup twice for each node, which will unblock the main goroutine prematurely.

I suggest having this extra goroutine waiting for the waitgroup, and closing the errQueue channel once it gets the signal. With this change, reading the errQueue channel can be moved to the main goroutine.

dkoshkin: Yep thats exactly what is happening.

pkg/util/ssh.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+// GetUnencryptedPublicKeyAuth parses SSH private key and returns PublicKeys AuthMethod
+func GetUnencryptedPublicKeyAuth(file string) (ssh.AuthMethod, error) {
+	buffer, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil, err
+	}
+
+	isEcnrypted, err := isEncrypted(buffer)
```
alexbrand: typo

pkg/util/ssh.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Parse SHH key error")
```
alexbrand: typo: SHH

pkg/util/ssh.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	if isEcnrypted {
+		return nil, fmt.Errorf("Encrypted SSH key is not permitted")
+	}
+
+	signer, err := ssh.ParsePrivateKey(buffer)
+	if err != nil {
+		return nil, fmt.Errorf("Parse SHH key error: %v", err)
```
alexbrand: typo: SHH

pkg/util/ssh.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	return ssh.PublicKeys(signer), nil
+}
+
+func isEncrypted(buffer []byte) (bool, error) {
+	// There is no error, just a nil block
+	block, _ := pem.Decode(buffer)
```
alexbrand: What's the second value returned here? If it's an error, seems like we could use it instead of block == nil?

pkg/util/ssh.go
```diff
@@ -0,0 +1,45 @@
...
+	// File cannot be decoded, maybe it's some unecpected format
+	if block == nil {
+		return false, fmt.Errorf("Parse SHH key error")
```
alexbrand: typo: SHH

pkg/install/validate.go
```diff
@@ -149,6 +180,67 @@ func (s *SSHConfig) validate() (bool, []error) {
...
+	}
+
+	return v.valid()
+}
+
+func verifySSH(node *Node, sshConfig *SSHConfig, sshClientConfig *ssh.ClientConfig) error {
+	server := node.IP + ":" + strconv.Itoa(sshConfig.Port)
+	conn, err := net.DialTimeout("tcp", server, time.Second*5)
+	if err != nil {
+		return fmt.Errorf("cannot connect to node: %v", server)
```
alexbrand: I suspect it would be valuable to include the `err` in the error message, as it might contain useful failure information.

pkg/install/validate.go
```diff
@@ -149,6 +180,67 @@ func (s *SSHConfig) validate() (bool, []error) {
...
+				// Need to send something the buffered channel
+				if sshErr != nil {
+					errQueue <- fmt.Errorf("error SSH into node: %s, %v", node.InternalIP, sshErr)
+				} else {
+					errQueue <- nil
+				}
+			}(node)
+		}
+		// Wait for all nodes to complete, then close channel
+		wg.Wait()
```
alexbrand: This should work. If you ever want to process the results as they come in, this is what I had originally suggested: 
https://play.golang.org/p/M4Ge3A6ECJ

alexbrand:
LGTM!

PR #146, opened 2016-12-02
Add passwordless sudo user requirement to PROVISION.md

jaycoon:
Creating a user with passwordless sudo privileges is a requirement for Kismatic to install packages across all nodes, so this needs to be part of the provisioning step. I've also added the commands that should be used to be clear, though we could remove these.


dasmb:
LGTM Thanks Jay!

Issue #142, opened 2016-12-02
allow_package_installation validate fails when installed with false then true


dkoshkin:
version: **Kismatic v1.0.3**

Running `validate` with `allow_package_installation=true` then `allow_package_installation=false` still returns 

```
Validating==========================================================================
Reading installation plan file "kismatic-cluster.yaml"                          [OK]
Validating installation plan file                                               [OK]
1/3....  Configure Cluster Prerequisites                                        [OK]
2/3....  Update Hosts File                                                      [OK]
3/3....  Run Cluster Pre-Flight Checks
=> The following checks failed on "node3":
   - Package Available: kismatic-kubernetes-node 1.4.6_1-1: kismatic-kubernetes-node 1.4.6_1-1 is not installed and isn't available in known package repositories


=> The following checks failed on "node2":
   - Package Available: kismatic-kubernetes-master 1.4.6_1-1: kismatic-kubernetes-master 1.4.6_1-1 is not installed and isn't available in known package repositories


=> The following checks failed on "node1":
   - Package Available: kismatic-etcd 1.4.6_1-1: kismatic-etcd 1.4.6_1-1 is not installed and isn't available in known package repositories
```

dkoshkin:
Closed with #149

Issue #105, opened 2016-11-21
Install validation terminates with error


egulacti:
I am using Ubuntu Xenial and Kismatic repo is added to the machine. I have package installation  enabled in the plan file. When I execute **kismatic install validate** I receive the following result. 

```
Reading installation plan file "kismatic-cluster.yaml"                          [OK]
Validating installation plan file                                               [OK]
1/2....  Configure Cluster Prerequisites                                        [OK]
2/2....  Run Cluster Pre-Flight Checks
=> The following checks failed on "192.168.20.222":
    Package Available: kismatic-etcd 1.4.5-1: kismatic-etcd 1.4.5-1 is not installed but is available in a package repository

=> The following checks failed on "192.168.20.223":
    Package Available: kismatic-kubernetes-node 1.4.5-1: kismatic-kubernetes-node 1.4.5-1 is not installed but is available in a package repository

=> The following checks failed on "192.168.20.221":
    Package Available: kismatic-kubernetes-master 1.4.5-1: kismatic-kubernetes-master 1.4.5-1 is not installed but is available in a package repository

error running preflight: error running playbook: error running ansible: exit status 2
```

dkoshkin:
Do you ming sharing your plan file (removed password and any other sensitive info) thanks.

egulacti:
My plan file is attached.
[kismatic-cluster-yaml.txt](https://github.com/apprenda/kismatic/files/605632/kismatic-cluster-yaml.txt)

alexbrand:
I suspect that what you ran into is that you enabled package installation after an initial execution of `kismatic install validate`. We have a bug where we are not shutting down the inspector after it runs, which would have resulted in the inspector thinking package installation was disabled.

The `always` block in the inspector play is not being executed due to the `any_errors_fatal: true` directive in `_preflight.yaml`.

@dkoshkin Do you recall why this was added?

alexbrand:
It's unclear if this is a bug in ansible. Opened https://github.com/ansible/ansible/issues/18602 to get some feedback from ansible team.

alexbrand:
Ansible team is considering this a bug.

dkoshkin:
Closed with #149

PR #149, opened 2016-12-05
Stop kismatic-inspector even on failure

dkoshkin:


alexbrand:
Does the ansible version that we use run the "rescue" block in case of failure? The ansible team mentioned that they had a bug that was fixed in 2.2.0

dkoshkin:
I tested with the version we have and it worked

alexbrand:
Cool. Sounds like we can merge this, and close #142 and #105.

dkoshkin:
Fixes #142 and Fixes #105

PR #150, opened 2016-12-05
Use IP instead of hostname for etcd bind address

### LINKED ISSUE #145: Etcd fails to start if the machine's hostname is mapped to 127.0.0.1 in the hosts file
Etcd fails to start with a misleading error when 127.0.0.1 is mapped to the hostname in the hosts file:

```
Dec 02 11:34:27 pbcoord01 etcd[31261]: listen tcp 127.0.0.1:2379: bind: address already in use
```

Potential resolutions:
* Use the internal ip provided in the plan file, instead of the node's hostname
* Add validation around this

alexbrand:
Specify the IP address that etcd should bind, instead of allowing it to choose one given the machine's hostname.

Fixes #145

### DIFF
--- a/ansible/group_vars/all.yaml
+++ b/ansible/group_vars/all.yaml
@@ -47,8 +47,7 @@ etcd_service_owner: root
 etcd_service_group: root
 etcd_service_mode: 0664
 # etcd cluster setup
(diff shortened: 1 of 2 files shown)


dkoshkin:
LGTM

PR #154, opened 2016-12-07
Nginx Ingress

dkoshkin:
See [proposal](https://github.com/apprenda/kismatic/blob/ingress/proposals/INGRESS.md).


dkoshkin:
Currently if `setup_internal: true` is used, a docker registry running on the first master will be configured with the port `443`. 

In a *minikube* scenario or where the `ingress` node is also the first `master` there will be a port collision.
A *minikube* with both an ingress and registry setup is a valid cluster. 

I propose to use a different port for the docker registry `8443` instead of 443.

With this change there will be some minor code and documentation changes. Also using the registry will now require to include the port, ie `docker pull fqdn:8443/nginx`

docs/INGRESS.md
```diff
@@ -0,0 +1,129 @@
...
+  nodes:
+  - host: node4.somehost.com
+    ip: 8.8.8.4
+    internalip: 8.8.8.4
+  - host: node1.somehost.com
+    ip: 8.8.8.1
+    internalip: 8.8.8.1
+```
+
+To `ingress` node types an optional section
```
alexbrand: Typo?

docs/INGRESS.md
```diff
@@ -0,0 +1,129 @@
...
+When an ingress section is not provided, the ingress controller will NOT be setup  
+`ingress` can have 1 or more nodes, these nodes can be unique from the other roles or can be shared
+* On an `ingress` node the kubelet will be **unschedulable**, ie. `node4.somehost.com` from the example
+* If the node is only shared with `etcd` or/and `master ` the kubelet will be **unschedulable**
+* If the `ingress` node is also a `worker` the kubelet will be **schedulable**, ie. `node1.somehost.com` from the example
+
+### Example Ingress Resources
+Assumptions:
+* at least 1 `ingress` node was provided when setting up the cluster
+* a service named `echoserver` with `port: 80` is running in the cluster and the `deafult` namespace
```
alexbrand: typo: deafult

docs/INGRESS.md
```diff
@@ -0,0 +1,129 @@
...
+  - host: mydomain.com
+    http:
+      paths:
+      - path: /echoserver
+        backend:
+          serviceName: echoserver
+          servicePort: 80
+```
+
+After running the above, your service will be accessible vi `http://mydomain.com/echoserver` and `https://mydomain.com/echoserver`
```
alexbrand: typo: vi

docs/INGRESS.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Out of Scope
+* Integrating with any cloud provider for Load Balance functionality - it is up to the user of the cluster to configure their Load Balancers with the ingress nodes
+* Automatic HTTPs cert generation
```
alexbrand: Typo: HTTPs

integration/install.go
```diff
@@ -136,6 +141,86 @@ func verifyMasterNodeFailure(nodes provisionedNodes, provisioner infrastructureP
...
+	url = "https://" + node.PublicIP + "/echo-tls"
+	if err := retry.WithBackoff(func() error { return ingressRequest(url) }, 7); err != nil {
+		return err
+	}
+	return nil
+}
+
+func ingressRequest(url string) error {
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
alexbrand: How much work would it be to set this up to trust the generated cert for the TLS endpoint? It'd be great to fully test the HTTPS endpoint, including the correct configuration of certs.

integration/install.go
```diff
@@ -136,6 +141,86 @@ func verifyMasterNodeFailure(nodes provisionedNodes, provisioner infrastructureP
...
+	if err != nil {
+		fmt.Printf("Could not reach ingress via %s, %v", url, err)
+		return fmt.Errorf("Could not reach ingress via %s, %v", url, err)
+	}
+	if resp.StatusCode != 200 {
+		fmt.Printf("Ingress status code is not 200, got %d vi %s", resp.StatusCode, url)
+		return fmt.Errorf("Ingress status code is not 200, got %d vi %s", resp.StatusCode, url)
+	}
+
+	fmt.Println("GOT TO THE END")
```
alexbrand: Can be removed

pkg/install/execute.go
```diff
@@ -212,6 +212,12 @@ func (ae *ansibleExecutor) buildInstallExtraVars(p *Plan, tlsDirectory string) (
 			ev[fmt.Sprintf("force_%s_restart", s)] = strconv.FormatBool(true)
 		}
 	}
+	// Need to pass variable to Ansible to skip cleanly
+	if p.Ingress.Nodes != nil && len(p.Ingress.Nodes) > 0 {
+		ev["confgiure_ingress"] = "true"
```
alexbrand: Typo

pkg/install/validate.go
```diff
@@ -42,6 +42,9 @@ func ValidatePlanSSHConnection(p *Plan) (bool, []error) {
 	v.validateWithErrPrefix("Etcd nodes", &SSHConnection{&p.Cluster.SSH, p.Etcd.Nodes})
 	v.validateWithErrPrefix("Master nodes", &SSHConnection{&p.Cluster.SSH, p.Master.Nodes})
 	v.validateWithErrPrefix("Worker nodes", &SSHConnection{&p.Cluster.SSH, p.Worker.Nodes})
+	if p.Ingress.Nodes != nil || p.Ingress.ExpectedCount > 0 {
```
alexbrand: From this check, I assume that if there are no ingress nodes defined, but expected count is greater than 0, `Ingress.Nodes` will be nil? Does the validation handle this?

I am actually not sure what this check is doing.

alexbrand: I looked into the NodeGroup validation code. I wonder if instead we should modify the validation code to allow for an "empty" node group. That is, allow a NodeGroup with expectedCount == 0 AND no nodes in the group.

docs/INGRESS.md
dkoshkin: Whats the correct form `HTTPs` or `HTTPS`?

pkg/install/validate.go
dkoshkin: Yeah this should be `p.Ingress.Nodes != nil && p.Ingress.ExpectedCount > 0` as this is an optional section.

integration/install.go
dkoshkin: Yeah Im going to look into this, I actually thought I was getting this because the cert is self-signed but reading the docs Im not sure anymore.

dkoshkin: I couldn't get this to work, kept getting `cannot validate certificate for because it doesn't contain any IP SANs ` even though I could see the IPs. I think this is caused by the `Host` header which is required for the ingress to work. I reverted back.

docs/INGRESS.md
```diff
@@ -81,10 +82,14 @@ spec:
           servicePort: 80
 
 ```
-Please note the `ingress.kubernetes.io/ssl-redirect: "false"` annotation,
+Please note the `annotations` optopns:
```
alexbrand: Typo

integration/install.go
```diff
@@ -136,6 +141,86 @@ func verifyMasterNodeFailure(nodes provisionedNodes, provisioner infrastructureP
...
+
+func newTestIngressCert() error {
+	err := exec.Command("openssl", "req", "-x509", "-nodes", "-days", "365", "-newkey", "rsa:2048", "-keyout", "tls.key", "-out", "tls.crt", "-subj", "/CN=kismaticintegration.com").Run()
+	return err
+}
+
+func verifyIngressPoint(node NodeDeets) error {
+	// HTTP ingress
+	url := "http://" + node.PublicIP + "/echo"
+	if err := retry.WithBackoff(func() error { return ingressRequest(url) }, 10); err != nil {
```
alexbrand: This retry count is too high (~17 min if my math is right)

integration/test-resources/ingress.yaml
```diff
@@ -0,0 +1,98 @@
...
+    kubernetes.io/ingress.class: "nginx"
+    ingress.kubernetes.io/rewrite-target: /
+spec:
+  tls:
+  - hosts:
+    - kismaticintegration.com
+    secretName: kismaticintegration-tls
+  rules:
+  - host: kismaticintegration.com
+    http:
```
alexbrand: Should this be https?

integration/install.go
dkoshkin: Is it 17 or 2^9 since the break is before the sleep?
Also the ingress do take some time to start.

alexbrand: Ah, yeah. It's 8 minutes. Still seems high, or does it really take that long?

dkoshkin: Ehh I dont think 4 will be enough.

integration/test-resources/ingress.yaml
dkoshkin: Looking at their spec its only `http`, when the `tls` section is included HTTPS in inferred.

Issue #153, opened 2016-12-06
TLS keys should be verified to contain the expected IP(s) from the plan file


dkoshkin:
When certs are not regenerated and the hostname and IPs in certs do not contain the values from the plan, there should at least be a warning possibly a failure.

dasmb:
Merged and Coming in v1.2.

PR #160, opened 2016-12-14
Validate generated certificates to contain the correct CN and SANs entries

dkoshkin:
Addresses #153


pkg/tls/cert.go
```diff
@@ -111,6 +111,48 @@ func CertKeyPairExists(name, dir string) (bool, error) {
 	return true, nil
 }
 
+// CertKeyPairExistsAndValid returns true if a key and matching certificate exist.
+// Matching is defined as having the expected file names and IP Addresse(s)
+func CertKeyPairExistsAndValid(CN string, SANs []string, name, dir string) (valid bool, warn []error, err error) {
```
alexbrand: Was there a need for making this function do two things, instead of just validating a given certificate? It feels like it's doing too much.

pkg/tls/cert_test.go
```diff
@@ -130,3 +134,210 @@ func TestGenerateNewCertificate(t *testing.T) {
...
+			expectedSANs: []string{"10.0.0.1", "192.168.99.101"},
+			certCN:       "node1",
+			certSANs:     []string{"10.0.0.1", "192.168.99.101"},
+			valid:        false,
+		},
+		{
+			expectedSANs: []string{"10.0.0.1", "192.168.99.101"},
+			certCN:       "node1",
+			certSANs:     []string{"10.0.0.1", "192.168.99.101"},
+			valid:        false,
```
alexbrand: Why is this case invalid?

dkoshkin: The CNs don't match

pkg/tls/cert.go
dkoshkin: I think you might be right, originally I did it to replace the existing `CertKeyPairExists` but the signature changed  anyways.
I will split it up.

pkg/install/pki.go
```diff
@@ -163,7 +154,28 @@ func (lp *LocalPKI) GenerateNodeCertificate(plan *Plan, node Node, ca *tls.CA) e
...
 		}
 	}
-	key, cert, err := generateCert(node.Host, plan, nodeSANs, ca)
+
+	// Don't generate if the key pair exists and valid
+	exists, err := tls.CertKeyPairExists(node.Host, lp.GeneratedCertsDirectory)
+	if err != nil {
+		return fmt.Errorf("error verifying if certificates exists for node %q: %v", node.Host, err)
+	}
+	valid, warn, err := tls.CertValid(CN, nodeSANs, node.Host, lp.GeneratedCertsDirectory)
```
alexbrand: Should we skip validation if it does not exist?

pkg/tls/cert.go
```diff
@@ -111,6 +111,51 @@ func CertKeyPairExists(name, dir string) (bool, error) {
 	return true, nil
 }
 
+// CertValid returns true if a matching certificate exist
+// Matching is defined as having the expected CN and SANs
+// A certificate with a wrong CN or that doesn't contain the expected SANs, will return a warning
+func CertValid(CN string, SANs []string, name, dir string) (valid bool, warn []error, err error) {
```
alexbrand: I don't see a case where we return multiple warnings. It seems like we should perform all the validation and then return all the warnings? Or just return a single warning.

dkoshkin: Golang complains about error not being the last return value. I've seen this pattern in kubeadm and k8s code.

alexbrand: Got it. So the "return a single warning" option is not viable. I think doing complete validation over failing fast would result in a better UX. For the case where there are multiple issues with the certs, the user will have to run validation multiple times to discover each one.

dkoshkin: I like it

pkg/install/pki.go
```diff
@@ -163,7 +154,30 @@ func (lp *LocalPKI) GenerateNodeCertificate(plan *Plan, node Node, ca *tls.CA) e
...
+		if errValid != nil {
+			return fmt.Errorf("error verifying certificates for node %q: %v", node.Host, errValid)
+		}
+		if valid {
+			util.PrettyPrintOk(lp.Log, "Found valid key and certificate for node %q", node.Host)
+			return nil
+		}
+		// cert exists but is not valid
+		if warn != nil && len(warn) > 0 {
+			util.PrettyPrintOk(lp.Log, "Found key and certificate for node %q but it is not valid: %v", node.Host, warn)
```
alexbrand: Is the intent here to overwrite the invalid certs? That might be problematic if a user provides certs and we blow them away

dkoshkin: Yeah it was, do you think it would be a better workflow to fail and print the errors?

dkoshkin: Hmm this is making me think that the validation should also happen on `validate` and here it should just exit with an error.

alexbrand: Yes. I wonder the same.

It just seems bad to overwrite certificates that might have been provided by the user.

dkoshkin: I don't think validation at `validation` is the right place, it gets complicated with add-worker workflow. 
There is also code that would need to be duplicated.
I propose printing the warn messages and exiting in this part of the code, below is what it would look like:

```
Generating cluster Certificate Authority                                        [OK]
Generating certificates for host "etcd"                                         [OK]
Generating certificates for host "master"                                       [OK]
Generating certificates for host "worker"                                       [OK]
Generating certificates for service accounts                                    [OK]
Generating certificates for user "admin"                                        [OK]
Regenerating Certs
Found valid key and certificate for node "etcd"                                 [OK]
Found valid key and certificate for node "master"                               [OK]
Found valid key and certificate for node "worker"                               [OK]
Found key and certificate for service accounts                                  [OK]
Found key and certificate for user "user" but it is not valid                   [ERROR]
- error validating CN, expected "user", instead got "admin"
- error validating SANs, expected:
	[user]
  instead got:
	[admin]
```
```
Generating cluster Certificate Authority                                        [OK]
Generating certificates for host "etcd"                                         [OK]
Generating certificates for host "master"                                       [OK]
Generating certificates for host "worker"                                       [OK]
Generating certificates for service accounts                                    [OK]
Generating certificates for user "admin"                                        [OK]
Regenerating Certs
Found valid key and certificate for node "etcd"                                 [OK]
Found valid key and certificate for node "master"                               [OK]
Found valid key and certificate for node "worker"                               [OK]
Found key and certificate for service account but it is not valid               [ERROR]
- error parsing cert service-account: {"code":1002,"message":"Failed to decode certificate"}
```
```
Generating cluster Certificate Authority                                        [OK]
Generating certificates for host "etcd"                                         [OK]
Generating certificates for host "master"                                       [OK]
Generating certificates for host "worker"                                       [OK]
Generating certificates for docker registry                                     [OK]
Generating certificates for service accounts                                    [OK]
Generating certificates for user "admin"                                        [OK]
Found key and certificate for docker registry but it is not valid               [ERROR]
- error validating SANs, expected:
	[master 11.12.13.14 22.33.44.55]
  instead got:
	[99.99.99.99 88.88.88.88 master]
```

We already do something similar in "add-worker", where we validate and exit.

dkoshkin:
I don't think validation at `validation` is the right place, it gets complicated with add-worker workflow. 
There is also code that would need to be duplicated.
I propose printing the warn messages and exiting in this part of the code, below is what it would look like:

```
Generating cluster Certificate Authority                                        [OK]
Generating certificates for host "etcd"                                         [OK]
Generating certificates for host "master"                                       [OK]
Generating certificates for host "worker"                                       [OK]
Generating certificates for service accounts                                    [OK]
Generating certificates for user "admin"                                        [OK]
Regenerating Certs
Found valid key and certificate for node "etcd"                                 [OK]
Found valid key and certificate for node "master"                               [OK]
Found valid key and certificate for node "worker"                               [OK]
Found key and certificate for service accounts                                  [OK]
Found key and certificate for user "user" but it is not valid                   [ERROR]
- error validating CN, expected "user", instead got "admin"
- error validating SANs, expected:
	[user]
  instead got:
	[admin]
```
```
Generating cluster Certificate Authority                                        [OK]
Generating certificates for host "etcd"                                         [OK]
Generating certificates for host "master"                                       [OK]
Generating certificates for host "worker"                                       [OK]
Generating certificates for service accounts                                    [OK]
Generating certificates for user "admin"                                        [OK]
Regenerating Certs
Found valid key and certificate for node "etcd"                                 [OK]
Found valid key and certificate for node "master"                               [OK]
Found valid key and certificate for node "worker"                               [OK]
Found key and certificate for service account but it is not valid               [ERROR]
- error parsing cert service-account: {"code":1002,"message":"Failed to decode certificate"}
```
```
Generating cluster Certificate Authority                                        [OK]
Generating certificates for host "etcd"                                         [OK]
Generating certificates for host "master"                                       [OK]
Generating certificates for host "worker"                                       [OK]
Generating certificates for docker registry                                     [OK]
Generating certificates for service accounts                                    [OK]
Generating certificates for user "admin"                                        [OK]
Found key and certificate for docker registry but it is not valid               [ERROR]
- error validating SANs, expected:
	[master 11.12.13.14 22.33.44.55]
  instead got:
	[99.99.99.99 88.88.88.88 master]
```

We already do something similar in "add-worker", where we validate and exit.

alexbrand:
Is the above output from test runs? Just want to clarify

dkoshkin:
Yes it is

PR #161, opened 2016-12-15
Use kubernetes service and service accounts for kubedns, instead of kubeconfig file

### LINKED ISSUE #123: Fix or Improve behavior of LB master endpoints when no DNS is in play
The entries for are wrong:
The newbie of kismatic is lead a merry dance....
```
update_hosts_files: true            # When true, the installer will add entries for all nodes to other nodes' hosts files. Use when you don't have access to DNS.
load_balanced_fqdn: "master-1"                 # If using a load balanced master fqdn enter it here; otherwise use the fqdn of any master node.
load_balanced_short_name: "master-1"           # If using a load balanced master enter its short name here; otherwise use the short name of any master node.
```
The above would suggest you can run without DNS, and use names resolved by /etc/hosts. BUT YOU CAN NOT! 

Why....
kube-dns is setup NOT to mount /etc/hosts into the container and so never starts failing on looking up with dns "master-1" (in this example) which 8.8.8.8 (in my case) can never resolve correctly.

TO FIX I have done the following:
load_balanced_fqdn: "10.10.0.21"    

ie setting it to the actual IP of my first apimaster node. I assume you could put in the IP address of the Load-balancer too ? however would the certs be created for it?

alexbrand:
In order to better support our hosts file modification feature, we make `kubedns` use the kubernetes service and service accounts for accessing the API server.

Fixes #123

### DIFF
--- a/ansible/roles/addon-kubernetes-dns/templates/kubernetes-dns.yaml
+++ b/ansible/roles/addon-kubernetes-dns/templates/kubernetes-dns.yaml
@@ -82,7 +82,6 @@ spec:
         # command = "/kube-dns"
         - --domain=cluster.local
         - --dns-port=10053
-        - --kubecfg-file={{ kubernetes_kubeconfig_path }}
         volumeMounts:
         - mountPath: {{ kubernetes_certificates_ca_path }}
           name: "ssl-certs"


ansible/roles/addon-kubernetes-dns/templates/kubernetes-dns.yaml
```diff
@@ -82,7 +82,6 @@ spec:
         # command = "/kube-dns"
         - --domain=cluster.local
         - --dns-port=10053
-        - --kubecfg-file={{ kubernetes_kubeconfig_path }}
```
dkoshkin: I guess it just defaults to using to service account if their is no kubeconfig?

alexbrand: That's correct.

dkoshkin:
LGTM pending integration tests

alexbrand:
Looks like we might be getting a legitimate failure now that master is fixed. I'll investigate this tomorrow.

dkoshkin:
You should merge master, that has the SDK vendoring and the integration test fix that specifies the package versions to use when installing manually.

alexbrand:
It gets merged automatically on Snap AFAIK:

```
$ git clone git@localhost:apprenda/kismatic .
Cloning into '.'...
Warning: Permanently added 'localhost' (ECDSA) to the list of known hosts.
Command git clone git@localhost:apprenda/kismatic . exited successfully with status 0.
Took 0.51775 seconds.

$ git checkout master
Already on 'master'
Your branch is up-to-date with 'origin/master'.
Command git checkout master exited successfully with status 0.
Took 0.02632 seconds.

$ git fetch origin pull/161/head:pull-161
Warning: Permanently added 'localhost' (ECDSA) to the list of known hosts.
From localhost:apprenda/kismatic
 * [new ref]         refs/pull/161/head -> pull-161
Command git fetch origin pull/161/head:pull-161 exited successfully with status 0.
Took 0.17403 seconds.

$ git merge 9a78598dced4e54bef8cf67f9f5ecf1ed10a5ae0
Merge made by the 'recursive' strategy.
 ansible/roles/addon-kubernetes-dns/templates/kubernetes-dns.yaml | 1 -
 1 file changed, 1 deletion(-)
Command git merge 9a78598dced4e54bef8cf67f9f5ecf1ed10a5ae0 exited successfully with status 0.
Took 0.04012 seconds.
```

Have you seen issues that would suggest the contrary?

dkoshkin:
Woah I was not aware that it gets merged in, thats actually pretty cool.

PR #162, opened 2016-12-15
1.5.1

dkoshkin:
- Updated to kubernetes v1.5.1
- Using test repos (TODO change to prod)
- Use a specific kismatic version in integration tests


ansible/group_vars/all.yaml
```diff
@@ -1,19 +1,22 @@
...
 calico_docker_version: v0.22.0
 calico_kube_policy_controller_version: v0.4.0
-kubedns_version: "1.8"
+kubedns_version: "1.9"
 kube_dnsmasq_version: "1.4"
 exechealthz_version: "1.2"
-kubernetes_dashboard_version: v1.4.2
+kubernetes_dashboard_version: v1.5.0
+defaultbackend_version: "1.0"
+nginx_ingress_controller_version: "0.8.3"
```
alexbrand: Is this branch somehow related to the ingress branch?

ansible/group_vars/etcd-networking.yaml
```diff
@@ -8,8 +8,8 @@ etcd_certificates_cert_file: "{{ etcd_install_dir }}/{{ etcd_certificates_cert_f
 etcd_certificates_key_file: "{{ etcd_install_dir }}/{{ etcd_certificates_key_file_name }}"
 # etcd-install
 etcd_install_dir: /etc/etcd_networking
-etcd_install_bin_name: etcd2
-etcdctl_install_bin_name: etcdctl2
+etcd_install_bin_name: etcd_networking
```
alexbrand: Curious about why renaming these was necessary.

ansible/group_vars/all.yaml
dkoshkin: The packages have these images in them now, these probably could have waited for the ingress PR since they aren't used but thats going to get merged right after.

alexbrand: Got it!

ansible/group_vars/etcd-networking.yaml
dkoshkin: Originally I modified to use etcd3 for both networking and k8s but then reverted the calico upgrade.

My plan is to upgrade to etcd3 with the new calico v1.0 in the future

ansible/group_vars/etcd-networking.yaml
```diff
@@ -8,8 +8,8 @@ etcd_certificates_cert_file: "{{ etcd_install_dir }}/{{ etcd_certificates_cert_f
...
+etcdctl_install_bin_name: etcdctl
```
alexbrand: Won't this overwrite the etcdctl v3?

dkoshkin: they both now use the same etcdctl v3 since its backwards compatible.

PR #163, opened 2016-12-15
Vendor integration test dependencies using Glide

alexbrand:
Created a glide.yaml file inside the integration test directory. This allows us to vendor the dependencies required for running the integration suite.

Also reduced the number of retries for the AWS Route 53 requests.


alexbrand:
I ran into issues with Ginkgo and Gomega with stuff being undefined. I think it was mainly Gomega which hasn't been released in a while. I'll check again though.

alexbrand:
Yeah. Gomega hasn't had a release in years, and we still need to `go get` Ginkgo to get binary in the path (so that we can execute it). So I don't think we can lock them down any tighter.

dkoshkin:
Ok, looks really good to me since we can get passing tests now :)

PR #167, opened 2016-12-19
Make host file modification more efficient

alexbrand:
With this change, modifying hosts files scales linearly. 

Added coverage in the integration suite for the hosts file modification feature.

### DIFF
--- a/ansible/roles/hosts/tasks/main.yaml
+++ b/ansible/roles/hosts/tasks/main.yaml
@@ -2,6 +2,14 @@
   lineinfile: dest=/etc/hosts regexp='.*{{ inventory_hostname }}$' state=absent
 
 - name: add nodes to hosts file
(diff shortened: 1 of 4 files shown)


dkoshkin:
What happens if you rerun this with different data, does the data between the markers gets replaced? 

Tying to think what would happen when adding a worker.

alexbrand:
Excellent observation. Thanks @dkoshkin 

I'll investigate and add a test as well.

alexbrand:
I added a test where we go through the add-worker workflow. I also verified that the hosts files get updated when adding a worker.

dkoshkin:
Did you do a manual sanity check to verify that the hosts file was updated or just relying on the successful install?

alexbrand:
Logged into the machine to verify.

dkoshkin:
That works, LGTM

PR #170, opened 2016-12-19
Add make target for focused integration test

alexbrand:
Makes it easier to run a specific test using an env var.

### DIFF
--- a/INTEGRATION_TESTING.md
+++ b/INTEGRATION_TESTING.md
@@ -39,15 +39,20 @@ In order to run tests against Packet.net, you'll need to define the following en
 
  You run integration tests via
 
- ```make integration-tests```
+ ```make integration-test```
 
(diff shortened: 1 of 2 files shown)


dkoshkin:
Yes!! Do you think it would be beneficial to also add some sort of an ID to each test to make it easier "focusing" on a single test?

alexbrand:
Our spec names should be all we need for this. If not, this means the names we have are not specific/descriptive enough.

dkoshkin:
They do, I meant having shorted names `INSTALL_1`, its just a small inconvenience copy/paste multiple lines :)

alexbrand:
We could, but I think it'll get messy over time and hard to maintain. I have found using regexp in the focus argument to be powerful enough in all cases I've run into.

dkoshkin:
LGTM

PR #175, opened 2016-12-19
Fix ingress node validation

### LINKED ISSUE #174: Validation fails if no ingress nodes are defined
I created a new plan file with zero ingress nodes, and validation is failing:

```
root@etcd01:~# ./kismatic install apply

- Ingress nodes: At least one node is required
- Ingress nodes: Node count must be greater than 0
error validating plan: validation error prevents installation from proceeding
```

This is my plan file:
```
cluster:
  name: kubernetes
  admin_password: tufgosilpatoltira      # This password is used to login to the Kubernetes Dashboard and can also be used for administration without a security certificate
  allow_package_installation: true       # When false, installation will not occur if any node is missing the correct deb/rpm packages. When true, the installer will attempt to install missing packages for you.
  networking:
    type: overlay                        # overlay or routed. Routed pods can be addressed from outside the Kubernetes cluster; Overlay pods can only address each other.
    pod_cidr_block: 172.16.0.0/16        # Kubernetes will assign pods IPs in this range. Do not use a range that is already in use on your local network!
    service_cidr_block: 172.17.0.0/16    # Kubernetes will assign services IPs in this range. Do not use a range that is already in use by your local network or pod network!
    policy_enabled: false                # When true, enables network policy enforcement on the Kubernetes Pod network. This is an advanced feature.
    update_hosts_files: false            # When true, the installer will add entries for all nodes to other nodes' hosts files. Use when you don't have access to DNS.
  certificates:
    expiry: 17520h                       # Self-signed certificate expiration period in hours; default is 2 years.
  ssh:
    user: root
    ssh_key: /root/id_rsa            # Absolute path to the ssh private key we should use to manage nodes.
    ssh_port: 22
docker_registry:                         # Here you will provide the details of your Docker registry or setup an internal one to run in the cluster. This is optional and the cluster will always have access to the Docker Hub.
  setup_internal: false                  # When true, a Docker Registry will be installed on top of your cluster and used to host Docker images needed for its installation.
  address: ""                            # IP or hostname for your Docker registry. An internal registry will NOT be setup when this field is provided. Must be accessible from all the nodes in the cluster.
  port: 8443                             # Port for your Docker registry.
  CA: ""                                 # Absolute path to the CA that was used when starting your Docker registry. The docker daemons on all nodes in the cluster will be configured with this CA.
etcd:                                    # Here you will identify all of the nodes that should play the etcd role on your cluster.
  expected_count: 1
  nodes:
  - host: "etcd01"                             # The (short) hostname of a node, e.g. etcd01
    ip: "192.168.205.10"                               # The ip address the installer should use to manage this node, e.g. 8.8.8.8.
    internalip: ""                       # If the node has an IP for internal traffic, enter it here; otherwise leave blank.
master:                                  # Here you will identify all of the nodes that should play the master role.
  expected_count: 1
  load_balanced_fqdn: "etcd01"                 # If using a load balanced master fqdn enter it here; otherwise use the fqdn of any master node.
  load_balanced_short_name: "etcd01"           # If using a load balanced master enter its short name here; otherwise use the short name of any master node.
  nodes:
  - host: "etcd01"
    ip: "192.168.205.10"
    internalip: ""
worker:                                  # Here you will identify all of the nodes that will be workers.
  expected_count: 1
  nodes:
  - host: "etcd01"
    ip: "192.168.205.10"
    internalip: ""
ingress:
  expected_count: 0
  nodes: []
```

alexbrand:
Fixes #174 

The problem is that the conditional check that was gating ingress node validation evaluates to true when the ingress section is present in the file. Specifically this: `p.Ingress.Nodes != nil`

### DIFF
--- a/pkg/cli/plan.go
+++ b/pkg/cli/plan.go
@@ -93,7 +93,7 @@ func buildPlan(etcdNodes int, masterNodes int, workerNodes int, ingressNodes int
 	}
 
 	if ingressNodes > 0 {
-		plan.Ingress = install.NodeGroup{
+		plan.Ingress = install.OptionalNodeGroup{
 			ExpectedCount: ingressNodes,
 		}
 	}
--- a/pkg/install/plan_types.go
+++ b/pkg/install/plan_types.go
@@ -44,6 +44,9 @@ type NodeGroup struct {
 	Nodes         []Node
 }
 
+// An OptionalNodeGroup is a collection of nodes that can be empty
+type OptionalNodeGroup NodeGroup
+
 // MasterNodeGroup is the collection of master nodes
 type MasterNodeGroup struct {
 	ExpectedCount         int    `yaml:"expected_count"`
@@ -67,5 +70,5 @@ type Plan struct {
 	Etcd           NodeGroup
 	Master         MasterNodeGroup
 	Worker         NodeGroup
-	Ingress        NodeGroup
+	Ingress        OptionalNodeGroup
 }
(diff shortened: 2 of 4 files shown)


dkoshkin:
Good catch, how did the integration tests not catch this?

alexbrand:
The plan files that do not have ingress defined in integration tests are missing that section entirely, instead of having it with expected count = 0.

dkoshkin:
Gotcha

dkoshkin:
LGTM

alexbrand:
Merging as this is caused by the known issue with kube-dns.

PR #176, opened 2016-12-20
Validate that python is installed when asking for an ansible runner

### LINKED ISSUE #173: Validation error is misleading when python is not installed
I am attempting to install KET on a 1 node environment (install and target node are the same). Started with the core Ubuntu server image which does not have all the standard libraries. 

I tried to run kismatic install validate and received the following error:
```
 ./kismatic install validate

error running preflight: error running ansible playbook: error running playbook: fork/exec ansible/bin/ansible-playbook: no such file or directory
```
There are two asks here:
* The validation step should make sure that you have the underlying libraries installed
* The documentation should list any foundational requirements to run kismatic

alexbrand:
Fixes #173 

Adds validation so that a more meaningful error is returned to the user when python is not available on the machine.


dkoshkin:
LGTM

PR #177, opened 2016-12-20
Improve SSH validation error message

alexbrand:
Before this change, the SSH validation error was using the InternalIP in the error message, which is not always defined, and is also not the IP we use for SSH:

```
Validating SSH connections to nodes                                             [ERROR]
- Etcd nodes: error SSH into node: , dial tcp 192.168.205.15:22: getsockopt: no route to host
node SSH error prevents installation from proceeding
```

The error message now shows the correct IP address:
```
Validating SSH connections to nodes                                             [ERROR]
- Etcd nodes: SSH connectivity validation failed for "192.168.205.15": dial tcp 192.168.205.15:22: getsockopt: no route to host
SSH connectivity validation failure prevents installation from proceeding
```

### DIFF
--- a/pkg/cli/validate.go
+++ b/pkg/cli/validate.go
@@ -68,7 +68,7 @@ func doValidate(out io.Writer, planner install.Planner, opts *validateOpts) erro
 	if !ok {
 		util.PrettyPrintErr(out, "Validating SSH connections to nodes")
 		printValidationErrors(out, errs)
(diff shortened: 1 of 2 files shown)


dkoshkin:
LGTM

alexbrand:
The test failure was caused by the known issue with kube-dns. Merging.

PR #178, opened 2016-12-20
Remove condition that prevents smoke test from running in minikube setup

alexbrand:
I set up a Minikube cluster and discovered that `kuberang` was not installed on the machine. This prompted me to look into it, and realized that the smoke test is not being run for minikube-style deployments.


dkoshkin:
I've noticed this, Im curious to see if all the tests pass I dont know if @dasmb did this after seeing some failures in minikube

alexbrand:
Failure is due to known kube-dns issue. Merging.

PR #179, opened 2016-12-20
Add newline character after version command

austinlparker:
Does what it says on the tin.


dkoshkin:
Looks good

dkoshkin:
Accidentally closed instead of merge

PR #180, opened 2016-12-20
Revert kubedns change. Go back to using kubeconfig.

alexbrand:
Due to the `kubernetes` service not really being a load balanced service, we have decided to continue using the kubeconfig in kube-dns. This ensures that kube-dns continues working correctly in an HA scenario, in the case where an API server goes down.


dkoshkin:
LGTM

alexbrand:
Going to mount the hosts file to the kube-dns container as part of this PR.

dkoshkin:
What does mounting the hosts file trying to solve, we currently don't have any entries for the FQDN. Even if you were to add the entry, when a non-resolvable address is supplied in the plan file what will be the address.
The only scenario I can see this being useful is for a single master setup, when the user provides the hostname instead of the IP in the FQDN, but that can be solved by changing the wording, something like:
```
load_balanced_fqdn: "192.168.205.11"   # If using a load balanced master fqdn enter it here; otherwise use the IP address of the master node.
```

alexbrand:
You are correct. That is the only scenario that gets solved by mounting the hosts file. There have been a number of users that have run into this, so it seems like a reasonable improvement.

dkoshkin:
Another point is that the `load_balanced_fqdn` is field is used when generating the kubeconfig for kubectl, I fear that mounting the hosts file might encourage the users to use addresses that are not resolvable from outside the cluster and that is most likely not the desired behavior for most cases.

alexbrand:
We decided to avoid mounting the hosts file because kubernetes seems to manage it. Instead, we have opted for improving the in-line documentation.

PR #182, opened 2016-12-20
Use the actual kube-dns version in rep. controller

alexbrand:
While debugging kube-dns, I noticed that the wrong version was being used for the replication controller's name.


dkoshkin:
LGTM

PR #183, opened 2016-12-20
Add release process docs [skip-ci]

alexbrand:
@dkoshkin PTAL


dkoshkin:
Yep looks good.

alexbrand:
Thanks!

PR #184, opened 2016-12-20
Added provision 1.0 to tarball

dasmb:
This PR adds the provision command to the tarballs for linux and darwin.


alexbrand:
LGTM

PR #190, opened 2016-12-22
Use ANSIBLE_CONFIG, use ansible.cfg file for all static variables

### LINKED ISSUE #93: SSH fails when unix socket domain is long
```
debug3: muxserver_listen: temporary control path /Users/dkoshkin/.ansible/cp/ansible-ssh-ec2-52-90-92-217.compute-1.amazonaws.com-22-centos.0zjeQYGHozrcP7hw
unix_listener: "/Users/dkoshkin/.ansible/cp/ansible-ssh-ec2-52-90-92-217.compute-1.amazonaws.com-22-centos.0zjeQYGHozrcP7hw" too long for Unix domain socket
```

dkoshkin:
Fixes #93 

Currently `ansible.cfg` only gets used in our integration tests and does not get packaged up in a distribution.
Moved `ansible.cfg` to a common place and use an ENV var to point to the file whenever ansible is executed both during integration tests and using the binary.

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -1,5 +1,6 @@
 [defaults]
 timeout=60
+host_key_checking = False
 
 [ssh_connection]
 ssh_args = -o ControlMaster=auto -o ControlPersist=1800s -o ControlPath=/tmp/ssh-%r-%h-%p
--- a/pkg/ansible/runner.go
+++ b/pkg/ansible/runner.go
@@ -137,10 +137,9 @@ func (r *runner) startPlaybook(playbookFile string, inv Inventory, vars ExtraVar
 	}
 
 	os.Setenv("PYTHONPATH", r.pythonPath)
-	os.Setenv("ANSIBLE_HOST_KEY_CHECKING", "False")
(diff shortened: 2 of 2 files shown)


ansible/ansible.cfg
```diff
@@ -1,5 +1,6 @@
 [defaults]
 timeout=60
+host_key_checking = False
 
 [ssh_connection]
 ssh_args = -o ControlMaster=auto -o ControlPersist=1800s -o ControlPath=/tmp/ssh-%r-%h-%p
```
alexbrand: Do we want to override this for all installs going forward?

dkoshkin: We are already doing this with `ANSIBLE_HOST_KEY_CHECKING`, we could take it out and just document how to use it.

alexbrand: Right, but the host key checking option was being used for all our installs already. I don't think the SSH args was?

dkoshkin: The `ControlPath` was but not the other two, I think there is an ordering thing, because `ssh_args` is defined defined in ansibl.cfg the `SSH_CONTROL_PATH`in SNAP was being overwritten by the file(which broke the builds), I added the new var in Snap with the working `ControlPath` along with the other options

alexbrand:
@dkoshkin Should we merge this?

PR #198, opened 2016-12-23
Added ssh validation retry

dkoshkin:


alexbrand:
LGTM

alexbrand:
How long does a user have to wait to realize that she used a wrong IP address with this change? 90 seconds seems a bit high

dkoshkin:
Im actually not really sure why that test is failing, I tested locally and it passed under 60 seconds, 4 machines for `validate` + 4 machines for `install`

PR #210, opened 2016-12-30
Some installation performance improvements

dasmb:
Our performance with what I'm going to call "global clusters" -- install machine, binary repositories and target infrastructure geographically distinct from each other -- is not up to par. And while we can encourage KET users to cache binaries locally and use an install machine that is geographically close to the cluster, there are certainly some tweaks we can perform to improve SSH & download performance for these users.

Note that this PR uses an experimental CloudFront repo for binaries. This causes issues when testing a binary release as it can take an indeterminate time, up to an hour, for all changes to propagate and propagation order matters.

### DIFF
--- a/ansible/group_vars/all.yaml
+++ b/ansible/group_vars/all.yaml
@@ -166,8 +166,8 @@ pause_img: gcr.io/google_containers/pause-amd64:{{ pause_version }}
 allow_package_installation: true
 yum_repository_url: https://kismatic-packages-rpm.s3-accelerate.amazonaws.com
 yum_gpg_key_url: https://kismatic-packages-rpm.s3-accelerate.amazonaws.com/public.key
-deb_repository_url: https://kismatic-packages-deb.s3-accelerate.amazonaws.com
-deb_gpg_key_url: https://kismatic-packages-deb.s3-accelerate.amazonaws.com/public.key
+deb_repository_url: https://d38kcxml09yeoe.cloudfront.net
+deb_gpg_key_url: https://d38kcxml09yeoe.cloudfront.net/public.key
 # Uncomment below to use test repo
 # yum_repository_url: https://kismatic-packages-rpm-test.s3-accelerate.amazonaws.com
 # yum_gpg_key_url: https://kismatic-packages-rpm-test.s3-accelerate.amazonaws.com/public.key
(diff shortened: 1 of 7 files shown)


pkg/ansible/runner.go
```diff
@@ -154,7 +157,12 @@ func (r *runner) startPlaybook(playbookFile string, inv Inventory, vars ExtraVar
 	os.Setenv("ANSIBLE_JSON_LINES_PIPE", r.namedPipe)
 
 	// Print Ansible command
-	fmt.Fprintln(r.out, strings.Join(cmd.Args, " "))
+	fmt.Fprintf(r.out, "export PYTHONPATH=%v\n", os.Getenv("PYTHONPATH"))
+	fmt.Fprintf(r.out, "export ANSIBLE_CALLBACK_PLUGINS=%v\n", os.Getenv("ANSIBLE_CALLBACK_PLUGINS"))
+	fmt.Fprintf(r.out, "export ANSIBLE_CALLBACK_WHITELIST=%v\n", os.Getenv("ANSIBLE_CALLBACK_WHITELIST"))
+	fmt.Fprintf(r.out, "export ANSIBLE_CONFIG=%v\n", os.Getenv("ANSIBLE_CONFIG"))
```
dkoshkin: What are these doing?

alexbrand:
LGTM!

pkg/ansible/runner.go
dasmb: We were previously outputting the ansible command line to an ansible log file, ostensibly for debug purposes. But that command is useless without these corresponding environment variables, so I've added them to the output.

ansible/group_vars/all.yaml
```diff
@@ -166,8 +166,8 @@ pause_img: gcr.io/google_containers/pause-amd64:{{ pause_version }}
 allow_package_installation: true
 yum_repository_url: https://kismatic-packages-rpm.s3-accelerate.amazonaws.com
 yum_gpg_key_url: https://kismatic-packages-rpm.s3-accelerate.amazonaws.com/public.key
-deb_repository_url: https://kismatic-packages-deb.s3-accelerate.amazonaws.com
-deb_gpg_key_url: https://kismatic-packages-deb.s3-accelerate.amazonaws.com/public.key
+deb_repository_url: https://d38kcxml09yeoe.cloudfront.net
+deb_gpg_key_url: https://d38kcxml09yeoe.cloudfront.net/public.key
```
dkoshkin: Are we switching to CF? If so why only for debs?

alexbrand: Ah! I glanced over this thinking it was switching both GPG keys to CF.

dkoshkin:
LGTM

alexbrand:
Failure seems like a flake, so I am merging this in.

PR #212, opened 2017-01-02
Dependency on nfs-utils

### LINKED ISSUE #200: Playbook for installing NFS
Worker nodes will need to be able to act as NFS clients.

To this end, we will need to ensure that `nfs-utils` will be installed on CentOS, RHEL and Ubuntu.

Kismatic packages may need to be updated to ensure this dependency, as should the requirements in docs.

dkoshkin:
Closes #200 

The only change required is the addition of a new dependency in `kismatic-kuberntes` packages:
* Deb: `nfs-common`
* RPM: `nfs-utils`  

Completed in https://github.com/apprenda/kismatic-distro-packages/commit/6168cca6f4fd88549858e1a5e5fff1611b1a303d

There is no Anisble playbook changes since the package will be automatically installed by the OS package manager
There is no Doc changes as this is just one of many underlying utilities required for kuberntes and is handled by the OS package manager

### DIFF
--- a/ansible/group_vars/all.yaml
+++ b/ansible/group_vars/all.yaml
@@ -1,11 +1,11 @@
 #===============================================================================
 # VERSIONS
-kismatic_kubernetes_master_yum_version: 1.5.1_1-1
-kismatic_kubernetes_node_yum_version: 1.5.1_1-1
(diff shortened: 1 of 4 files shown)


alexbrand:
LGTM. Let's merge this in if it's ready.

PR #216, opened 2017-01-03
Added kismatic ssh command

### LINKED ISSUE #194: Implement ssh command
This issue has been broken out from https://github.com/apprenda/kismatic/issues/191

When a user sets up a Kismatic installation, they are required to provide a public/private ssh key pair  for accessing the target hosts.  If a "minikube like" cluster is automatically provisioned and launched, it would be useful to allow the user to SSH into the VM host by issuing:

```
$ kismatic ssh <?nodename>
```

Obviously I am only addressing a single host machine installation with this syntax.  Maybe we default to the first node listed in the plan file if a node name is not provided.

dkoshkin:
Fixes #194 

Added `kismatic ssh HOST [commands] -f` command
* [commands] are optional, when provided they will be executed and the ssh sessions will end


alexbrand:
We discussed revisiting this so that we shell out to `ssh` instead of using the docker machine library.

pkg/install/validate.go
```diff
@@ -36,27 +32,36 @@ func ValidateNode(node *Node) (bool, []error) {
...
+func ValidateSSHConnections(con *SSHConnections, prefix string) (bool, []error) {
 	v := newValidator()
 
 	v.validateWithErrPrefix(prefix, con)
 
 	return v.valid()
 }
 
+// ValidateSSHConnection tries to establish SSH connection with the details provieded for a single node
+func ValidateSSHConnection(con *SSHConnection, prefix string) (bool, []error) {
```
alexbrand: Similar concern here with ValidateSSHConnection and ValidateSSHConnectionS. I suggest creating a single function that validates ssh connections. Maybe:

```
func ValidateSSHConnection(sshConfig SSHConfig, nodes ...Node) (bool, []error)
```

pkg/ssh/ssh.go
```diff
@@ -0,0 +1,71 @@
...
+	"io/ioutil"
+
+	"golang.org/x/crypto/ssh"
+
+	libmachine "github.com/docker/machine/libmachine/ssh"
+)
+
+// Client is an SSH client wrapper for libmachine
+type Client interface {
+	libmachine.Client
```
alexbrand: What's the purpose of this interface? If it's to isolate from underlying changes to libmachine, we should define our own interface and methods.

pkg/ssh/ssh.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// NewClient returns an SSH client
+func NewClient(details Details) (client Client, err error) {
```
alexbrand: It's weird to me that as a consumer of this package, I have to implement my own struct that implements `Details` to be able to get a client. Was there any reason for choosing this approach?

pkg/ssh/ssh.go
```diff
@@ -0,0 +1,71 @@
...
+// Details is an interface for the details to allow to SSH into nodes
+type Details interface {
+	GetSSHAddress() string
+	GetSSHPort() int
+	GetSSHKeyPath() string
+	GetSSHUsername() string
+}
+
+// ValidUnecryptedPrivateKey parses SSH private key
+func ValidUnecryptedPrivateKey(file string) error {
```
alexbrand: Typo

pkg/ssh/ssh.go
```diff
@@ -0,0 +1,71 @@
...
+		return fmt.Errorf("Parse SSH key error: %v", err)
+	}
+
+	return nil
+}
+
+func isEncrypted(buffer []byte) (bool, error) {
+	// There is no error, just a nil block
+	block, _ := pem.Decode(buffer)
+	// File cannot be decoded, maybe it's some unecpected format
```
alexbrand: typo

pkg/install/validate.go
```diff
@@ -230,20 +225,13 @@ func (s *SSHConnection) validate() (bool, []error) {
...
-
-	// Try to connect with a timeout
-	sshConn, _, _, err := ssh.NewClientConn(conn, server, sshClientConfig)
-	if err == nil {
-		conn.Close()
-		sshConn.Close()
-	}
-	return err
+	// just exit
+	return client.Shell("exit")
```
alexbrand: Seems like we lost the ability to set the connection timeout?

dkoshkin:
LGTM, I believe the CLI docs need to be regenerated to pickup the new description

pkg/cli/ssh.go
```diff
@@ -0,0 +1,88 @@
...
+
+// NewCmdSSH returns an ssh shell
+func NewCmdSSH(out io.Writer) *cobra.Command {
+	opts := &sshOpts{}
+
+	cmd := &cobra.Command{
+		Use: "ssh HOST [commands]",
+		Short: `ssh into a node in the cluster.
+
+HOST must be one of the following:
```
alexbrand: Does this look OK in the help screens? I don't quite recall the different between `Short` and `Long`. This might break the formatting in `kismatic -h`

pkg/install/plan_types.go
```diff
@@ -72,3 +74,69 @@ type Plan struct {
...
+func (p *Plan) GetUniqueNodeIPs() []string {
+	ipMap := make(map[string]bool)
+	nodes := p.getAllNodes()
+	for _, node := range nodes {
+		ipMap[node.IP] = true
+	}
+
+	ips := make([]string, len(ipMap))
+
+	i := 0
```
alexbrand: Don't need this?

alexbrand: Nope. We do need it :)

pkg/install/plan_types.go
```diff
@@ -72,3 +74,69 @@ type Plan struct {
...
+	if foundNode == nil {
+		switch host {
+		case "master":
+			foundNode = &p.Master.Nodes[0]
+		case "etcd":
+			foundNode = &p.Etcd.Nodes[0]
+		case "worker":
+			foundNode = &p.Worker.Nodes[0]
+		case "ingress":
+			foundNode = &p.Ingress.Nodes[0]
```
alexbrand: Will panic if there are no ingress nodes defined.

pkg/install/validate.go
```diff
@@ -184,35 +183,30 @@ func (s *SSHConfig) validate() (bool, []error) {
 }
 
 // validate SSH access to the nodes
-func (s *SSHConnection) validate() (bool, []error) {
+func (s sshConnectionSet) validate() (bool, []error) {
 	v := newValidator()
 
-	auth, err := util.GetUnencryptedPublicKeyAuth(s.SSHConfig.Key)
+	err := ssh.ValidUnecryptedPrivateKey(s.SSHConfig.Key)
```
alexbrand: s/Unecrypted/Unencrypted

pkg/cli/ssh.go
dkoshkin: ^ Yeah good question

dkoshkin: Looks good
```
➜  out git:(kismatic-ssh) ./kismatic ssh --help
ssh into a node in the cluster.

HOST must be one of the following:
- A hostname defined in the plan filepath
- An alias: master, etcd, worker or ingress. This will ssh into the first defined node of that type.

Usage:
  kismatic ssh HOST [commands] [flags]

Flags:
  -f, --plan-file string   path to the installation plan file (default "kismatic-cluster.yaml")
```

alexbrand: I meant `kismatic --help`

dkoshkin: Hmm yeah its confusing where that block belongs to, that might be better in the `Long`
```
➜  out git:(kismatic-ssh) ./kismatic --help
kismatic is the main tool for managing your Kubernetes cluster
more documentation is availble at https://github.com/apprenda/kismatic

Usage:
  kismatic [flags]
  kismatic [command]

Available Commands:
  install     install your Kubernetes cluster
  ssh         ssh into a node in the cluster.

HOST must be one of the following:
- A hostname defined in the plan filepath
- An alias: master, etcd, worker or ingress. This will ssh into the first defined node of that type.
  version     display the Kismatic CLI version

Use "kismatic [command] --help" for more information about a command.
```

alexbrand: Yeah. I think that's a problem.

PR #217, opened 2017-01-03
Release script for drafting release on GitHub

alexbrand:
See updated RELEASE.md for details.

In Snap, we will check if current commit == commit of latest tag. If true, we will run the release script.


dkoshkin:
LGTM, I guess we will find out when we test

PR #222, opened 2017-01-04
Support for storage nodes in the plan file

alexbrand:
This PR adds support for specifying storage nodes in the plan file. 

Changes:
* Ansible play for installing Gluster packages when package installation is enabled
* Ansible play for bootstrapping a Gluster cluster
* Deploy kubelet to storage nodes. Schedulable if listed in worker, otherwise unschedulable
* Add inspector checks for packages and NFS ports (111, 2049)
* Add prompt in `install plan` for storage nodes with default = 0
* Collapse multiple kubelet plays into a single play
* Create kubernetes service called `kismatic-storage`, with endpoints that point to the `storage` node's IPs and NFS ports (111, 2049)
* Integration test that stands up gluster cluster using kismatic, creates a replicated volume, writes a file on one node, and verifies the existence of the file on another


dkoshkin:
Do we also want to include an integration test with a full install and 2 storage nodes, for now it would just verify we didn't break anything else.

Later on it can be modified to add and list storage drives.

dkoshkin:
Looks like we aren't validating storage nodes in `validate.go`

alexbrand:
Added validation. Thanks @dkoshkin

ansible/group_vars/all.yaml
```diff
@@ -19,6 +19,10 @@ defaultbackend_version: "1.0"
 nginx_ingress_controller_version: "0.8.3"
 docker_registry_version: "2.5.1"
 pause_version: "3.0"
+glusterfs_server_version_rhel: "3.8.5-1.el7"
+glusterfs_server_version_ubuntu: "3.7.6-1ubuntu1"
```
dkoshkin: Is it possible to use the same version for both Linux distros?

alexbrand: I'll look into it. There might be a way by using the Gluster deb repo, instead of the Ubuntu repo.

dkoshkin:
LGTM pending on the merge conflict.

PR #224, opened 2017-01-06
Reduce retries on ingress integration test

alexbrand:
After catching an integration test that timed out, and doing a post-mortem on the nodes that were left over, I noticed that pulling the images used in the ingress integration test was failing:

```
 28m 23m 3 {kubelet ip-10-0-3-14} Warning  FailedSync	Error syncing pod, skipping: failed to "StartContainer" for "echoserver" with ErrImagePull: "image pull failed for gcr.io/google_containers/echoserver:1.0, this may be because there are no credentials on this request.  details: (net/http: request canceled)"
```

This leads me to believe that the integration tests are sometimes getting stuck in the ingress verification retry loop. 

As a first step, I propose reducing the number of retries as I prefer a test that fails over a test that times out the entire suite on Snap. Once we have a couple of runs with reduced retries, we can continue investigating / improving.

### DIFF
--- a/integration/install.go
+++ b/integration/install.go
@@ -20,9 +20,6 @@ import (
 func leaveIt() bool {
 	return os.Getenv("LEAVE_ARTIFACTS") != ""
 }
-func bailBeforeAnsible() bool {
-	return os.Getenv("BAIL_BEFORE_ANSIBLE") != ""
-}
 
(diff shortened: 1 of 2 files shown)


dkoshkin:
LGTM

PR #226, opened 2017-01-06
[DNM] Introduce test for disconnected install

alexbrand:
Introduce a test for testing the installation of kismatic on nodes that do not have access to the internet. The test found a couple of issues that are fixed in this PR.

### DIFF
--- a/ansible/group_vars/all.yaml
+++ b/ansible/group_vars/all.yaml
@@ -136,7 +136,12 @@ setup_internal_docker_registry: false
 docker_registry_address: "{{ hostvars[groups['master'][0]].internal_ipv4 }}"
 docker_registry_port: 8443
 
+# these images are included in the kismatic node package and get loaded on all k8s nodes
+# they will not be in the registry and loaded directly on each node in the cluster
 calico_node_img: calico/node:{{calico_docker_version}}
+defaultbackend_img: gcr.io/google_containers/defaultbackend:{{ defaultbackend_version }}
+nginx_ingress_controller_img: gcr.io/google_containers/nginx-ingress-controller:{{ nginx_ingress_controller_version }}
+pause_img: gcr.io/google_containers/pause-amd64:{{ pause_version }}
 
 calico_kube_policy_controller_orig_img: calico/kube-policy-controller:{{ calico_kube_policy_controller_version }}
 calico_kube_policy_controller_priv_img: "{{ docker_registry_address }}:{{ docker_registry_port }}/calico/kube-policy-controller:{{ calico_kube_policy_controller_version }}"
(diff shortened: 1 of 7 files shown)


dkoshkin:
I like the idea, but what will happen when we introduce the storage feature which rely on the EPEL repo.

alexbrand:
What will happen to the ability to install with no internet? I guess the user will have to pre-install the gluster packages we require, similarly to how the kismatic packages must be installed.

ansible/group_vars/all.yaml
```diff
@@ -158,8 +158,13 @@ kubernetes_dashboard_orig_img: gcr.io/google_containers/kubernetes-dashboard-amd
...
 
-defaultbackend_img: gcr.io/google_containers/defaultbackend:{{ defaultbackend_version }}
-nginx_ingress_controller_img: gcr.io/google_containers/nginx-ingress-controller:{{ nginx_ingress_controller_version }}
+defaultbackend_orig_img: gcr.io/google_containers/defaultbackend:{{ defaultbackend_version }}
+defaultbackend_priv_img: "{{ docker_registry_address }}:{{ docker_registry_port }}/defaultbackend:{{ defaultbackend_version }}"
+defaultbackend_img: '{% if use_private_docker_registry is defined and use_private_docker_registry|bool == true %}{{ defaultbackend_priv_img }}{% else %}{{ defaultbackend_orig_img }}{% endif %}'
+
+nginx_ingress_controller_orig_img: gcr.io/google_containers/nginx-ingress-controller:{{ nginx_ingress_controller_version }}
+nginx_ingress_controller_priv_img: "{{ docker_registry_address }}:{{ docker_registry_port }}/nginx-ingress-controller:{{ nginx_ingress_controller_version }}"
+nginx_ingress_controller_img: '{% if use_private_docker_registry is defined and use_private_docker_registry|bool == true %}{{ nginx_ingress_controller_priv_img }}{% else %}{{ nginx_ingress_controller_orig_img }}{% endif %}'
```
dkoshkin: The reason why these don't get loaded into the registry is because they are already included in the `node` package and are loaded into docker in https://github.com/apprenda/kismatic/blob/master/ansible/roles/docker/tasks/main.yaml

If you were seeing issues with those images I suspect when they are `docker load` into docker they get a different name or version.

ansible/roles/docker-registry-images/tasks/main.yaml
```diff
@@ -8,6 +8,8 @@
      - kube-dnsmasq
      - exechealthz
      - kubernetes-dashboard
+     - defaultbackend
+     - nginx-ingress-controller
```
dkoshkin: Same thing as above, should not be needed since they are loaded directly into docker.

ansible/roles/docker-registry-images/tasks/main.yaml
```diff
@@ -17,6 +19,8 @@
       - "{{kube_dnsmasq_orig_img}} {{kube_dnsmasq_priv_img}}"
       - "{{exechealthz_orig_img}} {{exechealthz_priv_img}}"
       - "{{kubernetes_dashboard_orig_img}} {{kubernetes_dashboard_priv_img}}"
+      - "{{defaultbackend_orig_img}} {{defaultbackend_priv_img}}"
+      - "{{nginx_ingress_controller_orig_img}} {{nginx_ingress_controller_priv_img}}"
```
dkoshkin: See above

ansible/roles/docker-registry-images/tasks/main.yaml
```diff
@@ -26,3 +30,5 @@
       - "{{kube_dnsmasq_priv_img}}"
       - "{{exechealthz_priv_img}}"
       - "{{kubernetes_dashboard_priv_img}}"
+      - "{{defaultbackend_priv_img}}"
```
dkoshkin: See above

ansible/group_vars/all.yaml
dkoshkin: The reasoning for this distinction is because these images are used by all of the nodes and it made sense to distribute them and load them instead of all of the nodes needing to pull it from a central registry.

dkoshkin:
LGTM

alexbrand:
Let's use Amazon SGs instead of monkeying around with iptables.

dkoshkin:
Should we merge this to be included in the release, we can modify the actual test later or exclude the test altogether from this PR.

PR #227, opened 2017-01-06
Add timeout to integration SSH check

alexbrand:
If a node provisioned by the integration test does not start up properly (I've seen it on AWS), the integration test will get stuck waiting for SSH. 

For this reason, the SSH check should be bounded with a timeout.


dkoshkin:
LGTM

Issue #201, opened 2016-12-27
Introduce NFS shares to Plan file


dasmb:
Plan file should have a new section for defining an NFS share.

This will be used during creation to set up a PersistentVolume in the created Kubernetes cluster, so that persistent workloads that are installed as part of creation can use something other than host storage.

This share should be validated -- we need to be sure it's mountable by worker nodes, that it has available storage and that it is writable -- and then added to the cluster as a PV using kubectl.

It's vitally important that the NFS share not be duplicated on the cluster. Offering the same store twice might make it available for a storage claim twice.

alexbrand:
Closed via #229

PR #229, opened 2017-01-10
Added Nfs shares to plan

dasmb:
In pursuit of https://github.com/apprenda/kismatic/issues/201, support for specifying existing NFS shares has been added.

This introduced the first collection of cluster-level details -- NFS shares -- that wasn't related to details of nodes, forcing a refactor of the technique we use to pass data out of kismatic into Ansible. All existing "extra-vars" have been moved into a `clustercatalog.yaml` file, which means that those extra-vars that had defaults set in `groupvars/all.yaml` now have these defaults set during `execute.go` instead. Since many of the defaults were already being set there, this should reduce confusion when building new cluster level defaults.

Tests of the new NFS behavior (using AWS EFS) are pending and should be created today. I wanted to get eyes on the changes while I work on them.

BTW, runner.go and execute.go could use a refactor. They're getting unmanageably confusing.


ansible/roles/addon-nfs-volume/templates/add-persistent-volume.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: PersistentVolume
+metadata:
+  name: pv{{index}}
+  annotations:
+    volume.beta.kubernetes.io/storage-class: "kismatic"
+spec:
+  capacity:
+    storage: {{nfs_size}}
+  accessModes:
+    - ReadWriteOnce
```
alexbrand: Did we want multi-reader / multi-writer?

pkg/ansible/clustercatalog.go
```diff
@@ -0,0 +1,73 @@
...
+	DockerRegistryAddress        string `yaml:"docker_registry_address"`
+	DockerRegistryPort           string `yaml:"docker_registry_port"`
+
+	ForceEtcdRestart              bool `yaml:"force_etcd_restart"`
+	ForceAPIServerRestart         bool `yaml:"force_apiserver_restart"`
+	ForceControllerManagerRestart bool `yaml:"force_controller_manager_restart"`
+	ForceSchedulerRestart         bool `yaml:"force_scheduler_restart"`
+	ForceProxyRestart             bool `yaml:"force_proxy_restart"`
+	ForceKubeletRestart           bool `yaml:"force_kubelet_restart"`
+	ForceCalicoNodeRestart        bool `yaml:"force_calic_node_restart"`
```
alexbrand: typo in yaml tag?

pkg/ansible/clustercatalog.go
```diff
@@ -0,0 +1,73 @@
...
+	KismaticPreflightCheckerLocal string `yaml:"kismatic_preflight_checker_local"`
+
+	WorkerNode string `yaml:"worker_node"`
+
+	NFSVolumes []NFSVolume
+
+	EnableGluster bool `yaml:"configure_storage"`
+}
+
+type NFSVolume struct {
```
alexbrand: Would it be useful to allow the user to provide a name for the NFS volume?

pkg/ansible/runner.go
```diff
@@ -104,32 +94,42 @@ func (r *runner) WaitPlaybook() error {
...
+
 	inventoryFile := filepath.Join(r.ansibleDir, "inventory.ini")
-	if err = ioutil.WriteFile(inventoryFile, inv.ToINI(), 0644); err != nil {
+	if err := ioutil.WriteFile(inventoryFile, inv.ToINI(), 0644); err != nil {
 		return nil, fmt.Errorf("error writing inventory file to %q: %v", inventoryFile, err)
 	}
-	cmd := exec.Command(filepath.Join(r.ansibleDir, "bin", "ansible-playbook"), "-i", inventoryFile, "-s", playbook, "--extra-vars", extraVars)
+
+	copyFileContents(clusterCatalogFile, filepath.Join(r.runDir, "clustercatalog.yaml"))
+	copyFileContents(inventoryFile, filepath.Join(r.runDir, "inventory.ini"))
```
alexbrand: Do we need error handling for these copy calls?

pkg/install/validate.go
```diff
@@ -347,3 +348,18 @@ func (dr *DockerRegistry) validate() (bool, []error) {
 	}
 	return v.valid()
 }
+
+func (nfs *NFS) validate() (bool, []error) {
```
alexbrand: Do we need to validate that the volumes use absolute paths?

ansible/roles/addon-nfs-volume/tasks/new-share.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+- name: mount share read/write
+  mount: 
+    name: "/mnt/test-{{index}}"
+    src: "{{host}}:{{path}}"
+    fstype: nfs 
+    state: mounted
+
+- name: Grab share size
+  shell: "df /mnt/test-{{index}} | grep -oP '\\K\\d+(?=\\s+\\d+%)'"
```
dasmb: I like it, consider it done.

ansible/roles/addon-nfs-volume/templates/add-persistent-volume.yaml
dasmb: Ah yes, I will revise.

pkg/ansible/clustercatalog.go
dasmb: yep

pkg/ansible/clustercatalog.go
dasmb: I am reticent, as this is another thing to validate.

pkg/install/validate.go
dasmb: We do mount the volume as part of apply. We can also mount it here if you think it's valuable, but the mount currently takes place through ansible, which isn't available in the inspector.

alexbrand: I think we could just check the string, no? The plan file annotation is what got me thinking about this.

`"mount_path":               "The mount path of an NFS share. Must start with /",`

pkg/ansible/runner.go
dasmb: I'll add it.

alexbrand:
LGTM!

PR #231, opened 2017-01-11
Implement command for adding volumes

### LINKED ISSUE #203: Kismatic function to create new Storage volumes
The new command
`kismatic storage new-volume <flags> <size in gigabytes> [volume name]` 

can be used to add storage volumes to the Kubernetes cluster. A Gluster store is required for this function and should be referenced by flags to the command, e.g.

`-f kismatic-cluster.yaml`

Other flags that should be introduced:

```
-r | --replica-count (default 2) The number of times each file will be written.

-d | --distribution-count (default 1) This is the degree to which data will be distributed
 across the cluster. By default, it won't be -- each replica will receive 100% of the data. 
Distribution makes listing or backing up the cluster more complicated by spreading data 
around the cluster but makes reads and writes more performant.

-a | --allow-addresses This is a list, comma delimited, of address wildcards permitted
access to the volume in addition to Kubernetes nodes.
```

Example:
`kismatic storage new-volume -r 2 -d 2 -a 10.10.*.* 10 kismatic` 

will create a distributed, replicated volume named kismatic with a 10 GB quota. We will grant access to any IP address starting with `10.10`. In pursuit of this, 4 bricks will be created -- 2 distributed bricks and 2 replicas of each.

The command will
1. List free space & known volumes on all gluster nodes
2. If there is insufficient space on enough nodes to create the described volume, the command will raise an error.
3. If the new volume name would make a bad directory name (e.g. has spaces, colons, slashes in it), the command will raise an error.
    1. If no volume name is introduced, use `kismatic-<some brief randomness>`
4. Otherwise, will make a list of the (num_replicas) most empty storage nodes.
5. For each node:
    1. Create a new "brick" directory off `/data` using the new volume name.
    2. Set its permissions to 777
6. Create a new volume, referencing each of the bricks created at 5 and named approriately
7. Start the volume
8. Enable NFS, quotas & quota-deem-status [allows df to accurately report quotas]
9. Apply a white list allowing the pod network and the install machine's IP address to access the drive (but nothing else)
10. Set the quota to the desire size
11. Add a new PV for the created volume to kubernetes

dkoshkin:
Fixes #203 

- [x] validate volume name and plan file in CLI
- [x] new CLI command to add volume
- [x] playbook to add new Gluster volume
- [x] IP whitelisting in Gluster
- [x] playbook to add PV
- [x] figure out what to do with `extra-vars` in `kismatic step` command
- [ ] update `Installing on machines with no internet` test with Gluster support


pkg/install/plan_types.go
```diff
@@ -83,3 +83,17 @@ type Plan struct {
 	Storage        OptionalNodeGroup
 	NFS            NFS
 }
+
+// StorageVolume managed by Kismatic
+type StorageVolume struct {
```
dkoshkin: Are we adding the volume details somewhere in the plan file?

alexbrand: We are not, no.

dkoshkin: Ok the reason I ask is because `StorageVolume` is in `plan_types`. 
I also see that there is `ValidateStorageVolume`, you might be right that validate interface is getting crowded with all of these none plan file validation (ssh, cert, storage) it might be due for some refactoring soon.

alexbrand: Indeed. I started exploring that path but there were a couple of more pressing issues for this release. I plan on spending time on this after we release.

dkoshkin:
LGTM

alexbrand:
LGTM

PR #234, opened 2017-01-16
Integration test that consumes kismatic volume from K8s cluster

### LINKED ISSUE #205: Test addition of a stateful feature to Kismatic
As part of this release, we likely won't be introducing a feature that will make use of the new Gluster store. But we will want to test that addition of a stateful feature to the platform would work as expected -- that a specified NFS volume would be tested and used, that specified Gluster cluster would have a new volume created of the correct storage class, and that any feature that was introduced would result in a volume being claimed and named "kismatic."

Kismatic doesn't really have a model of a "feature" at the moment, so picking up this issue should include a design for detecting the "features" of a platform. A list of "existing features" has been produced on the Apprenda wiki (comment here if you would like and I will make copy here).

The test should include a Reader service and a Writer service with at least 2 pods of each.

alexbrand:
Fixes #205 

The test consists of:
* Create a PVC for claiming kismatic storage
* Running a writer job, which will spin up multiple writers that write a file to the PV
* Running a reader job, which will spin up multiple readers that verify the files are on the PV

### DIFF
--- a/integration/storage.go
+++ b/integration/storage.go
@@ -166,7 +166,7 @@ func testVolumeAdd(masterNode NodeDeets, sshKey string) {
 	cmd.Stderr = os.Stderr
 	err := cmd.Run()
 	FailIfError(err, "Error creating a new volume")
-
+	
 	By("Verifying Kuberntes PV was created")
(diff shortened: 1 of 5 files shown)


dkoshkin:
LGTM

PR #237, opened 2017-01-16
Stop downloading unnecessary cfssl binaries

alexbrand:
We don't actually use these binaries, so it seems we can stop downloading them. 

They are vestiges of our initial work around generating certificates.


alexbrand:
Failure is unrelated flake.

dkoshkin:
LGTM

PR #244, opened 2017-01-19
Use nginx instead of the large echoserver image

dkoshkin:
The echoserver image is 650Mbs, the nginx box is already there from the smoketest


alexbrand:
Wow - 650MB for an echoserver?

LGTM

PR #250, opened 2017-01-22
Volume list

### LINKED ISSUE #204: Kismatic function to list storage volumes
One component of storage management is understanding what data is in use by which workloads.

`kismatic storage list-volumes -f kismatic-cluster.yaml`

will list all PVs on a cluster, along with
 - whether they've been claimed or not
 - Their size

& I dunno -- the workloads using them, details about how the volume is distributed around a Gluster cluster.

dkoshkin:
Fixes #204  

* Implement SSH and remove dependency on libmachine
* Volume list command
  * `gluster volume info all`  
  * `gluster volume quota $VOLUME list`
  * `kubectl get pv` 
  * `kubectl get pods --all-namespaces=true` 

Because the API does not return PVC -> pod relationship, the data is retrieved in reverse, a map of pods with PVCs is created and used when building data for each PV


dkoshkin:
```
➜  out git:(volume-list) ✗ ./kismatic volume list
Name:           storage1
Capacity:       1.00GB
Available:      1.00GB
Replica:        2
Distribution:   2
Bricks:         ip-10-0-3-240:/data/storage1,ip-10-0-3-111:/data/storage1,ip-10-0-3-93:/data/storage1,ip-10-0-3-137:/data/storage1
Status:         Available
Claim:
Pods:


Name:                storage2
Capacity:            1.00GB
Available:           1.00GB
Replica:             1
Distribution:        2
Bricks:              ip-10-0-3-240:/data/storage2,ip-10-0-3-111:/data/storage2
Status:              Bound
Claim:               default/claim1
Pods:
  default/mypod
    Containers:
      myfrontend
        MountPath:   /var/www/html


Name:           storage3
Capacity:       1.00GB
Available:      1.00GB
Replica:        2
Distribution:   1
Bricks:         ip-10-0-3-111:/data/storage3,ip-10-0-3-240:/data/storage3
Status:         Bound
Claim:          default/my-awesome-claim
Pods:


Name:                   storage4
Capacity:               1.00GB
Available:              1.00GB
Replica:                1
Distribution:           1
Bricks:                 ip-10-0-3-111:/data/storage4
Status:                 Bound
Claim:                  default/another-claim-frontend1
Pods:
  default/another-pod
    Containers:
      myfrontend
        MountPath:      /var/www/html
      sidecar
        MountPath:      /var/www/html
  default/mypod-2
    Containers:
      frontend
        MountPath:      /var/www/html


Name:           storage5
Capacity:       1.00GB
Available:      1.00GB
Replica:        1
Distribution:   1
Bricks:         ip-10-0-3-111:/data/storage5
Status:         Available
Claim:
Pods:
```

dkoshkin:
```
➜  out git:(volume-list) ✗ ./kismatic volume list -o json
{
    "items": [
        {
            "name": "storage1",
            "capacity": "1.00GB",
            "available": "1.00GB",
            "replicaCount": 2,
            "distributionCount": 2,
            "bricks": [
                {
                    "host": "ip-10-0-3-240",
                    "path": "/data/storage1"
                },
                {
                    "host": "ip-10-0-3-111",
                    "path": "/data/storage1"
                },
                {
                    "host": "ip-10-0-3-93",
                    "path": "/data/storage1"
                },
                {
                    "host": "ip-10-0-3-137",
                    "path": "/data/storage1"
                }
            ],
            "status": "Available"
        },
        {
            "name": "storage2",
            "capacity": "1.00GB",
            "available": "1.00GB",
            "replicaCount": 1,
            "distributionCount": 2,
            "bricks": [
                {
                    "host": "ip-10-0-3-240",
                    "path": "/data/storage2"
                },
                {
                    "host": "ip-10-0-3-111",
                    "path": "/data/storage2"
                }
            ],
            "status": "Bound",
            "claim": {
                "name": "claim1",
                "namespace": "default"
            },
            "pods": [
                {
                    "name": "mypod",
                    "namespace": "default",
                    "containers": [
                        {
                            "name": "myfrontend",
                            "mountPath": "/var/www/html"
                        }
                    ]
                }
            ]
        },
        {
            "name": "storage3",
            "capacity": "1.00GB",
            "available": "1.00GB",
            "replicaCount": 2,
            "distributionCount": 1,
            "bricks": [
                {
                    "host": "ip-10-0-3-111",
                    "path": "/data/storage3"
                },
                {
                    "host": "ip-10-0-3-240",
                    "path": "/data/storage3"
                }
            ],
            "status": "Bound",
            "claim": {
                "name": "my-awesome-claim",
                "namespace": "default"
            }
        },
        {
            "name": "storage4",
            "capacity": "1.00GB",
            "available": "1.00GB",
            "replicaCount": 1,
            "distributionCount": 1,
            "bricks": [
                {
                    "host": "ip-10-0-3-111",
                    "path": "/data/storage4"
                }
            ],
            "status": "Bound",
            "claim": {
                "name": "another-claim-frontend1",
                "namespace": "default"
            },
            "pods": [
                {
                    "name": "another-pod",
                    "namespace": "default",
                    "containers": [
                        {
                            "name": "myfrontend",
                            "mountPath": "/var/www/html"
                        },
                        {
                            "name": "sidecar",
                            "mountPath": "/var/www/html"
                        }
                    ]
                },
                {
                    "name": "mypod-2",
                    "namespace": "default",
                    "containers": [
                        {
                            "name": "frontend",
                            "mountPath": "/var/www/html"
                        }
                    ]
                }
            ]
        },
        {
            "name": "storage5",
            "capacity": "1.00GB",
            "available": "1.00GB",
            "replicaCount": 1,
            "distributionCount": 1,
            "bricks": [
                {
                    "host": "ip-10-0-3-111",
                    "path": "/data/storage5"
                }
            ],
            "status": "Available"
        }
    ]
}
```

glide.yaml
```diff
@@ -15,7 +15,7 @@ import:
   subpackages:
   - ssh
 - package: github.com/pkg/browser
-- package: github.com/docker/machine
-  version: ^0.9.0-rc2
+- package: k8s.io/kubernetes
```
alexbrand: This is a large dependency to take on for just a couple of types. I propose either a) re-implementing the two or three types we are depending on, or b) using https://github.com/kubernetes/client-go

I think I would prefer option A given the actual dependency we have, but could be convinced otherwise.

pkg/cli/volume_list.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	pvGetter := data.PlanPVGetter{SSHClient: clientMaster}
+	podGetter := data.PlanPodGetter{SSHClient: clientMaster}
+
+	resp, err := buildResponse(glusterGetter, pvGetter, podGetter)
+	if err != nil {
+		return err
+	}
+	if resp == nil {
+		fmt.Fprintln(out, "No resources found.")
```
alexbrand: Can we have a more descriptive message here? What does this really mean? Does it mean there are no PVs on the cluster?

pkg/cli/volume_list.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	// build a map of pods that have PersistentVolumeClaim
+	podsMap := make(map[string][]volume.Pod)
+	// iterate through all the pods
+	// since the api doesnt have a pv -> pod data, need to search through all the pods
+	// this will get PV -> PVC - > pod(s) -> container(s)
+	if pods != nil { // possible for no pods to have claimed a PV
+		for _, pod := range pods.Items {
+			if len(pod.Spec.Volumes) > 0 {
```
alexbrand: Is this check necessary? The `for` should take care of this by not doing anything if the collection is empty.

pkg/cli/volume_list.go
```diff
@@ -0,0 +1,199 @@
...
+				for _, v := range pod.Spec.Volumes {
+					if v.PersistentVolumeClaim != nil {
+						var containers []volume.Container
+						for _, container := range pod.Spec.Containers {
+							for _, volumeMount := range container.VolumeMounts {
+								if volumeMount.Name == v.Name {
+									containers = append(containers, volume.Container{Name: container.Name, MountName: volumeMount.Name, MountPath: volumeMount.MountPath})
```
alexbrand: Why are we going as deep as `containers`, instead of just `pods`? Is it because each container can mount the volume on a different point?

pkg/cli/volume_list.go
```diff
@@ -0,0 +1,199 @@
...
+	// iterate through PVs once and build a map
+	pvsMap := make(map[string]*v1.PersistentVolume)
+	if pvs != nil {
+		for _, pv := range pvs.Items {
+			pvsMap[pv.GetName()] = &pv
+		}
+	}
+
+	// build response object
+	var resp = volume.ListResponse{}
```
alexbrand: consistency: `resp :=`

pkg/cli/volume_list.go
```diff
@@ -0,0 +1,199 @@
...
+	// loop through all the gluster volumes
+	for _, gv := range glusterVolumeInfo.VolumeInfo.Volumes.Volume {
+		var v = volume.Volume{}
```
alexbrand: consistency: `v :=`

alexbrand: It also seems like some of the initial values can be set when building the struct

pkg/data/gluster.go
```diff
@@ -0,0 +1,52 @@
...
+		return nil, fmt.Errorf("error getting volume info data: %v", err)
+	}
+	glusterVolumeInfoRaw = strings.TrimSpace(glusterVolumeInfoRaw)
+	var glusterVolumeInfo GlusterVolumeInfoCliOutput
+	err = xml.Unmarshal([]byte(glusterVolumeInfoRaw), &glusterVolumeInfo)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshalling volume info data: %v", err)
+	}
+	if &glusterVolumeInfo == nil || glusterVolumeInfo.VolumeInfo == nil || glusterVolumeInfo.VolumeInfo.Volumes == nil || glusterVolumeInfo.VolumeInfo.Volumes.Volume == nil || len(glusterVolumeInfo.VolumeInfo.Volumes.Volume) == 0 {
+		return nil, nil
```
alexbrand: Are all of these cases acceptable outcomes? Some of these seem like errors (e.g. `glusterVolumeInfo == nil`)

pkg/ssh/ssh.go
```diff
@@ -5,38 +5,111 @@ import (
...
+	"-o", "PasswordAuthentication=no",
+	"-o", "StrictHostKeyChecking=no",
+	"-o", "UserKnownHostsFile=/dev/null",
+	"-o", "LogLevel=quiet", // suppress "Warning: Permanently added '[localhost]:2022' (ECDSA) to the list of known hosts."
+	"-o", "ConnectionAttempts=3", // retry 3 times if SSH connection fails
+	"-o", "ConnectTimeout=10", // timeout after 10 seconds
+	"-o", "ControlMaster=no", // disable ssh multiplexing
+	"-o", "ControlPath=none",
+	"-t",
+	"-t",
```
alexbrand: Is there a reason why the libmachine SSH client doesn't force TTY allocation? I wonder if there are any security or compatibility concerns here.

pkg/volume/formatter.go
```diff
@@ -0,0 +1,77 @@
...
+	case bytes >= MB:
+		return fmt.Sprintf("%.2fMB", bytes/MB)
+	case bytes >= KB:
+		return fmt.Sprintf("%.2fKB", bytes/KB)
+	}
+	return fmt.Sprintf("%.2fB", bytes)
+}
+
+// Print prints the volume list response
+func Print(out io.Writer, resp *ListResponse, format string) error {
```
alexbrand: This printing seems like a `cli` package concern?

pkg/data/gluster.go
```diff
@@ -0,0 +1,52 @@
+package data
```
alexbrand: It's unclear to me what should live in this package, and what the difference is between `data` and `volume`.

glide.yaml
dkoshkin: Hmm yeah I think you are right, I will work on reimplementing the structs

pkg/cli/volume_list.go
dkoshkin: This is consistent with the k8s cli.
In this case it means there were no gluster volumes.

pkg/cli/volume_list.go
dkoshkin: Exactly, the idea here is when debugging the user would see where the volumes are mounted in the containers and the actual directory

pkg/data/gluster.go
dkoshkin: I removed the volume package since it didn't really make sense having it there.
The `data` package is meant for interacting with the cluster and retrieving data, `volume list` for now but also for future work.

pkg/cli/volume_list.go
dkoshkin: The `pods` slice will be nil if there are no pods

dkoshkin: Nvm I will remove that

alexbrand:
Is there anything that can be done to reduce the size of the input data in the tests? I feel like maintaining that will be difficult. Would removing data we don't use be viable?

pkg/cli/volume_list.go
alexbrand: Got it. Is there a reason not to print something like:
```
No volumes were found on the cluster. You may use `kismatic volume add` to create new volumes.
```
Or something of the sort...

pkg/cli/volume_list.go
dkoshkin: Nope just realized that pods is a wrapper object not a slice so it can be nil

alexbrand: `pods` is, but do we need the check on `110`?

pkg/data/pod.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/apprenda/kismatic/pkg/ssh"
+)
+
+type PodGetter interface {
+	Get() (*PodList, error)
```
alexbrand: Thoughts on using List() instead of Get() here? And calling the interface `PodLister`?

pkg/data/pv.go
```diff
@@ -0,0 +1,37 @@
...
+	"strings"
+
+	"github.com/apprenda/kismatic/pkg/ssh"
+)
+
+type PVGetter interface {
+	Get() (*PersistentVolumeList, error)
+}
+
+type PlanPVGetter struct {
```
alexbrand: Seems like this doesn't depend on a plan? Thoughts on `KubectlPVGetter`?

pkg/data/kubernetes_type.go
```diff
@@ -0,0 +1,193 @@
+package data
+
+import "k8s.io/kubernetes/pkg/api/unversioned"
```
alexbrand: Looks like we are still pulling in k8s as a dependency

pkg/data/pv.go
dkoshkin: I want to emphasize thats its not a local `kubectl` install, `RemoteKubectlPVGetter` or is that too wordy?

pkg/data/pod.go
dkoshkin: I like it

pkg/data/pv.go
alexbrand: Yeah, that makes sense to me. I like that name.

An alternative that might be worth entertaining is to have a `RemoteKubectl` or `SSHKubectl` that implements both `PVLister` and `PodLister`? The method names would have to change to something like `ListPods` and `ListPersistentVolumes`. Haven't given much thought to this though.

pkg/cli/volume_list.go
```diff
@@ -0,0 +1,262 @@
...
+		Short: "list storage volumes to the Kubernetes cluster",
+		Long: `List storage volumes to the Kubernetes cluster.
+
+This function requires a target cluster that has storage nodes.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return doVolumeList(out, opts, *planFile, args)
+		},
+	}
+
+	//cmd.Flags().StringVarP(&opts.namespace, "namespace", "ns", "all", `limit output to a single namespace`)
```
alexbrand: Do we want this flag, or can we remove this comment?

alexbrand:
I suggest cleaning up the commit history, but otherwise LGTM! 👏

dkoshkin:
This was a biggie, I'm going to squash into a single commit when I merge as there are no cleaner way unfortunately :(

PR #253, opened 2017-01-23
Cleanup, tag and refactor integration tests

alexbrand:
Cleaned up and organized our current test corpus. This PR also introduces a convention for tagging tests. Three tags are being proposed:
* `[slow]`: Tests that take approx. 15 minutes or longer, which is about half our tests
* `[aws]`: Tests that run on AWS infrastructure
* `[packet]`: Tests that run on Packet infrastructure

The tags can be selected by passing the `--focus` or `--skip` flag to `ginkgo`. 

The total number of tests have been reduced to 31 from 35. The 4 tests that were removed due to little or zero value:
* Running validation when package installation is enabled on CentOS
* Running validation when package installation is enabled on RHEL
* Running validation when package installation is enabled in Ubuntu (This and above are redundant, as validation is part of installation).
* Installing with package installation disabled (This is covered by the disconnected install test)

This PR also proposes the use of feature-specific source files and test files for organizing the suite. Specs should live in `xxxxx_test.go` files, while anything that is not a spec should live inside `xxxxx.go` files. For example, the ingress related code lives in `ingress.go` and the ingress specs live in `ingress_test.go`.


integration/step_test.go
```diff
@@ -25,18 +26,19 @@ var _ = Describe("install step commands", func() {
...
 			})
 		})
-		Context("Running the smoketest play against an existing cluster", func() {
-			ItOnAWS("should return successfully", func(provisioner infrastructureProvisioner) {
-				WithMiniInfrastructure(CentOS7, provisioner, func(node NodeDeets, sshKey string) {
-					err := installKismaticMini(node, sshKey)
-					Expect(err).ToNot(HaveOccurred())
+	})
+
+	Describe("Running the smoketest play against an existing cluster", func() {
```
dkoshkin: Does this provide enough value to run on new cluster, can it be removed or move to the previous `Running the api server play against an existing cluster` test.

This was originally added when there was a bug with the smoketest.

integration/storage_test.go
```diff
@@ -4,80 +4,83 @@ import (
...
-	Context("Using a cluster with storage nodes configured", func() {
-		ItOnAWS("a stateful workload should be able to read/write state to a persistent volume", func(aws infrastructureProvisioner) {
-			WithInfrastructure(NodeCount{Etcd: 1, Master: 1, Worker: 2, Storage: 2}, CentOS7, aws, func(nodes provisionedNodes, sshKey string) {
-				By("Installing a cluster with storage")
-				opts := installOptions{
-					allowPackageInstallation: true,
-				}
-				err := installKismatic(nodes, opts, sshKey)
-				FailIfError(err, "Installation failed")
+	Describe("deploying a stateful workload", func() {
```
dkoshkin: Capitalize

integration/step_test.go
alexbrand: I am fine with removing the test.

dkoshkin:
LGTM

PR #258, opened 2017-01-23
Use labels when creating PVs

### LINKED ISSUE #249: No way to specify storage requirements in PVC
We need to support selectors in PVCs to allow for binding to PVs with a specific characteristics, ie `replication` and `distribution` counts.

dkoshkin:
Fixes #249 

Added new flag 
```
cmd.Flags().StringVarP(&opts.storageClass, "storage-class", "c", "kismatic", "The StorageClass to present for claims in Kubernetes. Classes should identify properties of volumes in business terms, such as 'durable' or 'fast-reads'")
```

This will set the `annotation` when creating PVs
```
  annotations:
    volume.beta.kubernetes.io/storage-class: "{{ volume_storage_class }}"
```


ansible/roles/addon-nfs-volume/templates/add-persistent-volume.yaml
```diff
@@ -2,6 +2,9 @@ apiVersion: v1
 kind: PersistentVolume
 metadata:
   name: pv{{ index }}
+  labels:
+    kismatic-volume-replica: "1"
```
alexbrand: Are these labels true? We don't really know how the storage is setup. I think I'd propose not to label the PVs added via NFS shares.

dkoshkin: Good point

alexbrand:
Ship it!

Issue #252, opened 2017-01-23
Kismatic should record the "version" of Kismatic used to install/upgrade each node


dasmb:
Possible locations for the version:
/etc/kismatic-version

Should be recorded after successful install/upgrade of the node, prior to execution of cluster level feature installs or smoke tests.

PR #260, opened 2017-01-26
Introduce kismatic versions to nodes

dasmb:
Re: https://github.com/apprenda/kismatic/issues/252.

I also added functionality to list nodes with their versions and their declared roles.


pkg/cli/list.go
```diff
@@ -0,0 +1,64 @@
...
+type listOpts struct {
+	planFilename string
+}
+
+// NewCmdSSH returns an ssh shell
+func NewCmdList(out io.Writer) *cobra.Command {
+	opts := &listOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "list",
```
dkoshkin: I'm not sure if `list` is a good verb for this, seems pretty ambiguous to what its going to list.

dkoshkin: Maybe `cluster-info`

dasmb:
Closing until upgrades are ready for delivery to master.

PR #261, opened 2017-01-27
Ping repo on one node, all or nothing

### LINKED ISSUE #241: Installer still doesn't exit on an repo ping error
Need an "all or nothing" condition on the `ping Kismatic deb repository` task, if some succeed but other nodes fail it continues to run and skips the next few tasks for the nodes that failed the ping.

Can probably be solved by adding a `run_once` to that task.

```
- Running task: ping Kismatic deb repository
  worker001                                                                     [OK]
  master001                                                                     [ERROR IGNORED]
  etcd001                                                                       [ERROR IGNORED]
- Running task: install apt-transport-https package
  master001                                                                     [SKIPPED]
  etcd001                                                                       [SKIPPED]
  worker001                                                                     [OK]
- Running task: add Kismatic deb key
  master001                                                                     [SKIPPED]
  etcd001                                                                       [SKIPPED]
  worker001                                                                     [OK]
- Running task: add Kismatic deb repository
  master001                                                                     [SKIPPED]
  etcd001                                                                       [SKIPPED]
  worker001                                                                     [OK]
- Running task: apt-get update
  master001                                                                     [SKIPPED]
  etcd001                                                                       [SKIPPED]
  worker001                                                                     [OK]
```

dkoshkin:
Fixes #241

### DIFF
--- a/ansible/roles/packages/tasks/main.yaml
+++ b/ansible/roles/packages/tasks/main.yaml
@@ -9,6 +9,7 @@
     retries: 3
     delay: 1
     changed_when: "result.status is defined and result.status == 200"
+    run_once: true
 
   # only configure repository if it is reachable
   - name: add Kismatic yum repository
@@ -33,6 +34,7 @@
     retries: 3
     delay: 1
     changed_when: "result.status is defined and result.status == 200"
+    run_once: true
 
   # only configure repository if it is reachable
   - name: install apt-transport-https package


alexbrand:
Something seems off to me with this play. I see two potential outcomes from the current way the play is set up:
* Ping fails: This means that we won't add the repositories, but we will attempt to install the packages anyway. What if the user is trying to do an online install, but for some reason is unable to connect to our repo?
* Ping succeeds: This means that we add the repositories, and we install the packages. (Less interesting case)

The problem lies in that we don't know if the ping failure is acceptable or not. I think what is missing here is the ability to tell whether we are doing an online or an offline install. If we have that boolean available, I think we would have a better experience:

* Online install: If the ping fails, the play fails. 
* Offline install: We don't attempt to ping, and also don't add our repo.

dkoshkin:
Yeah I think we will need a boolean for the changes to our packaging we discussed yesterday. Going to close this for now until we have a clearer idea.

PR #263, opened 2017-01-27
Deploy gluster healthz daemonset

### LINKED ISSUE #248: No liveness check on storage endpoint
In Kubernetes, services that are populated manually via the creation of endpoints do not have any concept of liveness checks.

In our storage setup, we use a kubernetes service as the gateway to the storage nodes of the cluster. In the case that a storage node goes down, the dead endpoint is not removed from the service. This results in the potential inability to consume the storage that might otherwise be available on other nodes.

alexbrand:
Fixes #248 

This PR proposes the deployment of a daemonset that uses the https://github.com/apprenda/tcp-healthz container to monitor the health of the gluster cluster.

To do:
- [x] Tag the docker image with a version, and update the daemonset spec once tests pass.


dkoshkin:
LGTM, waiting on the new Kismatic packages.

alexbrand:
Failure is flake. Merging.

PR #267, opened 2017-01-30
Add validation for duplicate IP/Hostname inside the same node group

### LINKED ISSUE #264: Preflight checks should check for duplicate hostnames and ip addresses
The execution of KET has corrupted a few times when I have had duplicate hostnames and/or ip addresses. The tool should check for these before running any of the ansible playbooks.

alexbrand:
Fixes #264 

Sample output:
```
./kismatic install validate

- Storage nodes: Node #2 has the same hostname "node001" as node #1
- Storage nodes: Node #2 has the same IP "192.168.42.2" as node #1
```


dasmb:
LGTM

PR #268, opened 2017-01-31
New disconnected_installation flag

### LINKED ISSUE #262: Improve install times
Currently even for a small cluster (3 etcd, 2 masters and 2 workers) the install time is about 15 minutes.
There are 2 plays that take up a significant portion of the install time:
* `Install Kismatic packages`
* `Install Docker -> load local docker images` 

Both of these are caused by the same underlying factor: the inclusion of docker images in our kismatic packages.

The contents of the k8s packages: `images` contain the docker images and are about 60% of the total size.
```
 27M	docker
 53M	etcd
905M	images
459M	kubernetes
 85M	networking
```

By including the images in the packages:
1. Those images are downloaded from a slow S3 source to each node
2. Cannot take advantage of docker caching and layers

The initial reasoning for this technique was to support offline installs. We can improve the performance by treating `offline` as an optional feature with a separate k8s package containing all the bits needed to support that offline workflow. 

**TODO:**
- [x] Create new package with the needed bits (docker images)
- [x] Modify `kismatic` CLI for standard/online flow
        * Don't load docker images from a local dir
        * Method to determine if the install is online or offline -> this will drive the url for the docker images used during the install: new plan file variable
- [x] Modify `kismatic` CLI for offline flow
        * Install new package
        * Load docker images in the local docker registry
        * Use the local docker registry url
- [x] Modify `kuberang` to accept an optional parameter `--docker-registry-url`

dkoshkin:
Fixes #262

### DIFF
--- a/ansible/group_vars/all.yaml
+++ b/ansible/group_vars/all.yaml
@@ -170,7 +170,7 @@ kubernetes_dashboard_img: '{% if load_private_images|bool == true %}{{ kubernete
 
 apprenda_tcp_healthz_orig_img: apprenda/tcp-healthz-amd64:{{ apprenda_tcp_healthz_version }}
 apprenda_tcp_healthz_priv_img: "{{ docker_registry_address }}:{{ docker_registry_port }}/tcp-healthz-amd64:{{ apprenda_tcp_healthz_version }}"
-apprenda_tcp_healthz_img: "{% if use_private_docker_registry is defined and use_private_docker_registry|bool == true %}{{ apprenda_tcp_healthz_priv_img }}{% else %}{{ apprenda_tcp_healthz_orig_img }}{% endif %}"
+apprenda_tcp_healthz_img: "{% if load_private_images|bool == true %}{{ apprenda_tcp_healthz_priv_img }}{% else %}{{ apprenda_tcp_healthz_orig_img }}{% endif %}"
 
 # kismatic packages
 yum_repository_url: https://kismatic-packages-rpm.s3-accelerate.amazonaws.com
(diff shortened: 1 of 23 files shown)


ansible/group_vars/all.yaml
```diff
@@ -122,33 +120,53 @@ kubernetes_certificates_service_account_key_path: "{{ kubernetes_install_dir }}/
...
+exechealthz_priv_img: "{{ docker_registry_full_url }}/exechealthz-amd64:{{ exechealthz_version }}"
+exechealthz_img: '{% if load_private_images|bool == true %}{{ exechealthz_priv_img }}{% else %}{{ exechealthz_orig_img }}{% endif %}'
 
 kubernetes_dashboard_orig_img: gcr.io/google_containers/kubernetes-dashboard-amd64:{{ kubernetes_dashboard_version }}
-kubernetes_dashboard_priv_img: "{{ docker_registry_address }}:{{ docker_registry_port }}/kubernetes-dashboard-amd64:{{ kubernetes_dashboard_version }}"
-kubernetes_dashboard_img: '{% if use_private_docker_registry is defined and use_private_docker_registry|bool == true %}{{ kubernetes_dashboard_priv_img }}{% else %}{{ kubernetes_dashboard_orig_img }}{% endif %}'
+kubernetes_dashboard_priv_img: "{{ docker_registry_full_url }}/kubernetes-dashboard-amd64:{{ kubernetes_dashboard_version }}"
+kubernetes_dashboard_img: '{% if load_private_images|bool == true %}{{ kubernetes_dashboard_priv_img }}{% else %}{{ kubernetes_dashboard_orig_img }}{% endif %}'
 
 apprenda_tcp_healthz_orig_img: apprenda/tcp-healthz-amd64:{{ apprenda_tcp_healthz_version }}
```
alexbrand: Do we need changes in line 173? (GitHub won't let me comment on it)

dkoshkin: Ahh good catch forgot during the merge.

alexbrand:
LGTM!

Issue #225, opened 2017-01-06
Support DNS names instead of IP addresses in kismatic-cluster.yaml


notuscloud:
Hi guys, 

Is it possible to only use DNS names for my hosts in the `kismatic-cluster.yaml` file?
My servers are automatically provisioned by terraform over a DHCP network. I only rely on DNS names to reach those machines.

When I try to valide my kismatic plan, I fall upon those errors:
```
Validating==========================================================================
Reading installation plan file "kismatic-cluster.yaml"                          [OK]
Validating installation plan file                                               [ERROR]
- SSH Key field must be an absolute path
- Etcd nodes: Node #1: Invalid IP provided
- Master nodes: Node #1: Invalid IP provided
- Master nodes: Load balanced FQDN is required
- Master nodes: Load balanced shortname is required
- Worker nodes: Node #1: Invalid IP provided
- Worker nodes: Node #2: Invalid IP provided
error validating plan: validation error prevents installation from proceeding
```

Which is quite normal cause I didn't provided any IP addresses...

Regards,
Thomas

dasmb:
I have a few quick questions for you.
1) Are your IP addresses stable after provisioning?
2) Is there more than one address attached to each host?
3) Does the IP address map to the same hostname everywhere? (e.g. can you `ping machine01` from machine01 and get the same IP you get when you ping it from your installation machine)

The reason Kismatic separates hostname, IP and internalIP into three different sections is that different tools need to access machines in different ways -- e.g. some of the Calico tools have _(or rather had, I think this has been relaxed)_ a requirement that the communicate over hostname, and some other tools in the K8s stack had requirements that they use IP addresses, with the internal IP address being a pretty substantial optimization for routing internal cluster traffic.

If your IP addresses aren't stable, when you obtain a new address for a node that node will lose identity on the cluster. And if your hostnames aren't the same both inside and out, some components just flat out won't work correctly.

If your hostnames always map to the same address, inside and out, and only that address matters for purposes of routing, and it's stable, we could introduce functionality to automatically fill in the IP address from the hostname that would meet your needs. It may also encourage people with more complicated network and naming rules to set their clusters up in a way that is likely to be unstable.

notuscloud:
Hi, 

1 - My IPs are stable, we have 24h DHCP lease and VMs keep their IP upon renewal. the only case an IP is changed would a more than 24h downtime for a VM.

2 - Each host comes with only a single nic, with a single IP

3 - The IP map the host name everywhere.
[...]

PR #272, opened 2017-02-03
[WIP] Use static pods

dkoshkin:
Converting kube control plane to static pods.

### Previous Install
* Used kismatic built OS packages for `etcd` `kubelet` `kube-apiserver` `kube-proxy` `kube-controller-manager` `docker` `networking` and `offline` 
* The packages contained binaries repackaged from Gihub and Google
* All components managed by systemd
* Unit files are written at runtime with required variables and services are started

In order to better align with the k8s community a new approach of running the pods as static pods was taken.

### Benefits
* Rely less on custom OS packages
* Utilize official upstream docker images for `kube-apiserver` `kube-proxy` `kube-controller-manager` and `calico` 
* Use k8s to manage and keep services alive
* BYO `docker` `kubelet` and `kubectl` should now work but not tested
### What Changed
* Use static pods
  * Added 2 new flags to the api-server `--anonymous-auth=false` this was added in `v1.5.1` and `--kubelet-preferred-address-types` this flag allows to drop `--node-ip` from the kubelet and support https://github.com/apprenda/kismatic/issues/225 -  this PR did NOT remove `--node-ip` from the kubelet config 
  * Deploy `Calico` as pods, this is the recommended by them.
* Only build `kubelet` `kubectl` `docker-engine` `etcd` `offline` packages. Packages were `renamed` to better align with the community. 
  * Packages used by `kubeadm` are lagging in versions and not mature enough to use
  * Doing `yum install docker` or `apt-get install docker.io` will install random versions of docker-engine, this will work but it is recommended to install the version from the kismatic repo.
* Move cert generation plays up, this avoids having to restart docker multiple times if using a registry
* Rename plays to better organize, ie. prefix `kube-...`, drop `add-on` if its not add-on etc

### New workflow
1.  Prereqs
2. Modify hosts file - if needed
3. Move certs
4. Install packages - if needed
5. Start docker and verify its running
6. Start docker registry - if `setup_internal_docker_registry`
7. Load docker images from `offline` package into registry - `use_private_docker_registry == true and disconnected_installation== true `
8. Copy kubelet unit fille, restart and confirm its running - at this point kubelet will be keep trying to connect to the API server (the service will not fail here)
9. Download `kube-apiserver` (gcr or local registry) copy to `/manifests` dir
10. ... `kube-scheduler`
11. ... `kube-controoler-manager`
12. ... `kube-proxy` 
13. Download `calico` pods, wait for API server to be up and `kubectl apply` - calico runs a `job` to copy the required CNI binaries from its own docker container and sets up the cni config file
14. Wait for all k8s controlplane + calico pods to be up -`validate-controlplane`
15. Config DNS
16. Ingress
17. Dashboard
18. Network Policy
19. Config storage
20. Config NFS


ansible/roles/kubelet/tasks/main.yaml
```diff
@@ -2,22 +2,25 @@
...
+  - meta: flush_handlers  #Run handlers
 
   # force_kubelet_restart=true to force restart
   - name: force restart kubelet
-    command: /bin/true
-    notify:
-      - enable kubelet
-      - restart kubelet service
-      - verify kubelet is running
+    command: systemctl restart kubelet.service
```
alexbrand: In some places we use the service task to do this. Is one better than the other? I suggest we keep them consistent.

ansible/roles/kubelet/tasks/main.yaml
```diff
@@ -2,22 +2,25 @@
...
       owner: "{{ kubernetes_owner }}"
       group: "{{ kubernetes_group }}"
       mode: "{{ kubernetes_service_mode }}"
     notify:
       - reload services
-      - enable kubelet
       - restart kubelet service
-      - verify kubelet is running
+
+  - meta: flush_handlers  #Run handlers
```
alexbrand: Do we need the handlers? Seems like we could move the start kubelet into the play?

ansible/roles/packages/tasks/main.yaml
```diff
@@ -1,88 +1,84 @@
...
+    when: ansible_os_family == 'RedHat' and kismatic_result.status is defined and kismatic_result.status == 200
+  - name: clean kismatic yum cache
     command: 'yum --enablerepo=kismatic clean metadata'
-    when: ansible_os_family == 'RedHat' and result.status is defined and result.status == 200
+    when: ansible_os_family == 'RedHat' and kismatic_result.status is defined and kismatic_result.status == 200
+
+  - name: install yum packages
+    command: /bin/true
+    notify:
+      - install etcd yum package
```
alexbrand: Do we gain anything from these being handlers instead of tasks?

ansible/_validate-controlplane.yaml
```diff
@@ -0,0 +1,10 @@
+---
+  - hosts: master[0]
+    any_errors_fatal: true
+    name: "Validate Kubernetes Controlplane is Running"
```
dkoshkin: :( damn I was so confident it was a compound word

ansible/roles/calico/tasks/main.yaml
```diff
@@ -1,58 +1,46 @@
...
-    when: force_calico_node_restart is defined and force_calico_node_restart|bool == true
+    when: "inventory_hostname == groups['master'][0]"
 
-  - meta: flush_handlers  #Run handlers
+  - name: run calico containers
+    command: kubectl apply -f /etc/calico/calico.yaml
+    run_once: true
+    register: result
+    until: result|success
+    retries: 10
```
dkoshkin: The API is not guaranteed to be up yet at this point and I added this before adding the task to wait for the API server. I will remove this.

ansible/roles/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - --insecure-bind-address=127.0.0.1
+      - --bind-address=0.0.0.0
+      - --client-ca-file={{ kubernetes_certificates_ca_path }}
+      - --etcd-cafile={{ kubernetes_certificates_ca_path }}
+      - --etcd-certfile={{ kubernetes_certificates_cert_path }}
+      - --etcd-keyfile={{ kubernetes_certificates_key_path}}
+      - --etcd-servers={{ etcd_k8s_cluster_ip_list }}
+      - --insecure-bind-address=127.0.0.1
+      - --insecure-port=8080
+      - --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
```
alexbrand: Was there a reason for preferring the ExternalIP over the Hostname?

ansible/roles/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - kube-apiserver
+      - --admission-control=NamespaceLifecycle,LimitRanger,ServiceAccount,PersistentVolumeLabel,DefaultStorageClass,ResourceQuota
+      - --advertise-address={{ internal_ipv4 }}
+      - --allow-privileged=true
+      - --anonymous-auth=false
+      - --apiserver-count={{ kubernetes_master_apiserver_count }}
+      - --authorization-mode=ABAC
+      - --authorization-policy-file={{ kubernetes_authorization_policy_path }}
+      - --basic-auth-file={{ kubernetes_basic_auth_path }}
+      - --insecure-bind-address=127.0.0.1
```
alexbrand: Duplicate flag

ansible/roles/kubelet/tasks/main.yaml
dkoshkin: Yes we do, when the play is rerun we dont want this to trigger unless the unit file changed

ansible/roles/kubelet/templates/kubelet.service
```diff
@@ -34,7 +36,8 @@ ExecStart=/usr/bin/kubelet \
   --network-plugin-dir=${NETWORK_PLUGIN_DIR} \
   --v=2
 Restart=on-failure
-RestartSec=3
+RestartSec=10
```
alexbrand: Was there a reason for increasing the restart latency?

ansible/roles/kubelet/tasks/main.yaml
alexbrand: Aaah got it! The notifications are dispatched only when the task "changed". Thanks!

ansible/roles/kube-apiserver/templates/kube-apiserver.yaml
dkoshkin: I'm going to remove this flag as more testing is needed around this. This is currently unnecessary as we have a mechanism to always use resolvable hostnames.

ansible/roles/kubelet/templates/kubelet.service
dkoshkin: Not anymore since I went back to using binaries, will change.

ansible/roles/packages/tasks/main.yaml
dkoshkin: Just not having to repeat `ansible_os_family ==` and grouping the tasks together. That play is already pretty crowded.

ansible/roles/calico/tasks/main.yaml
```diff
@@ -1,58 +1,42 @@
...
+      - "{{ calico_node_img }}"
+      - "{{ calico_cni_img }}"
+      - "{{ calico_ctl_img }}"
 
   - name: create /etc/calico directory
     file:
       path: "{{ calico_dir }}"
       state: directory
+
   - name: copy calicoctl.cfg to remote
```
alexbrand: Is this file used in this new deployment method?

dkoshkin: Yes, thats where the Calico yaml file is saved

docs/PROVISION.md
```diff
@@ -135,25 +135,25 @@ If you are building a large cluster or one that won't have access to these repos
     <td>yes</td>
   </tr>
   <tr>
-    <td>Kismatic package of Etcd 2.3.7 and 3.0.15</td>
```
alexbrand: Is there a reason for downgrading etcd to v3.0.10 from v3.0.15?

alexbrand: Oh, from the test below, it looks like we are installing v3.1.0. Typo here?

alexbrand:
Needs squash.

I reviewed a couple of times and don't see issues with it. Let's do a final discussion with team and make a decision.

docs/PROVISION.md
dkoshkin: Thanks, yep will change now.

PR #273, opened 2017-02-03
Update to ansible 2.2.1.0

alexbrand:
It's possible that some of the slowness we are seeing with ansible is due to a regression in ansible 2.1. I am basing this on https://bugzilla.redhat.com/show_bug.cgi?id=1360440

Opening this PR to see if we experience any drastic improvement in the test runs.


alexbrand:
There was no significant improvement to the installation time. I am closing this for now, as the tests discovered that some parts of our playbooks are not compatible with the new version of ansible.

See tests for more details.

PR #274, opened 2017-02-06
Add NFS volume validation

alexbrand:
Given that the default plan file includes a bad NFS volume, the installation fails at a late stage, resulting in confusion on the user's end. 

To fix this, I added validation to the NFS volume and  also included a new question in the planning stage:
```
Number of existing NFS volumes to be attached
```

### DIFF
--- a/integration/install_test.go
+++ b/integration/install_test.go
@@ -42,6 +42,7 @@ var _ = Describe("kismatic", func() {
 				Expect(helpText).To(ContainSubstring("3 worker nodes"))
 				Expect(helpText).To(ContainSubstring("2 ingress nodes"))
 				Expect(helpText).To(ContainSubstring("0 storage nodes"))
+				Expect(helpText).To(ContainSubstring("0 nfs volumes"))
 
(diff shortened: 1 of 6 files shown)


dkoshkin:
Is this right? https://github.com/apprenda/kismatic/blob/nfs-validation/pkg/install/plan.go#L177

alexbrand:
I think we need a way to let the user know we support NFS volumes. We can either keep doing this, which will now be caught by validation, or we can add the same block but commented out. 

I am torn between both approaches. The second will most likely result in a stale comment if we ever change that struct, but the first can be considered unfriendly UX.

dkoshkin:
On your second option I dont think this is unique just to the NFS, if any struct changes it will be stale. IMO I think it should be commented out.

@dasmb what are your thoughts since you originally added this?

alexbrand:
I am fine with both, really. Although I don't think we have precedent for commented out structs, do we? Another alternative would be to ask the user for the number of NFS shares in `./kismatic install plan`

dkoshkin:
I like the approach of asking, that works well with our existing UX.

dkoshkin:
LGTM

Issue #255, opened 2017-01-23
Unfiltered Offline upgrade


dasmb:
As a first step to all other forms of upgrade, we will add a command to kismatic that will allow for a cluster to be upgraded, one node at a time, with no draining or analysis of safety.

`kismatic upgrade offline [-f plan=kismatic-cluster.yaml]`

This process should start by detecting the current Kismatic version (ssh, grab /etc/kismatic-version). For each node, if that node is below the current version of the installer, then the node qualifies for upgrade. Nodes that do not qualify should be logged as skipped with their current version displayed. If no nodes needing upgrading are detected, the process should exit.

For the nodes that qualify to be upgraded, we will run a special form of validation against them to be sure they are ready. Readiness is a function primarily of disk space on the root volume and an ability to access the upgrade packages via existing repositories. If any qualifying nodes are unready, the process should exit, listing the nodes which are not ready.

Upgrades should occur, one node at a time, in the following order: 

1. All worker nodes (regardless of specialization)
2. All master nodes
3. All etcd nodes
4. “On-cluster” Docker Registry
5. Other on-cluster systems (DNS, dashboard, etc)

No draining or cordoning will be done. This may result in data loss or a reduction in availability. That's what makes this the "offline" upgrade.

alexbrand:
In terms of the upgrade order, I believe we want to upgrade the K8s control plane components before the worker components. K8s guarantees new API server -> old Kubelet compatibility, but not the other way around. 

I am also unsure about whether we should run ansible once against the entire inventory, or run ansible multiple times, one node at a time? If running against the whole inventory, we specify `serial: 1`, but I am not sure if that's enough.

alexbrand:
Initial implementation of this was completed in #275

PR #275, opened 2017-02-06
[WIP] Offline upgrade

alexbrand:
Initial pass at implementing offline upgrades. The upgrade follows the process outlined in #255. That is, the upgrade proceeds, one node at a time. Once the nodes are upgraded, the private registry and cluster services are upgraded next.

This PR also attempts to DRY up the `execute.go` file. I think it's a first step towards a refactor of this code.

The following is the output for a 3 node cluster on AWS:
```
> ./kismatic upgrade offline

Computing upgrade plan==============================================================
Validate SSH connectivity to nodes                                              [OK]

Validating nodes====================================================================
1/1....  Run Upgrade Pre-Flight Checks                                          [OK]

Upgrade Node "ip-10-0-3-105"========================================================
1/11...  Install Kismatic packages                                              [OK]
2/11...  Upgrade Kubernetes Etcd Cluster                                        [OK]
3/11...  Upgrade Networking Etcd Cluster                                        [OK]
4/11...  Upgrade Docker                                                         [OK]
5/11...  Upgrade Cluster Network                                                [OK]
6/11...  Update Kismatic version file                                           [OK]

Upgrade Node "ip-10-0-3-69"=========================================================
1/11...  Install Kismatic packages                                              [OK]
2/11...  Upgrade Kubernetes API Server                                          [OK]
3/11...  Upgrade Kubernetes Scheduler                                           [OK]
4/11...  Upgrade Kubernetes Controller Manager                                  [OK]
5/11...  Upgrade Docker                                                         [OK]
6/11...  Upgrade Cluster Network                                                [OK]
7/11...  Upgrade Kubelet                                                        [OK]
8/11...  Upgrade Kubernetes Proxy                                               [OK]
9/11...  Update Kismatic version file                                           [OK]

Upgrade Node "ip-10-0-3-200"========================================================
1/11...  Install Kismatic packages                                              [OK]
2/11...  Upgrade Docker                                                         [OK]
3/11...  Upgrade Cluster Network                                                [OK]
4/11...  Upgrade Kubelet                                                        [OK]
5/11...  Upgrade Kubernetes Proxy                                               [OK]
6/11...  Update Kismatic version file                                           [OK]

Upgrade Cluster Services============================================================
1/3....  Upgrade Kubernetes DNS                                                 [OK]
2/3....  Upgrade Kubernetes Ingress                                             [OK]
3/3....  Upgrade Kubernetes Dashboard                                           [OK]

Upgrade complete
```

Outstanding questions / challenges:
- [x] Play reuse? I don't know if we can reuse the play files, given that during the upgrade, the plays should have different names.
 - There might be a way forward here with `vars`
- [ ] Upgrade with internal docker registry - Since we introduced the disconnected installation flag with default = false, upgrading from a previous version that used an internal docker registry fails because `disconnected_installation: false` and the images are not available to be pushed into the registry.
- [x] Integration with changes to deployment approach (using pod specs)
- [ ] Figure out play numbering in output
 - It seems to me that we will have to remove the play numbering from the output. It is giving us trouble in various places and there doesn't seem to be a way around it.
- [x] Figure out why the docker and cluster network plays are running for etcd nodes
 - It turns out that this is because these plays are missing `serial: 1`. We'll probably run into this issue somehow, so this might require work in the explainer.
- [x] Run smoke test after upgrade?
- [x] How can we support upgrading everything, but only upgrade what actually needs to be upgraded? For example, right now, we are not upgrading to a new version of docker, but we are still restarting the docker service.
 - We will attempt to use ansible's constructs to detect "changes", and only perform an operation when required. For example, we can setup up the docker  play to only restart docker if a new docker package was installed.
- [x] How can we support multiple "from" versions? We'll need some kind of way to clean up stuff that is version specific. E.g. Clean up the systemd units when upgrading to a different deployment method.
 - We will not get into supporting "changeset" style upgrades at this point in time. Instead, we will ensure that the intent is met.
- [x] How to support upgrading worker nodes in parallel? A cluster with a large number of worker nodes will take a long time to upgrade.
 - Not worrying about this for now. There is nothing wrong with a long-running upgrade that is reliable.


ansible/roles/upgrade-preflight/tasks/main.yaml
```diff
@@ -0,0 +1,40 @@
...
+        service:
+          name: kismatic-inspector.service
+          state: stopped
+    always:
+      - name: stop kismatic-inspector service
+        service:
+          name: kismatic-inspector.service
+          state: stopped
+
+  - name: verify Kismatic Inspector succeeded
```
dkoshkin: I noticed that the verify step doesn't run on a failure, to fix that I moved that task up into the `rescue` and `always` blocks

ansible/roles/upgrade-preflight/templates/kismatic-inspector.service.j2
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Kismatic Inspector
+Documentation=https://github.com/apprenda/kismatic
+
+[Service]
+User=root
+ExecStart={{ bin_dir }}/kismatic-inspector server --node-roles {{ group_names|join(",") }} --port 8888 {{ (allow_package_installation|bool == true) | ternary('', '-e') }}
```
dkoshkin: Do you need to include the new `--disconnected-installation` flag?

ansible/upgrade-cluster-services.yaml
```diff
@@ -0,0 +1,26 @@
+---
```
dkoshkin: Probably need calico policy in here also.

ansible/upgrade-cluster-services.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - name: "Upgrade Kubernetes DNS"
+    hosts: master[0]
+    any_errors_fatal: true
+    remote_user: root
+    become_method: sudo
+    roles:
+      - addon-kubernetes-dns
```
dkoshkin: Currently the pod name for DNS is `kube-dns-{{kubedns_version_id}}`, I think `kubedns_version_id` should be removed to avoid having multiple versions running.

We also deploy it as a single RC instance, I wonder if we should use a `Deployment` with `replicas: 2` and use the k8s process to do a rolling upgrade.

Same thing would apply to calico policy

ansible/upgrade-docker-registry.yaml
```diff
@@ -0,0 +1,15 @@
...
+    roles:
+      - role: docker-registry-container
+
+  - name: Load Docker Images Into Private Registry
+    hosts: master[0]
+    remote_user: root
+    become_method: sudo
+    run_once: true
+    roles:
+      - role: docker-registry-images
```
dkoshkin: These images should only be loaded in an offline mode.

There is also a distinction for an internal registry and using a user provided registry. 
`plan.DockerRegistry.SetupInternal` but when offline the images should still be loaded.

ansible/upgrade-nodes.yaml
```diff
@@ -0,0 +1,114 @@
...
+  - name: Upgrade Docker
+    hosts: master:worker:ingress:storage
+    any_errors_fatal: true
+    remote_user: root
+    become_method: sudo
+    vars_files:
+      - group_vars/all.yaml
+    roles:
+      - role: docker
+      - role: docker-registry-cert
```
dkoshkin: I don't think this is needed as the certs shouldn't change.

alexbrand:
@dkoshkin Thanks for the review. I addressed your comments, minus the DNS which is outstanding. PTAL

pkg/cli/upgrade.go
```diff
@@ -0,0 +1,143 @@
...
+			return fmt.Errorf("Upgrade preflight check failed: %v", err)
+		}
+	}
+
+	// Run the upgrade on the nodes that need it
+	if err := executor.UpgradeNodes(*plan, toUpgrade); err != nil {
+		return fmt.Errorf("Failed to upgrade nodes: %v", err)
+	}
+
+	if plan.DockerRegistry.SetupInternal {
```
dkoshkin: Do we actually want to upgrade the running docker registy?

I suggest we just run the play to load docker images, in either case this condition would need to be something like `(plan.DockerRegistry.SetupInternal or plan.DockerRegistry.Address != "")  and !p.Cluster.DisconnectedInstallation` , similar to what we do in execute.

Also for this work properly `use_private_docker_registry` `disconnected_installation` `setup_internal_docker_registry` all need to be set.

alexbrand: What if we decide to move to another registry? I think we do want to do this. But this opens up an interesting question: how can we only do work when there's actual work to do? For example, right now, we are not modifying the registry, but we are still "upgrading" it.

ansible/upgrade-cluster-services.yaml
alexbrand: Addressed in master.

alexbrand:
We have decided to merge this branch and perform follow up work in subsequent PRs.

PR #277, opened 2017-02-07
Use Wants instead of Requires for systemd services

### LINKED ISSUE #276: Systemd units have strict dependencies between them
The current systemd units are using the `Requires` directive to set up dependencies between them. In the case of Kubelet requiring Docker, this results in the Kubelet service being stopped whenever the Docker service is. The same applies to the calico-node service.

Instead of using `Requires`, we should use `Wants` to setup the dependencies to avoid this. This is also the recommended way of setting up dependencies according to the systemd documentation:
> Wants= : A weaker version of Requires=. Units listed in this option will be started if the configuring unit is. However, if the listed units fail to start or cannot be added to the transaction, this has no impact on the validity of the transaction as a whole. This is the recommended way to hook start-up of one unit to the start-up of another unit.

alexbrand:
Fixes #276

### DIFF
--- a/ansible/roles/calico/templates/calico-node.service.j2
+++ b/ansible/roles/calico/templates/calico-node.service.j2
@@ -1,7 +1,7 @@
 [Unit]
 Description=Calico per-node agent
 Documentation=https://github.com/projectcalico/calico-docker
-Requires=docker.service
+Wants=docker.service
 After=docker.service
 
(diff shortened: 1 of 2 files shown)


dkoshkin:
LGTM

alexbrand:
Merging this as failures are flakes.

PR #278, opened 2017-02-07
Fix python bootstrap script

### LINKED ISSUE #270: Failure during python 2.7 installation is not being caught
We install python 2.7 on Ubuntu nodes, because otherwise we can't use ansible. If the installation fails, we ignore the failure and:
1) We create a symbolic link pointing to nowhere
2) The play moves on, because we are not checking for success

The fix to this should be two-fold:
1) Validate that the python installation script exited successfully in the ansible play
2) Create the symbolic only if python was installed successfully

It seems like we could ditch the script and use ansible tasks for installing python on Ubuntu nodes.

alexbrand:
Fixes #270

### DIFF
--- a/ansible/roles/all/files/python27.sh
+++ b/ansible/roles/all/files/python27.sh
@@ -1,4 +1,5 @@
 if ! which python > /dev/null; then
-  apt-get update -qq && apt-get install -qq python2.7
-  ln -s /usr/bin/python2.7 /usr/bin/python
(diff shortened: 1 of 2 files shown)


dkoshkin:
Nice, LGTM

alexbrand:
I had to remove the when condition, because ansible doesn't have the information yet.

dkoshkin:
Ahh yep

PR #279, opened 2017-02-07
Remove version from DNS name

dkoshkin:
Having the version in the dns pod name makes it more complicated to upgrade.


ansible/group_vars/all.yaml
```diff
@@ -7,7 +7,7 @@ kismatic_docker_engine_apt_version: 1.11.2-0~xenial
 calico_docker_version: "v1.0.0"
 calico_kube_policy_controller_version: "v0.4.0"
 kubedns_version: "1.9"
-kubedns_version_id: "v19" # Used for the kube-dns replication controller's name (e.g. kube-dns-v19)
+kubedns_version_id: "v19" # Used for the kube-dns replication controller's version label
```
alexbrand: Should we get rid of this var? Doesn't seem to be used.

alexbrand:
LGTM

PR #280, opened 2017-02-08
Pipelining support

alexbrand:
Enabling pipelining in Ansible reduces the number of SSH connections
opened for each task performed. It is not enabled by default because
it is incompatible with the `requiretty` setting that is found
in some Linux distributions' default sudoers configuration.

With this change, we enable pipelining at the node level if we detect
that the sudoers file does not have the `requiretty` setting.

Given that CentOS does not support pipelining out of the box, I have also changed most of our integration tests to use Ubuntu instead.

### DIFF
--- a/ansible/roles/all/tasks/main.yaml
+++ b/ansible/roles/all/tasks/main.yaml
@@ -1,3 +1,13 @@
 ---
 - name: Install python 2.7
   script: python27.sh
+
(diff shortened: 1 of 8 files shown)


dkoshkin:
Closing and opening another PR to use new SNAP pipeline

PR #281, opened 2017-02-08
Add k8s ports to verify

dkoshkin:


alexbrand:
Sweet! I'd like to better understand some of these ports we are openning, but this LGTM as a stop-gap.

PR #284, opened 2017-02-10
Unfiltered upgrade static

dkoshkin:
This PR:
* Merges in static pod changes
* Reuses the existing playbooks for the `upgrade` command
* Labels control-plane components with both the upstream `kubernetes` version and the `kismatic` version

In this PR I have also commented out upgrade steps that are higher risk and/or will not work with existing plays. This includes `docker-engine` and `etcd`

The first 2 commits have been reviewed as thats just a merge from the `updata` branch.


alexbrand:
LGTM

PR #287, opened 2017-02-10
Upgrade cleanup

### LINKED ISSUE #285: Upgrades: Perform clean up of systemd services and binaries
Given that we have decided to move most of the kubernetes components to run inside containers, we need to remove the systemd services and binaries from old installations during an upgrade.

This clean up shall be performed in two phases, mainly to enable the ability to rollback (manually) if something goes wrong:
1. Stop the service before deploying the new, containerized component
2. Delete the systemd unit file and binary once the service is up and running

dkoshkin:
Fixes #285 

Cleanup old packages and unit files when doing an upgrade.

### DIFF
--- a/ansible/_calico.yaml
+++ b/ansible/_calico.yaml
@@ -8,5 +8,14 @@
     vars_files:
       - group_vars/all.yaml
 
+    pre_tasks:  # stop all systemd services for older installs
+      - name: stop calico-node service
+        service:
+          name: calico-node.service
+          state: stopped
+          enabled: no
+        ignore_errors: yes  # this is not ideal but there is no good OS independent way to know if it failed or doesn't exist
+        when: upgrading is defined and upgrading|bool == true
+
     roles:
       - calico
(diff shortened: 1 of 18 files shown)


pkg/cli/upgrade.go
```diff
@@ -92,7 +92,12 @@ func doUpgradeOffline(out io.Writer, planFile string, opts upgradeOpts) error {
 	var toSkip []install.ListableNode
 	for _, n := range cv.Nodes {
 		if install.IsOlderVersion(n.Version) {
-			toUpgrade = append(toUpgrade, n)
+			// prepend so masters get upgraded first
+			if util.Subset([]string{"master"}, n.Roles) {
```
alexbrand: The business logic in the install package enforces ordering of upgrades, so this seems redundant.

ansible/_calico.yaml
```diff
@@ -8,5 +8,14 @@
     vars_files:
       - group_vars/all.yaml
 
+    pre_tasks:  # stop all systemd services for older installs
+      - name: stop calico-node service
+        service:
+          name: calico-node.service
+          state: stopped
+          enabled: no
+        ignore_errors: yes  # this is not ideal but there is no good OS independent way to know if it failed or doesn't exist
```
alexbrand: Agreed that this is far from ideal. Maybe we can run this task _if_ we find the systemd unit file in the system? Or maybe use systemctl to determine if the service is installed. I'll create a follow up task as I don't want to focus on this now.

alexbrand: Created #289 as follow up

PR #288, opened 2017-02-13
Use Node instead of just IP in ListableNode struct

alexbrand:
The ListableNode struct only included the node's IP, which resulted in some awkward usage and the need to lookup nodes in the plan using an IP address. With this change, the Node is part of the ListableNode struct.

Also added SSH connectivity validation to the info command, as it was succeeding with non-existent infrastructure.


alexbrand:
Test failure is flake. Merging as this is blocking other upgrades work.

PR #294, opened 2017-02-13
Create manifests dir using ansible

### LINKED ISSUE #293: Upgrades: /etc/kubernetes/manifests does not exist when upgrading
The `/etc/kubernetes/manifests` directory was introduced with the static pod self-hosting work. Attempting to upgrade an old cluster throws an error saying that this directory does not exist.

The problem is that the directory is being created in the systemd unit file as a pre-task, instead of in the kubelet ansible play.

alexbrand:
Fixes #293


dkoshkin:
LGTM

PR #305, opened 2017-02-21
Initial work on online upgrade command

### LINKED ISSUE #256: Online upgrade
An online upgrade is just an offline upgrade with two twists:

1. We will exit if any of the nodes is detected as being unsafe to upgrade online. This should be accompanied by a list of reasons and potential mediations for each node.
2. We will attempt to drain each worker node prior to upgrade, and uncordon each worker after it is upgraded.

`kismatic upgrade offline [-f plan=kismatic-cluster.yaml]`

**Node Safety**

For each node, we will check each of its roles to see if it has any of the offline preventing issues below. If so, we will generate a report of all such violations, output it and exit.

_Etcd_

- If there are less than 3 etcd nodes, upgrading one will make the cluster unavailable.

_Master_

- If there is a single master, upgrading it will make cluster features unavailable.
- If a master is also acting as the load balanced endpoint, upgrading it will make cluster features unavailable.

_Ingress_

- As we do not control the load balancing of ingress, if a node is an Ingress Node, upgrading it will make workloads unavailable.

_Gluster_

- As many different permutations of volume capabilities can be included on a Gluster node, if a node is a Gluster Node, upgrading it will potentially make workloads unavailable.

_Worker_

- If there is a single worker, upgrading it will make cluster features unavailable.
- If a node is host to a pod using a HostPath or emptyDir volume, upgrading it would potentially cause data loss.
- If a node is the only host to a DaemonSet, upgrading it could potentially decrease availability
- If a node is running a pod that has no peers, upgrading it could cause a loss of data or availability
- If a node is running a pod that is not managed by an RC, RS, Job or DS, upgrading it could cause a loss of data or availability.

For now, we're ignoring alpha features like [`taints, affinities and tolerations`](https://groups.google.com/forum/%23!msg/kubernetes-dev/-fwfHQtm9GE/H1TU4yqiBAAJ&sa=D&ust=1484942815608000&usg=AFQjCNHynESpssTvErCFdqV94oT2tdYiOA) in this detection algorithm, but these certainly should be taken into account once they become mainstream.

alexbrand:
This PR adds support for an online upgrade. During an online upgrade, Kismatic will run a variety of checks to ensure that we are able to perform an upgrade without affecting availability or data loss.

Resolves #256, minus cordon/uncordon of nodes, which will be addressed in #306 

- [x] Online upgrade command
- [x] Online upgrade detection algorithm
- [x] Add stateful set to detection alg

### DIFF
--- a/pkg/install/upgrade.go
+++ b/pkg/install/upgrade.go
@@ -1,0 +1,259 @@
+package install
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/apprenda/kismatic/pkg/data"
+)
+
+const kubeCreatedBy = "kubernetes.io/created-by"
+
+type upgradeKubeInfoClient interface {
+	data.PodLister
+	data.DaemonSetGetter
+	data.ReplicationControllerGetter
+	data.ReplicaSetGetter
+	data.PersistentVolumeClaimGetter
+	data.PersistentVolumeGetter
+	data.StatefulSetGetter
+}
+
+type masterNodeCountErr struct{}
+
+func (e masterNodeCountErr) Error() string {
+	return "This is the only master node in the cluster. " +
+		"Upgrading it will make the cluster unavailable."
+}
+
+type masterNodeLoadBalancingErr struct{}
+
+func (e masterNodeLoadBalancingErr) Error() string {
+	return "This node is acting as the load balanced endpoint for the master nodes. " +
+		"Upgrading it will make the cluster unavailable"
+}
+
+type ingressNotSupportedErr struct{}
+
+func (e ingressNotSupportedErr) Error() string {
+	return "Performing an online upgrade of ingress nodes is not supported."
+}
+
+type storageNotSupportedErr struct{}
+
+func (e storageNotSupportedErr) Error() string {
+	return "Performing an online upgrade of storage nodes is not supported."
+}
+
+type workerNodeCountErr struct{}
+
+func (e workerNodeCountErr) Error() string {
+	return "This is the only worker node in the cluster. " +
+		"Upgrading it will make cluster features unavailable."
+}
+
+type podUnsafeVolumeErr struct {
+	namespace string
+	name      string
+	volType   string
+	volName   string
+}
+
+func (e podUnsafeVolumeErr) Error() string {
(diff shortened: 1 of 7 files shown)


pkg/install/upgrade.go
```diff
@@ -0,0 +1,260 @@
...
+	data.ReplicationControllerGetter
+	data.ReplicaSetGetter
+	data.PersistentVolumeClaimGetter
+	data.PersistentVolumeGetter
+	data.StatefulSetGetter
+}
+
+type etcdNodeCountErr struct{}
+
+func (e etcdNodeCountErr) Error() string {
```
dkoshkin: I guess we need to make a decision if we are going to upgrade etcd and have this check.

pkg/install/upgrade.go
```diff
@@ -0,0 +1,260 @@
...
+			if err != nil {
+				errs = append(errs, fmt.Errorf("DaemonSet pod is running, but failed to get information about DaemonSet %s/%s", r.Reference.Namespace, r.Reference.Name))
+				continue
+			}
+			// Check if other nodes should be running this DS
+			if ds.Status.DesiredNumberScheduled < 2 {
+				errs = append(errs, podUnsafeDaemonErr{dsNamespace: r.Reference.Namespace, dsName: r.Reference.Name})
+			}
+		case "job":
+			continue
```
dkoshkin: Should we fail if there is a running job?

alexbrand: My thinking was that jobs are re-scheduled if they fail. However, I think I'd rather err on the safe side until someone complains. I'll add a new error for jobs.

pkg/install/upgrade.go
alexbrand: I think we can keep the etcd logic. We'll need it eventually. What do you think?

dkoshkin: Maybe comment it out? That might be the only problem a user has with the cluster and we block the upgrade because of it even though we aren't actually upgrading it.

alexbrand: Oh yeah! that's an excellent point. I don't know why I assumed it wouldn't run 😄

pkg/install/upgrade.go
alexbrand: Done. Thanks!

dkoshkin:
LGTM!

alexbrand:
Will re-run build now that we have tests.

PR #308, opened 2017-02-21
Run control-plane validation on each node during upgrade-node step

### LINKED ISSUE #302: Upgrades: Master nodes are being deemed successfully upgraded without actually checking
The play that validates the control plane is running after all master nodes have run. The problem here is that the version file is being updated on a master node without actually checking that the upgrade on the node was successful.

We need to ensure that a given master node was upgraded successfully before updating the version file. This might be achieved by running the control plane validation play on that node, or by any other means available.

dkoshkin:
Fixes #302 

Run `validate-control-plane` on each node after the node has been upgraded.
This does require the `kubectl` package to be installed on all nodes, not just on the masters.


alexbrand:
Is this PR for addressing #302? I am not following why we need to validate the control plane when upgrading nodes other than the master nodes (which is where the control plane is running). Looks like the kube-proxy is being lumped into the control plane components, which it is not.

Seems like we have two problems here:
- The one described in 302
- Validating that the kube-proxy is up and running after installing/upgrading a worker node. I suggest we address this as a separate issue. It seems like self-hosting raises a challenge here, as you can't verify if a pod is running without depending on another node (API Server on master). Ideally, we should be able to tell if the kube-proxy is running by inspecting the node itself. Maybe using the healthz endpoint?

alexbrand:
Looks like there are test failures on this PR. Will wait until addressed for review.

alexbrand:
This LGTM

PR #309, opened 2017-02-21
Restart services

dkoshkin:
* Move package installation to the roles that use them.
* Only restart services when the packages are changed on upgrade.


alexbrand:
LGTM!

PR #310, opened 2017-02-21
Check if systemd service is present before stopping

### LINKED ISSUE #289: Upgrades: Look into running systemd stop command only if service is there
As part of #287, we added commands to stop existing systemd files. The current implementation will ignore failures if anything goes wrong. Ideally, we would like to know if stopping the service was successful. Otherwise, we could potentially end up in a scenario where a service that should not be running remains running.

dkoshkin:
Fixes #289

### DIFF
--- a/ansible/_calico.yaml
+++ b/ansible/_calico.yaml
@@ -9,13 +9,19 @@
       - group_vars/all.yaml
 
     pre_tasks:  # stop all systemd services for older installs
+      - name: check if calico-node is active
+        command: systemctl is-active calico-node.service
+        register: status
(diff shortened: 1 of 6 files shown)


ansible/_calico.yaml
```diff
@@ -9,13 +9,19 @@
       - group_vars/all.yaml
 
     pre_tasks:  # stop all systemd services for older installs
+      - name: check if calico-node is active
+        command: systemctl is-active calico-node.service
```
alexbrand: I had no idea this was a thing. Nice! 👏

alexbrand:
LGTM!

alexbrand:
Closed by mistake.

PR #311, opened 2017-02-22
Refactor test setup, and add upgrade test

alexbrand:
Initial set of tests for upgrades. Upgrading on CentOS is currently broken, so I have disabled that spec in the suite.

In this PR, I also ensure that each test gets a pristine copy of kismatic. We have been able to get away with not doing this because the number of parallel test runners was higher than the number of tests. This meant that each runner was getting it's own temp directory. Conversely, each test was getting it's own copy of kismatic.

### DIFF
--- a/integration/upgrade_test.go
+++ b/integration/upgrade_test.go
@@ -1,0 +1,67 @@
+package integration
+
+import (
+	"os"
+	"os/exec"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("kismatic upgrade tests", func() {
+	BeforeEach(func() {
+		dir := setupTestWorkingDirWithVersion("v1.2.2")
+		os.Chdir(dir)
+	})
+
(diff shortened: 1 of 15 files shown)


integration/upgrade_test.go
```diff
@@ -0,0 +1,67 @@
...
+						Expect(err).ToNot(HaveOccurred())
+
+						// Extract current version of kismatic
+						pwd, err := os.Getwd()
+						Expect(err).ToNot(HaveOccurred())
+						err = extractCurrentKismatic(pwd)
+						Expect(err).ToNot(HaveOccurred())
+
+						// Perform upgrade
+						cmd := exec.Command("./kismatic", "upgrade", "offline", "-f", "kismatic-testing.yaml", "--skip-preflight")
```
dkoshkin: Do we need to `--skip-preflight`?

alexbrand: We need it for now as preflight is broken. I vote for having this test to at least have something, and fix it as we go.

dkoshkin:
LGTM

PR #318, opened 2017-02-22
Fix upgrades inspector

### LINKED ISSUE #290: Upgrades: The package checks are always enforced
Currently, the package checks are always being enforced, regardless of the `allow_package_installation` option. 

I discovered this when trying to run an upgrade against a cluster running on ubuntu. Given that we are hosting the packages under a different release (i.e. `kismatic-xenial` instead of `xenial`), the checks fail because they were not updated to target the specific release.

It seems like the fix to this would be to update the checks to specify the target release. However, I suggest that we run the checks if and only if `allow_package_installation` is set to false. This would cover us for cases where the repository changes location, or any other similar change.

alexbrand:
Fixes #290

### DIFF
--- a/pkg/inspector/check/package.go
+++ b/pkg/inspector/check/package.go
@@ -18,9 +18,9 @@ func (p PackageQuery) String() string {
 // The PackageCheck uses the operating system to determine whether a
 // package is installed or is available for installation.
 type PackageCheck struct {
-	PackageQuery        PackageQuery
-	PackageManager      PackageManager
-	InstallationAllowed bool
+	PackageQuery         PackageQuery
+	PackageManager       PackageManager
+	InstallationDisabled bool
 }
 
 // Check returns true if the package is installed. If installation is allowed,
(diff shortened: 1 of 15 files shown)


pkg/inspector/check/package.go
```diff
@@ -11,38 +15,36 @@ func (p PackageQuery) String() string {
...
-
-// Check returns true if the package is available. Otherwise returns false, or an error
-// if the check is unable to determine the condition.
-func (c PackageAvailableCheck) Check() (bool, error) {
-	ok, err := IsPackageReadyToContinue(c.PackageManager, c.PackageQuery)
+// Check returns true if the package is installed. If pkg installation is disabled,
+// we would like to check if the pacakge is available for install. However,
+// there is no guarantee that the node will have the kismatic package repo configured.
+// For this reason, this check is a no-op when package installation is disabled.
+func (c PackageCheck) Check() (bool, error) {
```
dkoshkin: On `pkg installation is disabled` don't we want to test if the package is installed or is available to install?

alexbrand: We can't perform that check because the node is not guaranteed to have the kismatic repository. One thing we could do I guess is check if we can reach the repo.

dkoshkin:
Got it, LGTM

PR #319, opened 2017-02-22
Stop systemd services if they are found to be running

### LINKED ISSUE #314: Upgrades: Systemd services are not being stopped
The current logic for stopping systemd services is incorrect since we are using `-q` on the systemd command, and then expecting a specific string in STDOUT.

This is blocking upgrades because the new API server fails to come up.

alexbrand:
Fixes #314


dkoshkin:
LGTM

PR #328, opened 2017-02-23
Better default explainer

### LINKED ISSUE #324: Upgrades: Remove play numbering
These numbers are not correct.

alexbrand:
Introduces an improved default explainer that performs live updates on the terminal screen.

Also removed play numbering to fix #324.


dkoshkin:
I like it, LGTM

PR #329, opened 2017-02-23
Lock golang version by using docker for building binaries

### LINKED ISSUE #327: Maintain go version in codebase, instead of CI system
The go version used in builds should be defined in code, instead of the build system.

alexbrand:
With this change, all our binaries are built using the official golang docker image.

Updating the go version is as simple as updating the variable in the Makefile.

We are still running a couple of go-dependent make targets (mainly the integration tests), so this is just an initial step towards not having to install Go in CI or dev machines.

Fixes #327

Notable changes:
* `make` no longer builds the inspector. It only builds kismatic.
* Mac users might observe a slight performance hit when building due to bugs in Docker for Mac (https://forums.docker.com/t/file-access-in-mounted-volumes-extremely-slow-cpu-bound/8076/269).
* Tests are also run inside a container


dkoshkin:
LGTM

PR #331, opened 2017-02-24
Etcd upgrade

### LINKED ISSUE #326: Upgrades: Upgrade Etcd
To keep the cluster at a consistent well defined state, etcd will also be upgraded when running `kismatic upgrade` for both `online` and `offline` commands.

- [x] Publish new "transitional" etcd package, this is required to upgrade old networking etcd from `v2.3.x` to `v3.0.x` then to the final `v3.1.x` versions
  - This means the whole cluster must be upgraded to `3.0` first then again to `3.1`
- [x] Fix bug with etcd service not restarting correctly
- [x] Will need to place an indicator on node that it has been upgraded to the transition etcd, without it rerunning the upgrade on a failure will fail due to the package conflict
  - No need to do this as we can rely on the `kismatic-version` file
- [x]  Add new upgrade code for etcd as described in [upgrade to 3.0](https://coreos.com/etcd/docs/latest/upgrades/upgrade_3_0.html) and [upgrade to 3.1](https://coreos.com/etcd/docs/latest/upgrades/upgrade_3_1.html)
- [x] Verify an etcd node has been upgraded and is ready
  - `cluster-health` seems to do this fine
- [x] Move cert copying to the certs play so it doesnt rerun on upgrades
- [x] Test the overall state of the etcd clusters, maybe another step after node-upgrade?
- [x] Figure out what happens for an `online` upgrade when an etcd node has multiple roles and those are `ingress`, `storage` - maybe these can be upgraded?

dkoshkin:
Fixes #326 

Depending on the outcome of https://github.com/apprenda/kismatic/pull/336 might need to keep uninstalling Docker or merge in the changes to not install when daemon is already running.


ansible/roles/etcd/tasks/main.yaml
```diff
@@ -16,15 +16,30 @@
...
+      ((etcd_installation_centos is defined and etcd_installation_centos.changed == true or etcd_installation_deb is defined and etcd_installation_deb.changed == true) and upgrading is defined and upgrading|bool == true)
 
   - meta: flush_handlers  #Run handlers
 
+    #TODO don't use a timeout
+  - name: wait for etcd node to be ready after upgrade
+    pause: minutes=1
+    when: >
+      (force_etcd_restart is defined and force_etcd_restart|bool == true) or
+      ((etcd_installation_centos is defined and etcd_installation_centos.changed == true or etcd_installation_deb is defined and etcd_installation_deb.changed == true) and upgrading is defined and upgrading|bool == true)
```
alexbrand: nit: can we move the if upgrading==true condition to the beginning of this line? I think it's going to be clearer

pkg/install/about.go
```diff
@@ -85,7 +92,8 @@ func ListVersions(plan *Plan) (ClusterVersion, error) {
 	nodes := plan.GetUniqueNodes()
 
 	cv := ClusterVersion{
-		Nodes: []ListableNode{},
+		Nodes:      []ListableNode{},
+		Etcd2Nodes: []ListableNode{},
```
alexbrand: Doesn't seem to be used

pkg/cli/upgrade.go
```diff
@@ -198,6 +198,23 @@ func upgradeNodes(out io.Writer, plan install.Plan, opts upgradeOpts, toUpgrade
 		}
 	}
 
+	// get all etcd nodes
+	var etcdToUpgrade []install.ListableNode
+	for _, n := range install.Subset(toUpgrade, "etcd") {
+		if install.IsOlderThanVersion("3.0.0", n.EtcdVersion) {
```
alexbrand: I am torn here. I think I'd prefer not getting into component-level versions, and always using kismatic version instead.

pkg/install/about.go
```diff
@@ -115,7 +123,13 @@ func ListVersions(plan *Plan) (ClusterVersion, error) {
 			}
 		}
 
-		cv.Nodes = append(cv.Nodes, ListableNode{node, plan.GetRolesForIP(node.IP), thisVersion})
+		roles := plan.GetRolesForIP(node.IP)
+		var etcdVersion Version
+		if util.Subset([]string{"etcd"}, roles) {
+			// TODO actually determine the version
+			etcdVersion = Version{Major: 2, Minor: 3, Patch: 7}
```
alexbrand: Using kismatic version (see previous comment) would remove the need for this.

pkg/install/about.go
```diff
@@ -126,3 +140,16 @@ func ListVersions(plan *Plan) (ClusterVersion, error) {
 func (v Version) String() string {
 	return fmt.Sprintf("%v.%v.%v", v.Major, v.Minor, v.Patch)
 }
+
+// Subset returns a filtered list of ListableNode slice based on the node's roles
+func Subset(nodes []ListableNode, roles ...string) []ListableNode {
```
alexbrand: NodesWithRoles() or something like that seems more appropriate here.

ansible/roles/packages-docker/tasks/main.yaml
```diff
@@ -2,8 +2,8 @@
   # YUM
   - name: install docker-engine yum package
     yum: name=docker-engine-{{ docker_engine_yum_version }} state=present
-    register: docker_installation
-    until: docker_installation|success
+    register: docker_installation_centos
```
dkoshkin: Good catch, what happens is `_installation` gets set by yum and then the same one gets reset by `deb`, need the 2 different vars

ansible/roles/packages-transition-etcd/tasks/main.yaml
```diff
@@ -0,0 +1,21 @@
+---
+  # YUM
+  - name: install transition-etcd yum package
+    yum: name=transition-etcd-{{ transition_etcd_yum_version }} state=present
```
dkoshkin: This package is v3 and the main reason was to avoid having anyone install this package by accident.

pkg/cli/upgrade.go
dkoshkin: I like this, this will simplify things alot. I will just make sure that if a node is already upgraded to a temporary etcd all the tasks are noop.

This was really for a scenario when an upgrade fails after installing the temp but before the node version is set.

ansible/_etcd-transition.yaml
```diff
@@ -0,0 +1,18 @@
...
+    vars_files:
+      - group_vars/all.yaml
+      - group_vars/etcd-networking.yaml
+
+    roles:
+      - role: packages-transition-etcd
+        when: allow_package_installation|bool == true
+      - etcd-backup
+      - etcd
+      - role: packages-transition-etcd-cleanup  # remove transition-etcd package
```
dkoshkin: I think a better solution would be to just name the package `etcd-3.0.17`, this way we wouldnt have to have: 
a. A special package
b. Cleanup, since it will just be upgraded in the next step.

pkg/cli/upgrade.go
```diff
@@ -135,7 +135,7 @@ func doUpgrade(out io.Writer, opts *upgradeOpts) error {
 		fmt.Fprintln(out, "All nodes are at the target version. Skipping node upgrades.")
 	} else {
 		if err = upgradeNodes(out, *plan, *opts, toUpgrade, executor); err != nil {
-			return err
+			return fmt.Errorf("Failed to upgrade nodes: %v", err)
```
dkoshkin: Need to remove this, not required after the merge

ansible/roles/etcd/tasks/main.yaml
```diff
@@ -12,19 +12,36 @@
...
+    when: >
+      (force_etcd_restart is defined and force_etcd_restart|bool == true) or
+      ((etcd_installation_rpm is defined and etcd_installation_rpm.changed == true or etcd_installation_deb is defined and etcd_installation_deb.changed == true) and
+      (upgrading is defined and upgrading|bool == true))
 
   - meta: flush_handlers  #Run handlers
 
+    #TODO don't use a timeout
+  - name: wait for {{ etcd_install_bin_name }} node to be ready after upgrade
+    pause: seconds=15
```
alexbrand: I wonder if we should use the official etcd guideline which is 2 minutes here?

ansible/upgrade-etcd2-node.yaml
```diff
@@ -0,0 +1,8 @@
+---
+  - include: _packages-etcd-cleanup.yaml
+    when: allow_package_installation|bool == true
+  - include: _packages.yaml
```
alexbrand: Is this going to install packages unrelated to etcd if the node has multiple roles? Trying to think if this would be problematic

ansible/upgrade-preflight.yaml
```diff
@@ -1,2 +1,5 @@
 ---
+  # make sure repos cache is updated
+  - include: _packages.yaml
```
alexbrand: Is this going to install packages during preflight?

ansible/_packages-etcd-cleanup.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
alexbrand: What's the reason for having this if we are cleaning up etcd packages in the packages-cleanup role?

dkoshkin: Not removing this package or the `etcd` (since its just going to be upgraded if needed) in the `_packages-cleanup.yaml` anymore.

This play is going to be used only when upgrading etcd2.

dkoshkin:
Renamed _packages.yaml` to `packages-repo.yaml` since its only responsible for configuring the repos and not actually installing anything.

ansible/_etcd-transition.yaml
```diff
@@ -0,0 +1,17 @@
+---
+  - hosts: etcd
+    any_errors_fatal: true
+    name: "Temporarily Install Transtion Etcd 3.x version"
```
alexbrand: typo: Transition

ansible/roles/etcd/tasks/main.yaml
```diff
@@ -11,20 +11,49 @@
       - reload services
       - restart etcd service
 
+  - name: get etcd server version
+    command: curl --cert {{ etcd_certificates_cert_file }} --key {{ etcd_certificates_key_file }} --cacert {{ etcd_certificates_ca_file }} https://127.0.0.1:{{ etcd_service_client_port }}/version
+    register: etcd_version
+    when: upgrading is defined and upgrading|bool == true
+
+  - name: check if etcd needs to be upgraded
```
alexbrand: I don't think this addresses the concern of re-installing the transition etcd server, does it? This check should probably gate the _etcd_transition play.

ansible/_packages-etcd-cleanup.yaml
alexbrand: What will remove old etcd packages when there is no need to upgrade etc2?

ansible/roles/packages-etcd/tasks/main.yaml
```diff
@@ -14,9 +14,16 @@
...
-    until: etcd_installation|success
+    register: etcd_installation_deb
+    until: etcd_installation_deb|success
     retries: 3
     delay: 3
     when: ansible_os_family == 'Debian'
+
+  # required until https://github.com/ansible/ansible/issues/14804
+  - name: install curl package
+    package: name=curl state=present
```
alexbrand: Needs Package preflight check

ansible/_packages-etcd-cleanup.yaml
dkoshkin: It should always be uninstalled already before running the regular cleanup but added it in there since its safe to rerun.

alexbrand:
Overall LGTM. 

@dkoshkin Do you think you could document the etcd upgrade process? I think I am following, but I want to be sure. We need the docs anyways :)

dkoshkin:
### Etcd Upgrade
Prior to `Kismatic v1.3.0`, all clusters were built using 2 versions of etcd.  
* Kubernetes etcd with versions `v3.0.x`
* Calico etcd with versions `v2.3.x` 

With `Kismatic v1.3.0` both versions of etcd will be upgraded to etcd `v.3.1.x` which is not backwards compatible with etcd `v2.3.x` or `v3.0.x`

#### Steps taken during the upgrade process are outlined below
##### Upgrading Calico etcd `v2.3.x` to `v.3.0.x`
On an etcd cluster that has more than 1 member, it is not possible to directly upgrading from `v2.3.x` to `v.3.1.x`, all nodes in the cluster need to be upgraded to `v3.0.x` then again to `v.3.1.x`

Run on each etcd node, one at a time:
1. Determine if a node is required to be upgraded by inspecting the `/etcd/kismatic-version` file
1. Install `transition-etcd` package containing the required upgraded binaries: `etcd_v3_0`, `etcdctl_v3_0`
1. Backup etcd data to `/etc/etcd_networking/backup/$epoch`
1. Restart the `etcd_networking` service
1. Wait until the node has caught up with the rest of the cluster
1. Verify cluster with `etcdctl_v3_0 --endpoint='https://127.0.0.1:6666 cluster-health` 

##### Upgrading Kubernetes and Calico etcd `v3.0.x` to `v.3.1.x`
Run on each etcd node, one at a time:
1. Determine if a node is required to be upgraded by inspecting the `/etcd/kismatic-version` file
1. Remove old `kismatic-etcd` and `transition-etcd` packages
1. Install new `etcd` package containing the required upgraded binaries: `etcd_k8s`, `etcd_networking` and `etcdctl`
1. Backup etcd data to `/etc/etcd_k8s/backup/$epoch`
1. Restart the `etcd_k8s` service
1. Wait until the member has caught up with rest of the cluster
1. Verify Kubernetes etcd cluster with `etcdctl --endpoint='https://127.0.0.1:2379 cluster-health` 
1. Backup etcd data to `/etc/etcd_networking/backup/$epoch`
1. Restart the `etcd_networking` service
1. Wait until the member has caught up with the rest of the cluster
1. Verify Calico etcd cluster with `etcdctl --endpoint='https://127.0.0.1:6666 cluster-health` 

During the upgrade the overall cluster version will operate at the lowest member version, once all etcd members have been successfully upgraded the cluster version will be upgraded and any new features will be turned on.

*Note* when all members are upgraded to `v3.1.x` downgrading **will not** be possible

PR #332, opened 2017-02-24
Drain nodes before online upgrade

### LINKED ISSUE #306: Upgrades: Cordon/uncordon nodes during an online upgrade
As defined in #256, each worker node should be cordoned before upgrade, and uncordoned after successful upgrade.

alexbrand:
Any node that has a kubelet running is drained before we touch it. If
the upgrade is successful, we uncordon the node only if it's a worker
node.

Fixes #306

### DIFF
--- a/ansible/_kube-drain-node.yaml
+++ b/ansible/_kube-drain-node.yaml
@@ -1,0 +1,7 @@
+---
+  - name: "Drain Node"
+    hosts: master:worker:ingress:storage
+    serial: 1
+    tasks:
+      - name: "Run kubectl drain"
+        command: "kubectl drain --ignore-daemonsets {{ inventory_hostname }}"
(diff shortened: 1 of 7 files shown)


ansible/_kube-drain-node.yaml
```diff
@@ -0,0 +1,7 @@
+---
+  - name: "Drain Node"
```
dkoshkin: Should we wait until everything has been drained?

alexbrand: I believe kubectl will block until the node is drained. I'll double check.

dkoshkin:
LGTM besides the one comment I had.

ansible/_kube-drain-node.yaml
alexbrand: Drain does indeed block until drain is complete.

PR #335, opened 2017-02-27
Detect restart loops

### LINKED ISSUE #307: Better Pod Status Validation
Currently the pod validation will sometimes report status as success even if a pod is in a restart loop.
Need to verify a pod is running and has been running for a certain period of time.

dkoshkin:
Closes #307 

Add a pause to try to detect restart loops on pods and kubelet service.

### DIFF
--- a/ansible/roles/kubelet/tasks/main.yaml
+++ b/ansible/roles/kubelet/tasks/main.yaml
@@ -37,3 +37,10 @@
     until: running|success
     retries: 3
     delay: 5
+
+  - name: pause to verify kubelet keeps running
+    pause:
+      seconds: 4
+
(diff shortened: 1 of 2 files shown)


alexbrand:
Relying on timing to detect this could work, but I don't think it'll catch everything. Could we make it more robust by using the restart count?

dkoshkin:
Just realized that this won't really work for a few reasons.
The `initialDelaySeconds` and `timeoutSeconds` wouldn't actually fail for 20 seconds. I've also seen that a component can restart 1 or 2 times and still be healthy after the install, removing this check for the pods and relying on the smoke test accomplishes this already.

PR #336, opened 2017-02-27
Don’t install docker if its already running

### LINKED ISSUE #330: Upgrades: Docker engine should not be installed if it already exists
After an upgrade 2 docker-engine packages are installed `kismatic-docker-engine` and `docker-engine`, the underlying docker-engine versions is the same for both.

In general if the node already has docker running there is no need to install a specific version since there is no hard dependency on any one version.

dkoshkin:
Closes #330 

Verify docker daemon is running with `docker ps`


dkoshkin:
Closing after discussing this further

PR #337, opened 2017-02-27
Add partial flag to the upgrades command

### LINKED ISSUE #257: Introduce partial flag
The partial flag would allow for either upgrade -- online or offline -- to apply the upgrade to the nodes that did not report a problem (unreadiness for offline upgrades, unreadiness/unsafety for online upgrades).

After applying upgrades, a report of which nodes that couldn't be upgraded will be delivered prior to exiting.

alexbrand:
Fixes #257

### DIFF
--- a/pkg/cli/upgrade.go
+++ b/pkg/cli/upgrade.go
@@ -39,7 +39,7 @@ func NewCmdUpgrade(out io.Writer) *cobra.Command {
 	cmd.PersistentFlags().StringVarP(&opts.outputFormat, "output", "o", "simple", "installation output format (options \"simple\"|\"raw\")")
 	cmd.PersistentFlags().BoolVar(&opts.skipPreflight, "skip-preflight", false, "skip upgrade pre-flight checks")
 	cmd.PersistentFlags().BoolVar(&opts.restartServices, "restart-services", false, "force restart cluster services (Use with care)")
-	cmd.PersistentFlags().BoolVar(&opts.partialAllowed, "partial-ok", false, "")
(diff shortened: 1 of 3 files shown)


pkg/cli/upgrade.go
```diff
@@ -38,6 +39,7 @@ func NewCmdUpgrade(out io.Writer) *cobra.Command {
 	cmd.PersistentFlags().StringVarP(&opts.outputFormat, "output", "o", "simple", "installation output format (options \"simple\"|\"raw\")")
 	cmd.PersistentFlags().BoolVar(&opts.skipPreflight, "skip-preflight", false, "skip upgrade pre-flight checks")
 	cmd.PersistentFlags().BoolVar(&opts.restartServices, "restart-services", false, "force restart cluster services (Use with care)")
+	cmd.PersistentFlags().BoolVar(&opts.partialAllowed, "partial-ok", false, "")
```
dkoshkin: Doe's this flag do anything in `offline` mode?

dkoshkin: Nvm `unreadyNodes` for those that fail the preflight.

pkg/cli/upgrade.go
```diff
@@ -163,38 +165,93 @@ func doUpgrade(out io.Writer, opts *upgradeOpts) error {
...
 			}
 		}
-		if foundErrs {
+		// If we found any unsafe nodes, and we are not doing a partial upgrade, exit.
+		if len(unsafeNodes) > 0 && !opts.partialAllowed {
 			return errors.New("Unable to perform an online upgrade due to the unsafe conditions detected.")
 		}
+		// Block the upgrade if partial is allowed but there is an etcd or master node
+		// that cannot be upgraded
+		if opts.partialAllowed {
```
dkoshkin: So we will only attempt a partial upgrade if all `master` and `etcd` nodes can be upgraded, got it.

PR #338, opened 2017-02-27
Move downloading docker images before stopping service

dkoshkin:
As recommended in the Calico [docs](http://docs.projectcalico.org/v2.0/getting-started/kubernetes/upgrade) pulling images before stopping the service.

Also doing the same for the k8s control-plane.

### DIFF
--- a/ansible/_calico.yaml
+++ b/ansible/_calico.yaml
@@ -9,6 +9,13 @@
       - group_vars/all.yaml
 
     pre_tasks:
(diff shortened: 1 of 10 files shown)


alexbrand:
LGTM 

Related: what does this mean for the case where there is no internet access?

dkoshkin:
It will pull from the provided registry, because we are upgrading that in the prior step `Upgrade Docker Registry` should be fine.

PR #342, opened 2017-02-28
Support pre-release version information in version

### LINKED ISSUE #340: Upgrades: Support semver pre-release version identifiers
The current version parsing code does not support pre-release version information. Add support for this following the semver.org guidelines.

alexbrand:
Fixes #340

### DIFF
--- a/pkg/install/about.go
+++ b/pkg/install/about.go
@@ -2,85 +2,59 @@ package install
 
 import (
 	"fmt"
-	"regexp"
-	"strconv"
-	"strings"
 
 	"github.com/apprenda/kismatic/pkg/ssh"
+	"github.com/blang/semver"
 )
 
-type KismaticInfo struct {
-	ShortVersion Version
-	BuildNumber  string
-}
-
-type Version struct {
-	Major int
-	Minor int
-	Patch int
-}
-
+// ClusterVersion contains version information about the cluster
 type ClusterVersion struct {
-	EarliestVersion Version
-	LatestVersion   Version
+	EarliestVersion semver.Version
+	LatestVersion   semver.Version
 	IsTransitioning bool
 	Nodes           []ListableNode
 }
 
+// ListableNode contains version and role information about a given node
 type ListableNode struct {
 	Node    Node
 	Roles   []string
-	Version Version
+	Version semver.Version
 }
 
-var AboutKismatic KismaticInfo
+// AboutKismatic contains the version information of the currently running binary
+var AboutKismatic semver.Version
 
-// Takes a version in form "v{short version}-{changeset number}-{changeset ID}[-{dirtyflag}]"
-func SetVersion(polyVersion string) {
-	re := regexp.MustCompile("v([^-]+)-([^-]+)-([^-]+)-?([^-]+)")
-	matches := re.FindStringSubmatch(polyVersion)
-	if len(matches) > 2 {
-		ver := parseVersion(matches[1])
-		AboutKismatic = KismaticInfo{ver, matches[2]}
-	} else {
-		fmt.Printf("Could not parse %v", polyVersion)
+// SetVersion parses the given version, and sets it as the global version of the binary
+func SetVersion(v string) {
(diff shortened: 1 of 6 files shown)


pkg/install/about.go
```diff
@@ -2,85 +2,151 @@ package install
...
+	this := AboutKismatic
 	return this.isNewerThan(that)
 }
 
-// Returns true if that is older than this; this > that
 func (this Version) isNewerThan(that Version) bool {
 	if this.Major > that.Major ||
 		this.Minor > that.Minor ||
 		this.Patch > that.Patch {
 		return true
```
dkoshkin: Is this valid? ex.
```
this=1.2.3
that=1.3.0
1 > 1 false
2 > 3 false
3 > 0 true

so false or false or true == true > 1.2.3 isNewerThan 1.3.0 >> but it shoudlnt be
```

pkg/install/about.go
```diff
@@ -2,85 +2,151 @@ package install
...
 	}
+	// From semver: When major, minor, and patch are equal, a pre-release version has lower precedence than a normal version
+	if this.PreRelease == "" && that.PreRelease != "" {
```
dkoshkin: We aren't checking if major, minor and patch versions are equal before this so.
```
this=1.2.3 
that=1.3.0-alpha

1.2.3 isNewerThan 1.3.0-alpha
```

pkg/install/about.go
```diff
@@ -2,85 +2,151 @@ package install
...
+	thisPreRelIdentifiers := strings.Split(this.PreRelease, ".")
+	thatPreRelIdentifiers := strings.Split(that.PreRelease, ".")
+
+	for i := range thisPreRelIdentifiers {
+		thisPre := thisPreRelIdentifiers[i]
+		// from semver: A larger set of pre-release fields has a higher
+		// precedence than a smaller set, if all of the preceding identifiers are equal
+		if i > len(thatPreRelIdentifiers)-1 {
+			return true
+		}
```
dkoshkin: What is this doing, seems like its checking the version number to the length of `that` pre-release tags?

pkg/install/about.go
```diff
@@ -2,85 +2,151 @@ package install
...
+	}
 	return false
 }
 
-func parseVersion(versionString string) Version {
+func isNumericIdentifier(s string) bool {
+	if _, err := strconv.Atoi(s); err != nil {
+		return false
+	}
+	if s[0] == '0' {
```
dkoshkin: Why is `0` not numeric?

pkg/install/about.go
```diff
@@ -2,85 +2,174 @@ package install
...
 
-// Returns true if that is older than this; this > that
 func (this Version) isNewerThan(that Version) bool {
-	if this.Major > that.Major ||
-		this.Minor > that.Minor ||
-		this.Patch > that.Patch {
+	if this.Major > that.Major {
+		return true
+	}
+	if this.Major > that.Major && this.Minor > that.Minor {
```
dkoshkin: Should this be `if this.Major == that.Major && this.Minor > that.Minor`?

pkg/install/about.go
```diff
@@ -2,85 +2,174 @@ package install
...
+		return true
+	}
+	if this.Major > that.Major && this.Minor > that.Minor && this.Patch > that.Patch {
```
dkoshkin: `this.Major == that.Major && this.Minor == that.Minor && this.Patch > that.Patch`

pkg/install/about.go
```diff
@@ -2,85 +2,174 @@ package install
...
+		return true
+	}
+	// If the above checks failed, and none of the versions have pre-release, there
+	// is nothing else to check
+	if this.PreRelease == "" && that.PreRelease == "" {
```
dkoshkin: `(this.PreRelease == "" && that.PreRelease == "") or  this.PreRelease == that.PreRelease` looks like this is needed for the next 2 checks to work correctly

dkoshkin:
LGTM!

PR #345, opened 2017-03-01
Rename task, always remove old unit files

dkoshkin:
* Rename task to make it clearer what it does
* Always remove old systemd files, this task is safe to run even if the file doesn't exist so a check is not required


alexbrand:
LGTM

PR #349, opened 2017-03-01
Static Pods and Upgrades

dkoshkin:


alexbrand:
Boom! Fixes #upgrades :)

PR #366, opened 2017-03-02
Run kube dashboard with 2 replicas

### LINKED ISSUE #352: HA deployment for Dashboard
Currently, the dashboard is deployed with a single replica on the cluster.

In order to support an online upgrade, the Dashboard must be running with more than one replica.

dkoshkin:
Fixes #352 

Deploy the kubernetes dashboard with 2 replicas, on a single worker the replica count will be 1

### DIFF
--- a/integration/install.go
+++ b/integration/install.go
@@ -43,8 +43,8 @@ func installKismaticMini(node NodeDeets, sshKey string) error {
 		Worker:                   []NodeDeets{node},
 		Ingress:                  []NodeDeets{node},
 		Storage:                  []NodeDeets{node},
-		MasterNodeFQDN:           node.Hostname,
-		MasterNodeShortName:      node.Hostname,
+		MasterNodeFQDN:           node.PublicIP,
+		MasterNodeShortName:      node.PublicIP,
 		SSHKeyFile:               sshKey,
 		SSHUser:                  sshUser,
 		AllowPackageInstallation: true,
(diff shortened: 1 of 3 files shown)


integration/install_test.go
```diff
@@ -124,6 +124,9 @@ var _ = Describe("kismatic", func() {
 					installOpts := installOptions{allowPackageInstallation: true}
 					err := installKismatic(nodes, installOpts, sshKey)
 					Expect(err).ToNot(HaveOccurred())
+					By("Trying to access the Dashboard")
```
alexbrand: I'd like to figure out a way for us to write individual specs for these, instead of asserting multiple things in the same spec.

dkoshkin: Sounds reasonable, I will hold off on merging

alexbrand:
Can you add a comment similar to this one in the test?
```
// This spec will be used for testing non-destructive kismatic features on
// an upgraded cluster.
// This spec is open to modification when new assertions have to be made
```
I just want to make sure we know which test is the one we should use for this.

integration/install.go
```diff
@@ -43,8 +43,8 @@ func installKismaticMini(node NodeDeets, sshKey string) error {
 		Worker:                   []NodeDeets{node},
 		Ingress:                  []NodeDeets{node},
 		Storage:                  []NodeDeets{node},
-		MasterNodeFQDN:           node.Hostname,
-		MasterNodeShortName:      node.Hostname,
+		MasterNodeFQDN:           node.PublicIP,
```
alexbrand: Curious about the reason for this change, and how/if it affects other tests.

dkoshkin: This was to be able to access the dashboard from the CI machine, since the local hostname isn't available outside the cluster.

IMO this should have been the external address all along since we would want the nodes accessible outside the cluster.

alexbrand:
LGTM. Not sure if we need a better run for merging though.

PR #368, opened 2017-03-03
Update kube-dns rc to deployment

### LINKED ISSUE #351: HA deployment for cluster DNS
Currently, we deploy kube-dns as a replication controller with a single replica. In order to support an online upgrade, DNS must have more than one replica running on the cluster.

- [x] Investigate community recommendations when it comes to multiple kube-dns pods running
- [x] Move kube-dns to a Kubernetes deployment if it makes sense
- [x] Increase replica count of kube-dns deployment

dkoshkin:
Fixes #351 

Update kube-dns yaml file to a k8s deployment similar to https://github.com/kubernetes/kubernetes/blob/release-1.5/cluster/addons/dns/skydns-rc.yaml.base

There is no image version changes.

### DIFF
--- a/ansible/group_vars/all.yaml
+++ b/ansible/group_vars/all.yaml
@@ -113,6 +113,7 @@ docker_certificates_group: root
 kubernetes_master_apiserver_count: "{{ groups['master'] | length }}"
 kubernetes_master_ip: https://{{ kubernetes_load_balanced_fqdn }}:{{ kubernetes_master_secure_port }}
 kubernetes_schedulable: "{% if 'worker' in group_names %}true{% else %}false{% endif %}"
+kube_dns_replicas: "{{ [2, groups['worker'] | length] | min }}"
 # kubernetes certificate config
 kubernetes_certificates_ca_file_name: ca.pem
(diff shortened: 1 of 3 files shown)


dkoshkin:
Running DNS with multiple replicas is supported in the community, in fact there is a DNS [autoscaler](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/dns-horizontal-autoscaler) that is just a k8s deployment, I dont think KET should install this as this will be better left to the operator. 

It is also safe for the operator to manually scale replicas up and down based on the load.

alexbrand:
Would having 100 DNS pods on a cluster with 100 worker nodes be detrimental to performance?

alexbrand:
I misread the change. LGTM!

alexbrand:
Although it would be good to know why the healthz endpoint changed.

dkoshkin:
Found the [issue](https://github.com/kubernetes/kubernetes/issues/30633) related to the health endpoint.

Looks like the way we had it was a bug and it should have been this way all along since it was merged in back in August of last year.

PR #369, opened 2017-03-03
Place cluster specs in /etc/kubernetes/specs instead of /tmp

### LINKED ISSUE #350: Install spec files to somewhere not in /tmp
The `/tmp` directory is unstable and it will be good to have the spec files for later inspection and debugging purposes.

dkoshkin:
Closes #350 

Moving specs generated by the installer to `/etc/kubernetes/specs`.
The `/tmp` directory is not a good place to store them.


alexbrand:
Does this have any implications at upgrade time? Will these be overwritten by ansible when upgrading?

alexbrand:
@dkoshkin and I discussed the upgrade concerns and we agreed that it should be OK. Deployments in K8s support rollback if needed.

PR #371, opened 2017-03-03
Implement skunkworks cluster upgrade test

### LINKED ISSUE #356: Upgrade Tests: Ability to perform an offline upgrade against a skunkworks cluster
This test should verify the ability to upgrade a Skunkworks KET deployment using `./kismatic upgrade offline`.

Infrastructure: 3/2/3/2/2
OS: CentOS
Pre-upgrade version: Ideally current releae - 1, but 1.2.2 works.

An upgrade is considered successful if ./kismatic info returns the right cluster and node version after upgrading.

alexbrand:
Fixes #356

### DIFF
--- a/integration/framework.go
+++ b/integration/framework.go
@@ -115,3 +115,60 @@ func WithMiniInfrastructure(distro linuxDistro, provisioner infrastructureProvis
 
 	f(nodes.worker[0], sshKey)
 }
+
+// SubDescribe allows you to define specifications inside another spec.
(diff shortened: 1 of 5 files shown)


dkoshkin:
LGTM! Although it looks like we need to increase the timeout in Snap

alexbrand:
Yeah, looks like it.

alexbrand:
The new test introduced in this PR passed, so I am merging this and unblocking a bunch of other test work.

PR #380, opened 2017-03-07
Add post-upgrade HA test

### LINKED ISSUE #361: Upgrade Tests: Verify cluster is functional if a master node goes away
This test is to ensure that an upgraded HA cluster is correctly configured in an HA manner.

If possible, this should be done in a non-destructive manner so that we can reuse the cluster stood up in #356. Otherwise, this test will need it's own dedicated cluster.

alexbrand:
Fixes #361

### DIFF
--- a/integration/upgrade_test.go
+++ b/integration/upgrade_test.go
@@ -3,6 +3,7 @@ package integration
 import (
 	"os"
 	"os/exec"
+	"time"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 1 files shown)


dkoshkin:
Is it possible to reuse the "skunkworks" cluster for this, maybe this can be the last test?

alexbrand:
Merging for now. We can look into reusing it if necessary.

PR #382, opened 2017-03-07
Update component versions

alexbrand:


dkoshkin:
LGTM, although we will need to figure out a way to update the docs automatically :(

PR #383, opened 2017-03-07
Add no internet upgrade test

### LINKED ISSUE #355: Upgrade Tests: Ability to perform an offline upgrade with no internet access
This test should verify the ability to upgrade a cluster that is deployed on machines that have no internet access.

Infrastructure: 1/1/1/0/0

An upgrade is considered successful if ./kismatic info returns the right cluster and node version after upgrading.

alexbrand:
Fixes #355


dkoshkin:
LGTM

PR #386, opened 2017-03-07
Add larger cluster upgrade tests, move HA test to skunksworks

### LINKED ISSUE #354: Upgrade Tests: Ability to perform offline upgrades (3/2/2)
This test should verify the ability to upgrade a Minikube-style KET deployment using `./kismatic upgrade offline`

Permutations: Minor versions of KET * RHEL

An upgrade is considered successful if `./kismatic info` returns the right cluster and node version after upgrading.

### LINKED ISSUE #390: Upgrade Testing: Fix disconnected install
On an upgrade we have conditions:
```
((kismatic_offline_installation_rpm is defined and kismatic_offline_installation_rpm.changed == true) or
(kismatic_offline_installation_deb is defined and kismatic_offline_installation_deb.changed == true))
```

During a disconnected install with `allow_package_installation = false` packages will not be installed and these flags will never be set. 

To support this case a new check will be added:
```
(allow_package_installation|bool == false or
((kismatic_offline_installation_rpm is defined and kismatic_offline_installation_rpm.changed == true) or
(kismatic_offline_installation_deb is defined and kismatic_offline_installation_deb.changed == true)))
```
This assumes that new packages were installed and to trigger all the tasks that depend on that.

dkoshkin:
Fixes #354 
Fixes #390 

Let me know if you see any drawbacks to including the HA test in the skunkworks one. I think if that is always the last test it should be safe.


alexbrand:
The one downside is that it is destructive, but addressing that with a comment should be fine.

LGTM

alexbrand:
In the disconnected install test we verify that there is indeed no internet access by trying to access google.com, or something. I think we should make the same assertion just to make sure we didn't break the test. Another alternative is to look into using AWS security groups for this test, instead of iptables.

PR #392, opened 2017-03-08
Minor text changes [ci skip]

### LINKED ISSUE #391: Misleading step name when multiple master nodes are being smoke-tested
After the initial installation of a multi-master cluster, we run the smoke test on each node serially.

Currently, the output looks like this:
```
Smoke Test Master Nodes                                                         [OK]
Smoke Test Master Nodes                                                         [OK]
```

This is misleading as it implies the masters are being smoke-tested multiple times.

dkoshkin:
Fixes  #391
Some minor text changes.

### DIFF
--- a/ansible/_smoketest.yaml
+++ b/ansible/_smoketest.yaml
@@ -1,7 +1,7 @@
 ---
   - hosts: master
     any_errors_fatal: true
-    name: "Smoke Test Master Nodes"
+    name: Smoke Test Master Node
     remote_user: root
     become_method: sudo
     serial: 1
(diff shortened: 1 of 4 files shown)


ansible/_smoketest.yaml
```diff
@@ -1,7 +1,7 @@
 ---
   - hosts: master
     any_errors_fatal: true
-    name: "Smoke Test Master Nodes"
+    name: Smoke Test Master Node
```
alexbrand: Can we print the host name here?

dkoshkin: No unfortunately, the variable does not get substituted.

alexbrand: Bummer. Oh well, it is what it is.

alexbrand:
LGTM

PR #397, opened 2017-03-09
Increase retries for control-plane validation and calico

### LINKED ISSUE #389: Control plane validation intermittently fails when getting kube-apiserver pods
We have seen cases where the pods don't come up in time and the ansible play fails. It seems like we have to increase the timeout on this.

dkoshkin:
Fixes #389


alexbrand:
LGTM

PR #398, opened 2017-03-09
Remove calico config job after network comes up successfully

### LINKED ISSUE #395: Unable to run conformance test because calico job sticks around
The first thing the conformance test does is check that all pods in the kube-system namespace are Ready, or Failed. This seems like a bug in the conformance test code as it should allow completed jobs.

Fixing this in upstream will take some time, so I am suggesting we delete the job after it completes.

alexbrand:
Fixes #395

### DIFF
--- a/ansible/_calico.yaml
+++ b/ansible/_calico.yaml
@@ -38,3 +38,8 @@
 
     roles:
       - calico
+
+    post_tasks:
+      - name: remove calico config job
+        command: kubectl delete job configure-calico -n kube-system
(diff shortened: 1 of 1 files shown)


dkoshkin:
I like LGTM

dkoshkin:
Can you move this to a `post-task` in `_calico.yaml`? This will work better with the PR Im working on.

alexbrand:
Done. PTAL.

PR #401, opened 2017-03-09
adding logo folder and files to the project

jessicamaslyn:
changing the README.md to source the new logo location


alexbrand:
Thanks @jessicamaslyn! I'd rather not have another top-level folder. What do you think about putting the logos inside the `docs` dir, or maybe `docs/logo`?

jessicamaslyn:
Hey Alex, I was following the same format as https://github.com/kubernetes/kubernetes but moving it to docs/logo is no problem.

jaycoon:
@alexbrand - should we create a docs/images or docs/img? We will likely have other diagrams to add later

alexbrand:
@jessicamaslyn gotcha! I think having it inside docs makes more sense. What do you think?

@jaycoon That sounds like a great idea. Feel free to move existing images into that dir if it makes sense.

I know some of this will have an effect on our docs site when we build one. CC'ing @dkoshkin for his thoughts on this.

jessicamaslyn:
how about 2 separate folders within docs, one for logos and one for diagram images? Or would this be too many different folders?

alexbrand:
@jessicamaslyn I have no problem with that. It's the root dir that I'd like to declutter 😄

jessicamaslyn:
all set

alexbrand:
LGTM! We can rework this if necessary for docs site.

PR #406, opened 2017-03-09
Upgrade calico during node-upgrade and not at the end

### LINKED ISSUE #402: Upgrades: Calico Service
Currently we cleanup packages including the `kismatic-kubernetes-networking` in the beginning of the upgrade-node play. 

Because the docker engine restarts the `calico-node` container also restarts but the binaries it relies are on no longer available on the machine. 

For `online` upgrades this is problematic, even though we upgrade  the nodes serially, the networking doesnt start up until later in the process.

### LINKED ISSUE #416: Calico config job deletion fails in some cases because it does not exist
Currently, we always delete the calico job, which results in failures when it no longer exists (e.g. adding a worker).

Instead, we should only delete the job if it exists.

dkoshkin:
Fixes #402 by splitting out the ippool config and upgrading `calico/node` during the node upgrade step
Fixes #416 by not using a k8s job and instead relying on `calicoctl` 

Upgrade Calico during the `upgrade-nodes` play.
This is achieved by labeling the node as they are upgraded and installing the new DS calico on the labeled nodes.


ansible/upgrade-cluster-services.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-  - include: _calico.yaml play_name="Upgrade Cluster Network" serial="1" upgrading=true
+  - include: _calico-config.yaml
```
alexbrand: Does this mean that the network is still non-functional until we hit this?

dkoshkin: The only thing we do here is setup the cluster POD CIDR, that should already be present from the initial installation, this will ensure that if we for some reason updated the configs, Calico also gets updated.

alexbrand:
What will happen when we are upgrading from V+1 to V+2? All nodes will have the calico=true label.

PR #407, opened 2017-03-09
Fix retry output

alexbrand:
The number we get from ansible is retryCount + 1. Subtracting one so that in case of failure, the final retry is 20/20 instead of 20/21.


dkoshkin:
LGTM, why are there 2 explainers again?

alexbrand:
Even thought they somewhat share the event strings, each one has it's own way of formatting and writing to the terminal.

PR #414, opened 2017-03-10
Put kubeconfig in the dir used by kubectl

### LINKED ISSUE #411: Setup kubectl config file on all nodes that have kubectl installed
Currently, we install kubectl on worker nodes, however, we don't create a kube config file and it's necessary to pass all the cert flags to be able to talk to the API server. To make troubleshooting easier, we should configure kubectl on the nodes that have it.

alexbrand:
Moving the file was not enough, as we this didn't cover the upgrade scenario. 

I decoupled certs from the kubeconfig, and added a post task to remove the old kubeconfig from the old location.

Fixes #411


dkoshkin:
LGTM

PR #415, opened 2017-03-10
Test changes

### LINKED ISSUE #410: Dashboard accessibility test fails intermittently
We have seen a couple of these failures:
```
Using an upgraded cluster
should allow adding a new storage volume: PASSED
should allow adding a worker node: PASSED
should be able to deploy a workload with ingress: PASSED
should have an accessible dashboard: FAILED: exit status 1
should still be a highly available cluster after upgrade: PASSED
```

dkoshkin:
Fixes #410 

* Don't use `kismatic dashboard` command when testing dashboard access
* Move some tests into the "skunkworks" install test


alexbrand:
LGTM. I have a feeling that keeping track of and organizing those "sub" tests is going to be a pain point, but we'll see when we get there.

PR #418, opened 2017-03-13
Online upgrade: Error if all replicas are running on the same node

### LINKED ISSUE #413: Online checks should verify where the pods are running
Currently, we just verify that pods have at least one more sibling on the cluster. However, we don't verify where the pods are running. They could be running on the same node.

alexbrand:
Fixes #413


dkoshkin:
LGTM, I do wonder how this will work when upgrading multiple nodes in parallel but we'll cross that bridge when we get there.

alexbrand:
Reviewed the test run. The failures look like flakes. Merging.

PR #419, opened 2017-03-13
Fix add worker when using an internal registry

### LINKED ISSUE #271: Add worker command is broken when using an internal docker registry
The problem is the way we run preflight on the new node, and the fact that we need to build the entire cluster catalog.

alexbrand:
Fixes #271 

Cleaned up the add worker code and fixed the bug.


dkoshkin:
Nice LGTM

PR #421, opened 2017-03-13
Use k8s annotations instead of labels to support easier upgrades

### LINKED ISSUE #420: Kube does not support updating labels on DaemonSets without also changing the spec
Related to https://github.com/kubernetes/kubernetes/issues/26202

When rerunning `kubectl apply` with the same DS spec but different labels I get:

```
The DaemonSet "default-http-backend" is invalid: spec.template.metadata.labels: Invalid value: {"app":"default-http-backend","kismaticVersion":"1.3.0-alpha.0-51-ged12a81-dirty"}: `selector` does not match template `labels`
```

The labels will change when new versions of KET are released. 

I believe this is a DS issue and not on other resources.

dkoshkin:
Fixes #420 

Use annotations instead of labels.

### DIFF
--- a/ansible/roles/addon-network-policy/templates/policy-controller.yaml
+++ b/ansible/roles/addon-network-policy/templates/policy-controller.yaml
@@ -6,7 +6,8 @@ metadata:
   labels:
     k8s-app: calico-policy
     kubernetes.io/cluster-service: "true"
-    kismaticVersion: "{{ kismatic_short_version }}"
+  annotations:
+    kismatic/version: "{{ kismatic_short_version }}"
 spec:
   replicas: 1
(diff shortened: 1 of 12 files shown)


alexbrand:
This makes much more sense. LGTM!

alexbrand:
Does it make sense to use kismatic-version to keep consistent with other versions and annotations?

dkoshkin:
Yeah I dont like `kismaticVersion` either, I think something like `kismatic/version` is even closer to how k8s does it, ie `scheduler.alpha.kubernetes.io/critical-pod`. We probably dont need to do reverse DSN :)

PR #422, opened 2017-03-14
Lock down kuberang to v1.0.0

### LINKED ISSUE #399: Lock down kuberang version
Currently, we use the latest version of kuberang when building Kismatic.

Ideally, we would have this version locked down, similarly to how we have provision locked down.

alexbrand:
Fixes #399


dkoshkin:
LGTM!

PR #425, opened 2017-03-15
Set hostname field in the cni config

### LINKED ISSUE #424: Upgrading Calico: IP address is already in use by host
Calico relies on the hostname to define its endpoints.

A bug was introduced when switching to k8s pod method of running `calico-node` when not including `"hostname": "{{ inventory_hostname }}",` in the cni config.

This is a simple fix, however for versions prior to KET `v1.2` the `hostname` flag was not being set when configuring calico-node. This means that for a clusters configured with `modify_hostnames` and when the provided hostname != `hostname` the upgrade will not work.

dkoshkin:
Fixes #424 

Move cni config to outside the calico pods, need to be able to dynamically set the `hostname` field for each node and that is not possible when running as k8s pods.

This does not solve the problem that will be faced when a cluster was configured with:
- KET <v1.2.0
- `update_hosts_files=true`
- And provided `host` is different than what the `hostname` command returns

But it will cover all other cases.

The integration tests were modified to use the FQDN, to fix the issue described above.

### DIFF
--- a/ansible/_kube-config.yaml
+++ b/ansible/_kube-config.yaml
@@ -1,7 +1,7 @@
 ---
   - hosts: master:worker:ingress:storage
     any_errors_fatal: true
-    name: Generate kubectl config file
+    name: Generate Kubectl Config File
     remote_user: root
     become_method: sudo
     vars_files:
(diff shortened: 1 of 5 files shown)


dkoshkin:
Not needed because of #428, using real hostnames in the plan file is the correct way to solve this behavior.

PR #427, opened 2017-03-15
Improve failure message when calico or DNS fail to start up

### LINKED ISSUE #423: Get logs for components that fail to start up
Whenever a component such as Calico or KubeDNS fails to start up, we currently just show a failure message. 

In order to help diagnose issues, we will also show logs whenever a component fails to start up. Which logs to show are most likely dictated by the component.

alexbrand:
Fixes #423 

Getting the actual logs is challenging because we don't exactly what failed. However, listing the deployments and pods might help the user find the issue.


alexbrand:
This doesn't have much value after all.

PR #428, opened 2017-03-16
Use correct hostnames in integration tests and verify in preflight

### LINKED ISSUE #424: Upgrading Calico: IP address is already in use by host

dkoshkin:
Fixes #424 

We were using incorrect hostnames in some integration tests.
In `KET v1.2.x` we were setting the Calico `hostname` flag instead of relying on the OS reported one. Fo Calico upgrade to work correctly the planfile node name should be the correct `hostname`

### DIFF
--- a/ansible/roles/preflight/tasks/main.yaml
+++ b/ansible/roles/preflight/tasks/main.yaml
@@ -1,4 +1,9 @@
 ---
+  - name: verify hostname
+    fail: msg="provided hostname does not match reported hostname of {{ ansible_nodename }}"
+    when: "inventory_hostname != ansible_nodename"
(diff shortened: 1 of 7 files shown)


alexbrand:
LGTM!

PR #430, opened 2017-03-17
Perform full storage test for upgrades

alexbrand:
Instead of just creating a volume, perform a full storage test where we deploy a workload that attempts to use the storage.


dkoshkin:
LGTM!

PR #445, opened 2017-03-22
Show node roles when running online upgrade checks

### LINKED ISSUE #435: Upgrades UX: Display node roles when running checks

alexbrand:
Fixes #435 

Sample:

```
Validate Online Upgrade=============================================================
node001 [master etcd worker ingress]                                            [ERROR]
- This is the only master node in the cluster. Upgrading it will make the cluster unavailable.
- This node is acting as the load balanced endpoint for the master nodes. Upgrading it will make the cluster unavailable
- This node is part of an etcd cluster that has less than 3 members. Upgrading it will make the cluster unavailable.
```


dkoshkin:
LGTM!

PR #450, opened 2017-03-27
K8s 1.6 upgrade

dkoshkin:
TODO use prod repos and k8s 1.6 when it comes out.


ansible/roles/calico/templates/calico.yaml
```diff
@@ -11,29 +11,6 @@ data:
   # Configure the Calico backend to use.
   calico_backend: "bird"
 
-  # The CNI network configuration to install on each node.
```
alexbrand: Let's document why we are not putting this in the ConfigMap. I am sure we'll ask this question in the future. A comment in the PR description should work.

pkg/cli/upgrade.go
```diff
@@ -320,6 +320,14 @@ func upgradeNodes(out io.Writer, plan install.Plan, opts upgradeOpts, nodesNeedU
 		}
 	}
 
+	// Migate Kubernetes etcd to v3
+	// TODO in future KET releases need to check for a minimum version
+	// All KET releases with v1.6+ do not need this to run
+	util.PrintHeader(out, "Migrate: Kubernetes Etcd Cluster", '=')
```
alexbrand: Will this run again if the etcd nodes have already been migrated?

dkoshkin: It would but in my testing rerunning the migration was a noop

dkoshkin:
Moved CNI config out to an Ansible task from a ConfigMap. It is not possible to support rolling and to control when the config gets updated on each node individually when using a single ConfgiMap that all pods read from.

PR #453, opened 2017-03-29
Bump kuberang version to v1.1.0

alexbrand:


alexbrand:
Merging as test passed

PR #459, opened 2017-03-30
Support upgrades from v1.3.0

dkoshkin:
To support upgrading from `v1.3.0` including alpha and beta

* ~~Git tags are not valid semver and future dev builds `v1.3.0-#-alphanum` are not greater than `v1.3.0`, bumping patch version when detecting its a dev git tag~~
* Skip etcd2 upgrade for `alpha` and `beta` versions
* Skip etcdv3 migration for `v1.3.0` 
* Increase probes on control-plane components to match that of kubeadm

### DIFF
--- a/ansible/_calico-validate-node.yaml
+++ b/ansible/_calico-validate-node.yaml
@@ -1,0 +1,24 @@
+---
+  - hosts: master:worker:ingress:storage
+    any_errors_fatal: true
+    name: "{{ play_name | default('Validate Network Components') }}"
+    serial: "{{ serial_count | default('100%') }}"
+    remote_user: root
+    become_method: sudo
+    vars_files:
+      - group_vars/all.yaml
+
+    tasks:
+      - name: wait until the calico pod on this node is running
+        shell: kubectl get pods `kubectl get pods -l=k8s-app=calico-node --template {%raw%}'{{range .items}}{{if eq .spec.nodeName{%endraw%} "{{ inventory_hostname }}"{%raw%}}}{{.metadata.name}}{{"\n"}}{{end}}{{end}}'{%endraw%} -n kube-system` -n kube-system -o=jsonpath='{.status.phase}'
+        register: calico_pod_name
+        until: calico_pod_name|success and calico_pod_name.stdout == "Running"
+        retries: 10
+        delay: 6
+        failed_when: false # We don't want this task to actually fail (We catch the failure with a custom msg in the next task)
+        run_once: true
+      - name: fail if the calico pod on this node is not ready
+        fail:
+          msg: "Timed out waiting for calico pod on this node to be ready."
+        run_once: true
+        when: calico_pod_name|failed
--- a/ansible/roles/calico/tasks/main.yaml
+++ b/ansible/roles/calico/tasks/main.yaml
@@ -49,24 +49,3 @@
     retries: 20
     delay: 6
     when: calico_pod_name is defined and calico_pod_name.stdout != ""
-
-  - name: get desired number of calico pods
-    command: kubectl get ds calico-node -o=jsonpath='{.status.desiredNumberScheduled}' --namespace=kube-system --kubeconfig {{ kubernetes_kubeconfig_path }}
-    register: desiredPods
-    until: desiredPods|success
-    retries: 20
-    delay: 6
-    run_once: true
-  - name: wait until all calico pods are ready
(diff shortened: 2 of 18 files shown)


alexbrand:
We should have a conversation around how far back we want to support upgrades. I don't think it'll be reasonable to support an upgrade from v1.2.0 to v.3.0.0 for example. I am thinking this mainly for two reasons:
1. Judging from the special casing we have started to do here, I suspect we'll continue seeing those as we move forward, and it will probably be a maintainability nightmare.
2. The upgrade test matrix will never stop growing.

We might be able to address the first issue with implementing the concept of "migrations", where we keep version specific tasks, but it is not clear to me what this would entail.

pkg/cli/upgrade.go
```diff
@@ -316,22 +316,34 @@ func upgradeNodes(out io.Writer, plan install.Plan, opts upgradeOpts, nodesNeedU
 	}
 
 	// get all etcd nodes
-	etcdToUpgrade := install.NodesWithRoles(toUpgrade, "etcd")
-	// it's safe to upgrade one node etcd cluster from 2.3 to 3.1
-	// it will always be required for this version because all prior ket versions had a etcd2
+	etcdToUpgrade := make([]install.ListableNode, 0)
+	for _, n := range install.NodesWithRoles(toUpgrade, "etcd") {
+		// only transition nodes that are not 1.3.0...
+		if install.IsLessThanVersion(n.Version, "v1.3.0-alpha.0") {
```
alexbrand: Thinking out loud here... does it make sense to do these version checks in the ansible plays? Basically bringing the "Do I need to do X?" question closer to the implementation of "X".

I have been trying to think through failure cases, and it's unclear what would happen. For example:
What happens if the etcd2 -> etcd3 upgrade fails on the third etcd node? Would we try to do something one the first and second etcd nodes that would fail?

ikester:
Yes. I think we should support a very limited number of released versions. Maybe 2? Probably no more than that. That may also change over time, as the project matures and the changes between versions are less drastic. But that won't be the case for the foreseeable future.

ansible/roles/calico/tasks/main.yaml
```diff
@@ -32,6 +37,19 @@
     command: kubectl apply -f /etc/calico/calico.yaml --kubeconfig {{ kubernetes_kubeconfig_path }}
     run_once: true
 
+  - name: delete calico pod running on this node
+    command: kubectl delete pod {{ calico_pod_name.stdout }} -n kube-system
+    when: calico_pod_name is defined and calico_pod_name.stdout != ""
```
dkoshkin: Not needed as  `calico_pod_name` will already be undefind

ansible/_calico-validate-node.yaml
```diff
@@ -0,0 +1,24 @@
...
+    name: "{{ play_name | default('Validate Network Components') }}"
+    serial: "{{ serial_count | default('100%') }}"
+    remote_user: root
+    become_method: sudo
+    vars_files:
+      - group_vars/all.yaml
+
+    tasks:
+      - name: wait until the calico pod on this node is running
+        shell: kubectl get pods `kubectl get pods -l=k8s-app=calico-node --template {%raw%}'{{range .items}}{{if eq .spec.nodeName{%endraw%} "{{ inventory_hostname }}"{%raw%}}}{{.metadata.name}}{{"\n"}}{{end}}{{end}}'{%endraw%} -n kube-system` -n kube-system -o=jsonpath='{.status.phase}'
```
alexbrand: What happens if the kubectl call to get the pod name fails for some reason in the subshell? I think that would resolve to an empty string which would list all calico pods? I am not 100% sure though. It might make more sense to split this into separate tasks: One that gets the pod name, and one that gets the phase.

dkoshkin: If the inner command fails or returns an empty response, both cases the condition will not be met.

alexbrand: Thinking through this scenario:
* 1 master node, 1 worker node
* We want to verify calico-node on master node
* Calico-node on master was not created for whatever reason

I think this means the inner command will fail, but the validation will succeed. Is that the case? I know it's a contrived case, but it makes me wonder.

dkoshkin: This won't happen, when the inner one fails or returns an empty string the data struct returned will be a list and not a single object, with that the `jsonpath` will fail.

alexbrand: Got it!

PR #463, opened 2017-03-31
Show download link for integration test logs

alexbrand:


alexbrand:
This works to an acceptable degree. Merging.

PR #465, opened 2017-04-03
Don't print admin password to the terminal

### LINKED ISSUE #283: Dashboard command prints password to console on error
We should not print the dashboard password to the console.

alexbrand:
Fixes #283

### DIFF
--- a/pkg/cli/ip.go
+++ b/pkg/cli/ip.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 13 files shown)


pkg/cli/ip.go
```diff
@@ -41,15 +35,26 @@ func NewCmdIP(out io.Writer) *cobra.Command {
 	return cmd
 }
 
-func doIP(out io.Writer, planner install.Planner, opts *ipOpts) (string, error) {
+func doIP(out io.Writer, planner install.Planner, opts *ipOpts) error {
 	// Check if plan file exists
 	if !planner.PlanExists() {
-		return "", fmt.Errorf("plan does not exist")
+		return fmt.Errorf("plan does not exist")
```
dkoshkin: Can we make this error consistent across the CLI, maybe pull this out into a common function.

alexbrand: Good call

dkoshkin:
LGTM

PR #468, opened 2017-04-04
Dockerize kismatic installer

dkoshkin:
First steps in dockerizing `kismatic`

This does not include any work related to the provisioner.

For OSX and Linux place something like below in your $PATH, `/usr/local/bin/kismatic` should work:
```
#!/bin/bash
docker run -it $KISMATIC_EXTRA_MOUNTS -v "$(pwd)":/workdir -v "$HOME/.ssh/kismaticuser.key":/root/.ssh/kismaticuser.key -w /workdir apprenda/kismatic "$@"
```

This will mount the `$HOME/.ssh/kismaticuser.key` and the current path in the container.   
If you need to reference any other local directory use just `export KISMATIC_EXTRA_MOUNTS` with the mounts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,6 +125,25 @@ dist: vendor-ansible/out vendor-provision/out vendor-kuberang/$(KUBERANG_VERSION
 	tar -czf kismatic.tar.gz -C out .
 	mv kismatic.tar.gz out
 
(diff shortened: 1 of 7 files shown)


dkoshkin:
Closing for now

PR #475, opened 2017-04-06
Support machines with docker already installed

### LINKED ISSUE #474: Support nodes with existing docker installation
Kismatic should be able to detect if docker is already installed and skip the package installation.

Supporting machines with running docker will allow users to configure docker in custom ways and possibly use different versions of docker.

The only requirement for the machine will be to have the `docker` command line and docker be managed by `systemd` with a service `docker`. (this is the default configuration when installing docker packages)

dkoshkin:
Fixes #474


alexbrand:
We need to tell whether the docker installation is under our control, or the user's. A flag in the plan file is probably what we need.

integration/install_test.go
```diff
@@ -80,12 +80,14 @@ var _ = Describe("kismatic", func() {
 			})
 		})
 
-		Context("when deploying a cluster with all node roles", func() {
+		Context("when deploying a cluster with all node roles and docker already installed", func() {
```
alexbrand: Was there a reason for modifying this test instead of creating a new one that is specific to a pre-installed docker? We are potentially losing the coverage that this was providing before.

pkg/install/plan_types.go
```diff
@@ -82,6 +82,7 @@ type DockerRegistry struct {
 
 // Docker includes the configuration for the docker installation owned by KET.
 type Docker struct {
+	AllowInstallation bool `yaml:"allow_installation"`
```
alexbrand: In my mind, this not only controls installation, but configuration and management. In other words, I think we should support two use cases:
a. Kismatic installs docker and configures it with specific config options exposed in the plan file. Kismatic will also be responsible for upgrading docker when the time comes.
b. The user installs docker and configures it according to their requirements. The user is responsible for managing docker moving forward, including upgrades, etc.

With that in mind, this flag should also control configuration (e.g. docker storage). I am wondering if there is a more descriptive name for it.

ansible/_docker.yaml
```diff
@@ -10,6 +10,6 @@
 
     roles:
       - role: packages-docker
-        when: allow_package_installation|bool == true
+        when: allow_package_installation|bool == true and allow_docker_installation|bool == true
       - role: docker-storage
```
alexbrand: It seems like this shouldn't run if the user is managing the docker installation / configuration.

integration/install_test.go
dkoshkin: I think this scenario is already covered in the _skunkworks_ cluster.

alexbrand:
I am not convinced the flag should only control configuration of docker. In my mind, it should explicitly control installation, configuration and upgrades. I wonder if docker should be under a `container_runtime` option under the `features` section of the plan file being discussed in #518

dkoshkin:
Closing this for now.

PR #479, opened 2017-04-07
Fix gluster volume_mode

jeffmhastings:
The gluster `/data` directory was being created with incorrect permissions due to the fact that it was missing a leading `0`.


dkoshkin:
LGTM

alexbrand:
Thanks @jeffmhastings

PR #480, opened 2017-04-07
[WIP] Support for direct-lvm Docker storage driver

### LINKED ISSUE #109: Docker storage driver "devicemapper" is not suitable for production use
The storage driver that we are deploying is not suitable for production according to https://docs.docker.com/engine/userguide/storagedriver/selectadriver/

alexbrand:
Fixes #109 

This PR introduces support for setting up devicemapper in direct-lvm mode. The user is expected to provide a block device that will be used specifically for docker's storage needs.

The plan file now has a new section for configuring this:
```
docker:
  storage:
    direct_lvm:                          # Configure devicemapper in direct-lvm mode (RHEL/CentOS only).
      enabled: true
      block_device: "/dev/xvdb"                   # Path to the block device that will be used for direct-lvm mode. This device will be wiped and used exclusively by docker.
      enable_deferred_deletion: false    # Set to true if you want to enable deferred deletion when using direct-lvm mode.
```

TODO:
- [x] Automated test
- [ ] Look into making the play idempotent
- [ ] Run package checks conditionally, or remove current pkg checks and document limitation.
- [ ] Figure out why tasks are being printer instead of skipping

### DIFF
--- a/ansible/roles/docker-storage/tasks/main.yaml
+++ b/ansible/roles/docker-storage/tasks/main.yaml
@@ -1,21 +1,4 @@
 ---
-  - name: Stat the block device path
-    stat:
-      path: "{{ docker_direct_lvm_block_device_path }}"
-    register: block_device_stat
-  - name: Verify that the block device exists 
-    fail:
-      msg: "Block device specified for docker storage does not exist."
-    when: block_device_stat.stat.exists == False
-  - name: Verify that the provided path is a block device
-    fail:
-      msg: "{{ docker_direct_lvm_block_device_path }} is not a block device."
-    when: block_device_stat.stat.isblk == False
-  - name: Verify that the block device is not mounted
-    fail:
-      msg: "Block deviced specified for docker storage is currently mounted. This should be an unmounted, unused device"
-    with_items: ansible_mounts
-    when: item.device == "{{ docker_direct_lvm_block_device_path }}"
   - name: Install lvm2 package
     package:
       name: lvm2
--- a/ansible/roles/docker-storage/tasks/main.yaml
+++ b/ansible/roles/docker-storage/tasks/main.yaml
@@ -1,0 +1,41 @@
+---
+  - name: Stat the block device path
+    stat:
+      path: "{{ docker_direct_lvm_block_device_path }}"
+    register: block_device_stat
+  - name: Verify that the block device exists 
+    fail:
+      msg: "Block device specified for docker storage does not exist."
+    when: block_device_stat.stat.exists == False
+  - name: Verify that the provided path is a block device
+    fail:
+      msg: "{{ docker_direct_lvm_block_device_path }} is not a block device."
+    when: block_device_stat.stat.isblk == False
+  - name: Verify that the block device is not mounted
+    fail:
+      msg: "Block deviced specified for docker storage is currently mounted. This should be an unmounted, unused device"
+    with_items: ansible_mounts
+    when: item.device == "{{ docker_direct_lvm_block_device_path }}"
+  - name: Install lvm2 package
+    package:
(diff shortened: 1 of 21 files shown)


ansible/roles/docker-storage/tasks/main.yaml
```diff
@@ -0,0 +1,41 @@
+---
+  - name: Stat the block device path
```
alexbrand: I think we should run these checks in preflight.

ansible/roles/docker-storage/tasks/main.yaml
```diff
@@ -0,0 +1,41 @@
+---
```
alexbrand: Right now, this play is not idempotent. Not sure what it will take to make it idempotent.

pkg/install/validate.go
```diff
@@ -370,6 +371,31 @@ func (dr *DockerRegistry) validate() (bool, []error) {
...
+	return v.valid()
+}
+
+func (dlvm DockerStorageDirectLVM) validate() (bool, []error) {
+	v := newValidator()
+	if dlvm.Enabled {
+		if dlvm.BlockDevice == "" {
+			v.addError(errors.New("DirectLVM is enabled, but no block device was specified"))
+		}
+		if dlvm.BlockDevice != "" && dlvm.BlockDevice[0] != '/' {
```
dkoshkin: Can we use `filepath.IsAbs()` here to be consistent

alexbrand:
Going to add inspector check for new package dependency.

ansible/roles/docker-storage/tasks/main.yaml
dkoshkin: I agree

pkg/install/validate.go
alexbrand: Good catch

alexbrand:
Needs squash.

alexbrand:
> Do we also need the rules in the upgrade set?

I don't think so? We won't support migration to direct-lvm during an upgrade. At least in this iteration of this.

ansible/roles/preflight/tasks/direct_lvm_preflight.yaml
```diff
@@ -1,18 +1,18 @@
 ---
-  - name: Stat the block device path
+  - name: stat the block device path
     stat:
       path: "{{ docker_direct_lvm_block_device_path }}"
     register: block_device_stat
-  - name: Verify that the block device exists 
+  - name: fail if the block device exists
```
TimWoolford: fail if the block device does not exist

ansible/roles/preflight/tasks/direct_lvm_preflight.yaml
```diff
@@ -1,18 +1,18 @@
...
     fail:
       msg: "Block device specified for docker storage does not exist."
     when: block_device_stat.stat.exists == False
-  - name: Verify that the provided path is a block device
+  - name: fail if the provided path is a block device
```
TimWoolford: fail if the provided path is not a block device

ansible/roles/preflight/tasks/direct_lvm_preflight.yaml
```diff
@@ -1,18 +1,18 @@
...
     fail:
       msg: "{{ docker_direct_lvm_block_device_path }} is not a block device."
     when: block_device_stat.stat.isblk == False
-  - name: Verify that the block device is not mounted
+  - name: fail if the block device is not mounted
```
TimWoolford: fail if the block device is mounted

dkoshkin:
Merging, we can come beta to the remaining tasks past beta.

PR #481, opened 2017-04-07
Updating Ports Table in PLAN file

eriklg:
I updated the plan file with the following changes:

- Bold the headers of the ports table
- Added the port requirements for the kismatic inspector


dkoshkin:
Looks good thanks.

PR #482, opened 2017-04-08
Use ansible's modules for LVs and LVGs, and allow the play to be run multiple times

alexbrand:
In order to allow running `./kismatic install apply` against an existing cluster, we need some special handling in the docker storage play.

Specifically, we need to detect whether the thin pool logical volume has already been created. If it has, we will not:
1. Create the metadata logical volume required for a thin pool
2. Recreate the thin pool logical volume

### DIFF
--- a/ansible/roles/docker-storage/tasks/direct_lvm.yaml
+++ b/ansible/roles/docker-storage/tasks/direct_lvm.yaml
@@ -1,0 +1,70 @@
+---
+  - name: install lvm2 package
+    package:
+      name: lvm2
+      state: present
+  - name: create docker volume group
+    lvg:
+      pvs: "{{ docker_direct_lvm_block_device_path }}"
+      state: present
+      vg: docker
+  - name: create thinpool logical volume
+    lvol:
+      lv: thinpool
+      vg: docker
+      size: 95%VG # Use 95% of the volume group for this logical volume
+      opts: --wipesignatures y
+  
+  # The following two tasks are used to enable this play to be idempotent. We detect whether
+  # the logical volume (LV) is already a thin LV using `lvs`, which returns a list
+  # of all the LVs. `lvs` also displays the LV attributes in the
+  # lvattr field. According to the `lvs` manpage, the first bit in the attribute list determines
+  # the LV type. A value of "t" in this bit means "thin pool".
+  # See `man lvs` for more details.
+  # Sample output when the "thinpool" LV is of type thin pool:
+  # > lvs -o lvname,lvattr --no-headings
+  # thinpool     twi-aot---
+  - name: list logical volumes
+    command: lvs -o lvname,lvattr --noheadings
+    register: logical_volume_list
+  - name: determine if thin pool conversion is necessary
+    set_fact:
+      requires_thin_pool_conversion: True
+    # Regex: match whitespace, the name of the thin pool, whitespace, and then any character other than 't' or ' '. 
+    # If the item doesn't match, this means that the "thinpool" LV is already a thin pool, 
+    # and no conversion is necessary.
+    when: item | match("\s+thinpool\s+[^t ]+")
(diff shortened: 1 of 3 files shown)


ansible/roles/docker-storage/tasks/main.yaml
```diff
@@ -1,33 +1,4 @@
...
-    when: "ansible_os_family == 'RedHat'"
-  - name: apply lvm profile
-    command: lvchange --metadataprofile docker-thinpool docker/thinpool
-    when: "ansible_os_family == 'RedHat'"
-  - name: enable monitoring to ensure autoextend executes
-    command: lvchange --metadataprofile docker-thinpool docker/thinpool
-    when: "ansible_os_family == 'RedHat'"
+  - name: configure devicemapper in direct-lvm mode
+    include: direct_lvm.yaml
+    when: "ansible_os_family == 'RedHat' and docker_direct_lvm_enabled|bool == true"
```
dkoshkin: Will this skip the tasks when `docker_direct_lvm_enabled` is false I wasn't seeing that before

ansible/roles/docker-storage/tasks/direct_lvm.yaml
```diff
@@ -0,0 +1,70 @@
...
+  # The following two tasks are used to enable this play to be idempotent. We detect whether
+  # the logical volume (LV) is already a thin LV using `lvs`, which returns a list
+  # of all the LVs. `lvs` also displays the LV attributes in the
+  # lvattr field. According to the `lvs` manpage, the first bit in the attribute list determines
+  # the LV type. A value of "t" in this bit means "thin pool".
+  # See `man lvs` for more details.
+  # Sample output when the "thinpool" LV is of type thin pool:
+  # > lvs -o lvname,lvattr --no-headings
+  # thinpool     twi-aot---
+  - name: list logical volumes
```
dkoshkin: do any of these commands create directories or files we can use instead?

ansible/roles/docker-storage/tasks/main.yaml
alexbrand: Judging from the integration test logs, yes it does skip the direct_lvm.yaml include if lvm is not enabled.

ansible/roles/docker-storage/tasks/direct_lvm.yaml
alexbrand: Maybe? I did some research and couldn't find anything at all. Also cloned the `lvm2` repo and attempted to look into it there, but it's a huge project.

alexbrand:
I share the concern of relying on a command's stdout, but I am not familiar with any other option. I guess an alternative is to leave the play as is, and have a disclaimer somewhere saying that you can't run `apply` against an existing cluster that is using LVM. I am sure it will come back and bite us if we choose to do this.

The more I think about this, the more it seems like setting up the logical volumes is a provisioning step that `kismatic install apply` should not own.

alexbrand:
Thinking about this: If the installation fails *after* docker has been configured, the user will not be able to re-run the installer without making this play idempotent.

There are two possible paths we can take:
1. (Current plan) Own the creation of LVs and LVGs on the block storage device provided by the user. In this scenario, we need to make the play idempotent. The testing we have done on this is fairly limited in terms of time and environment setup (AWS).
2. Treat the creation of LVs and LVGs as a provisioning step. This means the user is responsible for provisioning and configuring the logical volumes that are to be consumed by docker. In the plan file, the user will provide the name of the LV instead of the path to the block storage device.

I am torn between the two - maybe leaning a bit on option #2 as I did originally on the hangout, but I don't have a strong preference.

@dkoshkin @swade1987 What do you think?

dkoshkin:
LGTM

alexbrand:
Discussed with @swade1987 and @dkoshkin and decided to continue down this path of configuring the LVs and LVGs.

Issue #483, opened 2017-04-08
Verify Hostname task fails if inventory hostname includes domain.


TimWoolford:
This worked fine in v1.2.2 but new verification fails in latest release.

```
[user@vm002734 kismatic-v1.3.1-beta0]$ ./kismatic install validate -f staging-slu.yaml

Validating==========================================================================
Reading installation plan file "staging-slu.yaml"                               [OK]
Validating installation plan file                                               [OK]
Validating SSH connectivity to nodes                                            [OK]
Configure Cluster Prerequisites                                                 [OK]
Run Cluster Pre-Flight Checks                                                   [ERROR]
- Task: verify hostname
  vm002756.bskyb.com: provided hostname does not match reported hostname of vm002756[ERROR]
  vm002757.bskyb.com: provided hostname does not match reported hostname of vm002757[ERROR]
  vm002758.bskyb.com: provided hostname does not match reported hostname of vm002758[ERROR]
  vm002754.bskyb.com: provided hostname does not match reported hostname of vm002754[ERROR]
  vm002755.bskyb.com: provided hostname does not match reported hostname of vm002755[ERROR]
  vm002752.bskyb.com: provided hostname does not match reported hostname of vm002752[ERROR]
  vm002753.bskyb.com: provided hostname does not match reported hostname of vm002753[ERROR]
  vm002759.bskyb.com: provided hostname does not match reported hostname of vm002759[ERROR]
  vm002760.bskyb.com: provided hostname does not match reported hostname of vm002760[ERROR]
  vm002761.bskyb.com: provided hostname does not match reported hostname of vm002761[ERROR]
error running playbook: error running ansible: exit status 2
[sonique@vm002734 kismatic-v1.3.1-beta0]$
```

alexbrand:
Calico performs auto-detection of hostnames, and it uses the result of `hostname` to do so. We have seen issues in the past with calico where machines that had the FQDN configured as the machine's `hostname`. This is what led us to add this validation. 

Is it possible that the machine is misconfigured? `hostname` should return the hostname, and `hostname -f` should return the FQDN.

TimWoolford:
```

alexbrand:
Excellent. Thanks @TimWoolford!

alexbrand:
I think I misread the original issue. It seems like you are using FQDNs in the plan file? Is that the case? If so, is there a reason why you need to use FQDNs?

TimWoolford:
Yes, I am using FQDNs.

[...]

PR #484, opened 2017-04-08
verify hostname to allow hostname with or without domain

TimWoolford:
Fix for - https://github.com/apprenda/kismatic/issues/483

### DIFF
--- a/ansible/roles/preflight/tasks/main.yaml
+++ b/ansible/roles/preflight/tasks/main.yaml
@@ -1,7 +1,7 @@
 ---
   - name: verify hostname
     fail: msg="provided hostname does not match reported hostname of {{ ansible_nodename }}"
-    when: "inventory_hostname != ansible_nodename"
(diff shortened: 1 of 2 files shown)


alexbrand:
Initially, we added this hostname verification when working on upgrades. The reason for adding this check was that a previous KET version specified the hostname in calico's configuration. In newer versions we stopped doing this. This could result in the following upgrade issue:
1. KET version X set the hostname option in calico's config using the value entered in the plan file
2. Upgrade to KET version X + 1. Calico auto-detects hostname.
3. If the detected hostname is different than the hostname set in option 1, calico would fail to start up.

In order to prevent this issue, we added the hostname validation to make sure that the hostname that calico would auto-detect would be the same as the one specified in the plan file.

Moving forward, this validation is really about making sure that the user did not use the wrong hostname (or FQDN) for the machine with the given IP.

Issue #487, opened 2017-04-12
Cannot specify custom IP address range for Services or Pods


Anarethos:
Hi,

I currently have an issue with Kismatic (1.2.2 and 1.3.1). If I change the IP address range for the Services or Pods in the "kismatic-cluster.yaml", the deployment fails.

Everything is running under Vagrant on a Windows host.

The "smoke" test fails. I end up with no kubeconfig file at the end and cannot do anything. I have to revert my Vms and use the default IP settings.

All my Vms are in the 192.168.56.0/24 network.

Local hosts file are correctly set between all Vms.

Master1 + Node1 + Node2 ==> Oracle Linux 7.2 (with modified os-release file to act as a CentOS Linux).
Management node / Ansible ==> Ubuntu Trusty
Host : Windows 10 (with Virtual Box + Vagrant)

We also have the same problem with out lab environemnt, using Oracle Linux under OVM.

Anarethos:
Here is the error message when I change the default settings.

[ERROR]
- Task: fail if any DNS pods are not ready
  master1: Timed out waiting for DNS pods to be in the ready state.
[ERROR]
error installing: error running playbook: error running ansible: exit status 2

dkoshkin:
@Anarethos  What are you using for `master.load_balanced_fqdn` and `master.load_balanced_short_name`, we intentionally do not mount the `/etc/hosts` files in the kube-dns pod.

Those addresses need to be resolvable without relying on the `hosts` file.

dkoshkin:
You can also run `kismatic diagnose` to collect cluster data. This will include numerous things including _systemd_ services and component logs.

Anarethos:
I use the IP of the single Master.
```
load_balanced_fqdn: "192.168.56.10"
load_balanced_short_name: "192.168.56.10"  
```

I will retry the deployment and, then, use the diagnose command and post the result here.

Like I said, if I use the default IP range 
```
pod_cidr_block: 172.16.0.0/16
service_cidr_block: 172.17.0.0/16 
```
it works.

But if I chnage it to
```
pod_cidr_block: 172.20.0.0/16
service_cidr_block: 172.21.0.0/16 
```

it doen't works and fails.

Anarethos:
I'm using the following Vms in my scenario. The "ansible" Vm if also my deplyment client.

```
master1 : 192.168.56.10
node1 : 192.168.56.30
node2 : 192.168.56.31
ansible : 192.168.56.40
```

[...]

PR #493, opened 2017-04-12
Switch to using a different services CIDR from Docker’s

### LINKED ISSUE #492: Modify default service CIDR
Currently KET uses `172.17.0.0/16` as the default services CIDR.
Docker is already using this block for its own [networking](https://docs.docker.com/engine/userguide/networking/).

We have seen a few issues come up due to this: https://github.com/apprenda/kismatic/issues/487 and https://github.com/apprenda/kuberang/issues/32.

This should not have an impact on existing clusters but it is recommended for users to use the next version of KET with this fix.

dkoshkin:
Fixes #492, #487 

Switch to using `172.20.0.0/16` services CIDR


alexbrand:
LGTM

PR #494, opened 2017-04-12
Fix v1.3.x integration tests with a bad CIDR

dkoshkin:
A few tests were failing when using KET v1.3.x, need to use the old `172.16.0.0/16` CIDR for these tests.


alexbrand:
LGTM

PR #496, opened 2017-04-14
Add kubernetes-rbac proposal

alexbrand:
Mainly the rbac proposal, but also:
* I am introducing a status field on our proposals so that we can keep track of what has been implemented and what has not (Inspired by upstream project doing this as well).
* I also renamed all our current proposals to lower case.

cc: @swade1987 @codymccain @robbfoster @TimWoolford @WillPlatnick

### DIFF
--- a/proposals/ingress.md
+++ b/proposals/ingress.md
@@ -1,5 +1,7 @@
 # Basic Ingress for Kismatic
 
+Status: Implemented
+
 The very first question anybody has upon installing Kubernetes is “how do I access my workloads?”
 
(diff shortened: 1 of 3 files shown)


swade1987:
Is the plan to run something like [Dex](https://github.com/coreos/dex) under the covers?

Also, how will the RBAC JSON file be provided at the initial bootstrap stage?

Finally how will this be modified over time going forward?

alexbrand:
This proposal does not cover dex or any third party integration. This is only about enabling RBAC, which is configured via the API and not a JSON file.

alexbrand:
Merging proposal.

Issue #498, opened 2017-04-14
generated certs have an empty DNS SAN


kwtalley:
When certs are generated at deployment, there is an empty DNS SAN:

X509v3 Subject Alternative Name:
                DNS:kubernetes, DNS:kubernetes.default, DNS:kubernetes.default.svc, DNS:kubernetes.default.svc.cluster.local, DNS:router, **DNS:**, IP Address:127.0.0.1, IP Address:10.200.0.1, IP Address:10.2.0.11, IP Address:10.2.0.11, IP Address:10.2.0.11

This can break some applications (deis) which attempt to contact the apiserver via https due to cert/hostname mismatch.

alexbrand:
@kwtalley Was this the certificate for a specific node?

alexbrand:
I am unable to repro after the changes in #499, which included refactoring of cert generation code. I am going to attempt to repro without those changes now.

alexbrand:
Thanks @kwtalley

PR #499, opened 2017-04-17
Enable RBAC support in Kubernetes API server

### LINKED ISSUE #471: Enable RBAC support
One of the flagship features of Kubernetes 1.6 was the graduation of RBAC authorization to beta. With that in mind, we should allow our users to use RBAC as their authorization method.

alexbrand:
Implementation of proposal #496 

Did some cleanup of the cert generation code.

Added integration test that exercises RBAC functionality.

Fixes #471


dkoshkin:
Looks good, do you think #498 should be part of this also?

alexbrand:
Addressed #498 as a separate thing as it's unrelated to RBAC.

dkoshkin:
LGTM!

Issue #478, opened 2017-04-07
Allow adding options to apiserver runtime-config


TimWoolford:
I would like to add extra options to the runtime-config of the api-server. 

This would be very useful as an optional value in the cluster yaml.

In particular, we are using CronJobs and need `batch/v2alpha1=true`

dkoshkin:
Closed woth #500

PR #500, opened 2017-04-18
Override config options for api-server

TimWoolford:
Changes to allow passing of runtime-config options to the Api Server from the plan file.

Ref: https://github.com/apprenda/kismatic/issues/478

'Tested' on our internal vSphere Data Centre.

My first attempt at writing in Go, so any feedback much appreciated.


alexbrand:
Thanks @TimWoolford!

I am thinking it might be better to expose a top-level `api_server` struct in the plan file, which contains the `runtime_config map[string]bool`. This would allow us to expose more options in the future.

E.g.
```
cluster:
  api_server:
    runtime_config: {"beta/v2api":true, "alpha/v1api": true}
```

@dkoshkin thoughts?

dkoshkin:
I agree with @alexbrand , although I think we should use `map[string]string` to be consistent with upstream.

TimWoolford:
How about this? @alexbrand @dkoshkin

alexbrand:
Looks great @TimWoolford 

However, I am now wondering if we should relax the structure here and make this more powerful. That is, allow overriding _any_ API Server flag. I am not entirely sure what this would look like in implementation though, as the flags defaults are set in the ansible template. 

The upside is pretty clear: it would make kismatic more powerful and enable power-users to customize their cluster. The downsides that I can think of: higher potential for a poor overall experience if the user inputs invalid flags, higher variability in deployed clusters (pain for debugging).

What do you think?

dkoshkin:
I think this is a good starting point, lets see how this works first and then we can expand.
I do like the structure though, this should be sufficient to expand in the future.
```
+  api_server:
 +    runtime_config:
 +      "batch/v2alpha1": true
 +      "extensions/v1beta1/ingress": false
```

docs/PLAN.md
```diff
@@ -319,3 +319,18 @@ udp:0-65535</td>
...
+## Kubernetes Api Server Options
+
+Kubernetes api versions and extensions can be enabled or disabled as required in the plan file
+
+```
+cluster:
+...
+  api_server:
+    runtime_config:
+      "batch/v2alpha1": true
```
dkoshkin: Can you modify this example to be a string `"true"`.

pkg/install/plan_types.go
```diff
@@ -288,3 +295,46 @@ func hasIP(nodes *[]Node, ip string) bool {
...
 }
+
+func (apiServer ApiServer) RuntimeConfig() string {
+	var configOptions = apiServer.ApiRuntimeConfigOptions
+
+	if len(configOptions) == 0 {
+		configOptions = make(map[string]string)
+	}
+
+	enableIfNotSet(configOptions, "extensions/v1beta1", "true");
```
dkoshkin: `enableIfNotSet` should be renamed as it is possible to not be _enabling_ it by passing `"false"`

pkg/install/test/cluster-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    update_hosts_files: false
+  certificates:
+    expiry: 17520h
+  ssh:
+    user: kismatic
+    ssh_key: .ssh/id_rsa
+    ssh_port: 22
+  api_server:
+    runtime_config:
+      "beta/v2api": true
```
dkoshkin: `true` probably needs to change to `"true"`

alexbrand:
Nit: can we rename all `Api` appearances in code to `API`?

TimWoolford:
I thought it was a bit early talk about Religion or Politics or CamelCase for Acronyms :oP

TimWoolford:
@alexbrand Api -> API for consistency (which is usually better than being right).

Am interested in increasing the configurability of the API (:stuck_out_tongue_winking_eye:) server but, as @dkoshkin suggests, its probably better tackled as a separate issue.

TimWoolford:
Been thinking this over.

If I change the runtime-config entry in the plan file to be the same comma separated value-pair as is passed to the API server, we can then parse that to ensure the KET defaults are present or overridden.

All API server config options could then be configured the same way in the plan file. Having a special structure for runtime-config would require changes for any similar type of properties (`feature-gates` etc).

_Random example:_ 
```
  api_server:
    runtime-config: "batch/v2alpha1=true,extensions/v1beta1/ingress=false"
    audit-log-maxage: "5"
    event-ttl: "24h0m0s"
```

I'd think there are some values that should be protected and cause the plan validation to fail if a user attempts to override them. If needed these should be set using some more global setting - eg `secure-port`.

Does this match with your thinking @alexbrand, @dkoshkin ?

Should I start again with a new PR, or refactor this?

alexbrand:
I think that is what I was suggesting in my previous comment https://github.com/apprenda/kismatic/pull/500#issuecomment-295228838

We could have a `map[string]string`, where the key is the API server flag name, and the value is the value that should be passed in.

alexbrand:
@TimWoolford Is this ready for a review? Sorry we haven't been on top of this one. It also looks like something went wrong with merging master?

TimWoolford:
Hi @alexbrand, yes, I think this is pretty much ready. There's some discussion to be had over the validation, as plan failure stops all processing. Should we totally prevent someone from spinning up a cluster that will never start?

I was trying to do something a bit more sophisticated, but the separation of concerns is much clearer if the majority of the work is done on the ansible side. This version is much simpler.

alexbrand:
Yeah, I think it's a good idea to prevent certain flags from being user-configurable. 

This change also opens us up to a lot more variability in the clusters installed with KET, which we'll have to account for when thinking about upgrades.

One side effect that might be undesirable is that the flags in the config file will not be sorted. Not sure if there's a way around that.

Overall this looks good to me, but need to think more about the upgrade scenario.

alexbrand:
Another thing to think about is that we might need to set some flags under specific conditions. For example, if we decide to deploy an OIDC provider, we will need to own the OIDC related flags.

pkg/install/api_server_config.go
```diff
@@ -0,0 +1,41 @@
+package install
+
+import (
+	"fmt"
+)
+
+type APIServerConfig struct {
+	RawData map[string]string `yaml:"api_server"`
```
alexbrand: Was there a reason for wrapping the map in a struct?

TimWoolford: This allows us to run validation on the struct and fits better into the existing validation pattern

alexbrand: Would it make sense to do `type APIServerConfig map[string]string`?

ansible/group_vars/all.yaml
```diff
@@ -155,6 +155,33 @@ kubernetes_certificates:
...
+  "client-ca-file":  "{{ kubernetes_certificates.ca }}"
+  "enable-swagger-ui": "true"
+  "etcd-cafile":  "{{ kubernetes_certificates.ca }}"
+  "etcd-certfile":  "{{ kubernetes_certificates.etcd_client }}"
+  "etcd-keyfile":  "{{ kubernetes_certificates.etcd_client_key }}"
+  "etcd-servers":  "{{ etcd_k8s_cluster_ip_list }}"
+  "insecure-bind-address": "127.0.0.1"
+  "insecure-port": "8080"
+  "runtime-config": "extensions/v1beta1=true,extensions/v1beta1/networkpolicies=true"
+  "secure-port": "6443"
```
alexbrand: We actually have a variable in this file called `kubernetes_master_secure_port`. We should point to that here.

pkg/install/api_server_config.go
```diff
@@ -0,0 +1,41 @@
...
+	"client-ca-file",
+	"etcd-cafile",
+	"etcd-certfile",
+	"etcd-keyfile",
+	"etcd-servers",
+	"service-account-key-file",
+	"service-cluster-ip-range",
+	"tls-cert-file",
+	"tls-private-key-file",
+	"v",
```
alexbrand: Was there a reason for protecting the log level? Seems like something that should be configurable.

pkg/install/api_server_config.go
```diff
@@ -0,0 +1,34 @@
...
+	"basic-auth-file",
+	"client-ca-file",
+	"etcd-cafile",
+	"etcd-certfile",
+	"etcd-keyfile",
+	"etcd-servers",
+	"service-account-key-file",
+	"service-cluster-ip-range",
+	"tls-cert-file",
+	"tls-private-key-file",
```
alexbrand: I think we need to protect the `secure_port`. @dkoshkin were there other fields you thought we should protect?

dkoshkin: I agree we should protect `secure-port` but I would argue we shouldn't `authorization-policy-file` `basic-auth-file` there are use cases where a user doesn't want basic auth or wants to provide their own process of creating those files. @alexbrand thoughts?

pkg/ansible/clustercatalog.go
```diff
@@ -21,6 +21,8 @@ type ClusterCatalog struct {
 	KuberangPath              string `yaml:"kuberang_path"`
 	LoadBalancedFQDN          string `yaml:"kubernetes_load_balanced_fqdn"`
 
+	APIServerConfig           map[string]string `yaml:"kubernetes_api_server_config_overrides"`
```
dkoshkin: Do we want the map to be a subsection of `kubernetes_api_server_config_overrides`, maybe something like:
```
kube_apiserver:
  options_overrides:
    ...
```
Also note `kube_apiserver` consistent with the name in k8s docs. `options_overrides` comes from their "Options". https://kubernetes.io/docs/admin/kube-apiserver/

pkg/install/api_server_config.go
alexbrand: Agreed

pkg/ansible/clustercatalog.go
dkoshkin: @alexbrand what do you think?

TimWoolford: Hi @dkoshkin, I felt that this was more of a kismatic convention. Most other vars in `clustercatalog` follow _kubernetes_...._ . 

Personally I find _api_server_ easier on the eye than _apiserver_. 

I think you're right about _options_ over _config_ and will update accordingly

alexbrand: I think @dkoshkin meant to put his comment in the plan file struct, and not the cluster catalog?

dkoshkin: @TimWoolford sorry I left the comment on the wrong file, I meant it to do it in the `plan_types.go`
Basically it would still be under `cluster` but the `options_overrides` would be its own section.
```
...
cluster
  kube_apiserver:
    options_overrides:
    ...
```
The cluster_catalog could have a similar struct and then the ansible plays can use that.

ansible/roles/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -17,39 +17,18 @@ spec:
...
-      - --runtime-config=extensions/v1beta1=true,extensions/v1beta1/networkpolicies=true
-      - --secure-port=6443
-      - --service-account-key-file={{ kubernetes_certificates.service_account_key }}
-      - --service-cluster-ip-range={{ kubernetes_services_cidr }}
-      - --tls-cert-file={{ kubernetes_certificates.api_server }}
-      - --tls-private-key-file={{ kubernetes_certificates.api_server_key }}
-      - --v=2
+{% endfor -%}
     ports:
-    - containerPort: 6443
```
dkoshkin: The `containerPort` should always be 6443 and 8080, only the `hostPorts` will change

ansible/roles/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -62,7 +41,7 @@ spec:
       httpGet:
         host: 127.0.0.1
         path: /healthz
-        port: 8080
```
dkoshkin: Can we use the name of the port here `local`

dkoshkin:
Ignore the deleted comments about the ports, both of the flags are blocked.

dkoshkin:
LGTM after the merge conflict.

alexbrand:
Looks good to me as well. Thanks @TimWoolford!

TimWoolford:
Just validating with a fresh cluster install, will fix conflicts and update PLAN.md to reflect kube_apiserver changes

TimWoolford:
Ok, that _works on my machine_. Thanks for the feedback @alexbrand & @dkoshkin!

ansible/roles/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -17,39 +17,18 @@ spec:
...
+      - --{{ option[0] }}={{ option[1] }}
 {% endif %}
-      - --runtime-config=extensions/v1beta1=true,extensions/v1beta1/networkpolicies=true
-      - --secure-port=6443
-      - --service-account-key-file={{ kubernetes_certificates.service_account_key }}
-      - --service-cluster-ip-range={{ kubernetes_services_cidr }}
-      - --tls-cert-file={{ kubernetes_certificates.api_server }}
-      - --tls-private-key-file={{ kubernetes_certificates.api_server_key }}
-      - --v=2
+{% endfor -%}
```
dkoshkin: When I was testing this the api manifest was not a proper format, the `ports` tag did not have the proper indents.

This happens because of the trailing  `-` in `{% endfor -%}`, removing that fixed it.
@TimWoolford did you see a similar issue during your testing?

pkg/install/api_server_options.go
```diff
@@ -0,0 +1,36 @@
...
+	"etcd-servers",
+	"insecure-port",
+	"secure-port",
+	"service-account-key-file",
+	"service-cluster-ip-range",
+	"tls-cert-file",
+	"tls-private-key-file",
+}
+
+func (options *APIServerOptions) validate() (bool, []error) {
```
dkoshkin: Instead of printing a line per error can all `overrides` be in a single line.
```
func (options *APIServerOptions) validate() (bool, []error) {
	v := newValidator()
	overrides := make([]string, 0)
	for _, opt := range protectedOptions {
		_, found := options.Overrides[opt]
		if found {
			overrides = append(overrides, opt)
		}
	}
	if len(overrides) > 0 {
		v.addError(fmt.Errorf("KubeApiserver Options %v should not be overridden", overrides))
	}
	return v.valid()
}
```
This will look something like
```
- KubeApiserver Options [advertise-address apiserver-count] should not be overridden
```

ansible/group_vars/all.yaml
```diff
@@ -157,6 +157,34 @@ kubernetes_certificates:
...
+  "bind-address": "0.0.0.0"
+  "client-ca-file":  "{{ kubernetes_certificates.ca }}"
+  "enable-swagger-ui": "true"
+  "etcd-cafile":  "{{ kubernetes_certificates.ca }}"
+  "etcd-certfile":  "{{ kubernetes_certificates.etcd_client }}"
+  "etcd-keyfile":  "{{ kubernetes_certificates.etcd_client_key }}"
+  "etcd-servers":  "{{ etcd_k8s_cluster_ip_list }}"
+  "insecure-bind-address": "127.0.0.1"
+  "insecure-port": "{{ kubernetes_master_insecure_port }}"
+  "kubelet-preferred-address-types": "{% if modify_hosts_file is defined and modify_hosts_file|bool == true %}InternalIP,ExternalIP,Hostname{% endif %}"
```
dkoshkin: I was worried about this but with `{% if option[1] is defined and option[1] | string | length > 0 %}` works out great.

pkg/install/test/cluster-config.yaml
```diff
@@ -0,0 +1,20 @@
+cluster:
+  name: my_cluster_name
+  admin_password: secret_admin_password
+  allow_package_installation: true
```
dkoshkin: Can you rename this field to `disable_package_installation` default `false`.

pkg/install/test/cluster-config.yaml
```diff
@@ -0,0 +1,20 @@
...
+  disconnected_installation: false
+  networking:
+    type: overlay
+    pod_cidr_block: 172.16.0.0/16
+    service_cidr_block: 172.17.0.0/16
+    policy_enabled: false
```
dkoshkin: This is an old field, please remove.

pkg/install/api_server_options.go
dkoshkin: Looks like the test string will also need to change.

ansible/roles/kube-apiserver/templates/kube-apiserver.yaml
TimWoolford: oops, yes I did have a couple of glitches like that. Thought they were all fixed 😞

dkoshkin:
Closes #478.
@TimWoolford just verified this manually this looks great! Thanks for all you work.

PR #501, opened 2017-04-18
Add http proxy options in the plan file

dkoshkin:


pkg/install/plan.go
```diff
@@ -221,4 +221,7 @@ var commentMap = map[string]string{
 	"direct_lvm":               "Configure devicemapper in direct-lvm mode (RHEL/CentOS only).",
 	"block_device":             "Path to the block device that will be used for direct-lvm mode. This device will be wiped and used exclusively by docker.",
 	"enable_deferred_deletion": "Set to true if you want to enable deferred deletion when using direct-lvm mode.",
+	"http_proxy":               "A comma-separated list of http domains proxy should be used for.",
```
alexbrand: Would it make sense to add something like: "This is equivalent to setting the 'http_proxy' environment variable"?

alexbrand:
Fixes #495

PR #502, opened 2017-04-18
Add package_repo_urls option in the plan file

dkoshkin:
To better support an installation with no internet access (disconnected installation), the ability to specify the location of an internal package repository has been added to the plan file. This enables the user to install kismatic without having to configure the repos on the nodes prior to installing.

Given that users are free to setup their own repositories, we support the addition of multiple package repositories via a comma-separated list.

```
cluster:
  package_repo_urls:
```
`package_repo_urls` is a comma-separated list of RPM or DEB repositories.

It's important to keep in mind that the default repositories (Kismatic, Gluster, etc) will **not** be configured when custom repositories are provided.


ansible/roles/packages-repo/tasks/main.yaml
```diff
@@ -17,9 +17,23 @@
...
+    when: ansible_os_family == 'RedHat' and package_repo_url is defined and package_repo_url is not empty
+  - name: add Gluster yum repository
+    yum_repository:
+      name: gluster
+      description: Gluster 3.8
+      baseurl: "http://buildlogs.centos.org/centos/7/storage/x86_64/gluster-3.8/"
+      gpgcheck: no
+    when: ansible_os_family == 'RedHat' and 'storage' in group_names
   - name: clean kismatic yum cache
     command: 'yum --disablerepo="*" --enablerepo=kismatic makecache'
```
alexbrand: Do we need a separate task here to clean the `kismatic-custom` yum repo? I believe this will fail if the `kismatic` repo was not configured.

ansible/roles/packages-repo/tasks/main.yaml
```diff
@@ -17,9 +17,23 @@
...
+      description: Kismatic Packages
+      baseurl: "{{ package_repo_url }}"
+      gpgcheck: no
+    when: ansible_os_family == 'RedHat' and package_repo_url is defined and package_repo_url is not empty
+  - name: add Gluster yum repository
+    yum_repository:
+      name: gluster
+      description: Gluster 3.8
+      baseurl: "http://buildlogs.centos.org/centos/7/storage/x86_64/gluster-3.8/"
+      gpgcheck: no
```
alexbrand: Does this fail if we enable gpgchecking? I can't remember why we disabled it.

pkg/install/plan.go
```diff
@@ -194,6 +194,7 @@ func generateAlphaNumericPassword() (string, error) {
 var commentMap = map[string]string{
 	"admin_password":             "This password is used to login to the Kubernetes Dashboard and can also be used for administration without a security certificate",
 	"allow_package_installation": "When false, installation will not occur if any node is missing the correct deb/rpm packages. When true, the installer will attempt to install missing packages for you.",
+	"package_repo_url":           "A URL pointing to a package repo containing kismatic packages.",
```
alexbrand: This repo must contain all the packages required for the installation, right? Including transitive dependencies.

Maybe something like: "The URL of the repository that should be used during installation. This repository must contain the kismatic packages, and all their transitive dependencies."

ansible/roles/packages-repo/tasks/main.yaml
```diff
@@ -17,9 +17,23 @@
       gpgkey: "{{ yum_gpg_key_url }}"
       gpgcheck: yes
     when: ansible_os_family == 'RedHat' and kismatic_result.status is defined and kismatic_result.status == 200
```
alexbrand: Seems like providing a custom repo url should prevent this repo from being configured.

ansible/roles/packages-repo/tasks/main.yaml
```diff
@@ -1,54 +1,152 @@
 ---
```
alexbrand: I wonder if it makes sense for us to split this file into two: yum and apt

ansible/roles/packages-repo/tasks/main.yaml
```diff
@@ -1,54 +1,152 @@
...
+          (kismatic_result is defined and kismatic_result.status is defined and kismatic_result.status == 200)
+
+  - name: add Kismatic custom deb keys
+    apt_key:
+      url: "{{ item }}"
+      validate_certs: no
+    when: >
+          ansible_os_family == 'Debian' and
+          (gpg_keys is defined and gpg_keys is not empty) and
+          (gpg_key_is_a_file is unedefined or (gpg_key_is_a_file is defined and gpg_key_is_a_file|bool == no))
```
alexbrand: typo: unedefined

ansible/roles/packages-repo/tasks/main.yaml
```diff
@@ -1,54 +1,152 @@
...
+    with_items: "gpg_keys.split(',')"
+  - name: add Kismatic custom deb keys from file
+    apt_key:
+      url: "/tmp/key{{ item.0 }}"
```
alexbrand: should this be `file:` instead of `url:`?

pkg/install/plan.go
```diff
@@ -194,6 +194,8 @@ func generateAlphaNumericPassword() (string, error) {
 var commentMap = map[string]string{
 	"admin_password":             "This password is used to login to the Kubernetes Dashboard and can also be used for administration without a security certificate",
 	"allow_package_installation": "When false, installation will not occur if any node is missing the correct deb/rpm packages. When true, the installer will attempt to install missing packages for you.",
+	"package_repo_url":           "The URL of the repository that should be used during installation. This repository must contain the kismatic packages and all their transitive dependencies.",
+	"package_gpg_keys":           "Comma-separated list of URLs or absolute paths of the GPG public keys used to sign required packages. Kismatic: `https://kismatic-packages-[rpm,deb].s3-accelerate.amazonaws.com/public.key`, Gluster: https://download.gluster.org/pub/gluster/glusterfs/3.8/3.8.7/rsa.pub",
```
alexbrand: Are the URLs at the end of the comment examples? Not sure what the purpose of having them is.

dkoshkin: No these are the ones that are actually needed. Thinking about it, we can just include these keys in KET.

ansible/_docker-registry-container.yaml
```diff
@@ -6,4 +6,8 @@
     run_once: true
```
alexbrand: Should we rename this file to `_docker-registry.yaml`?

ansible/roles/packages-repo-offline/tasks/main.yaml
```diff
@@ -0,0 +1,48 @@
+---
+
+  - name: copy Kismatic GPG keys to /tmp
```
alexbrand: Wondering why this is necessary

ansible/roles/packages-repo-offline/tasks/main.yaml
```diff
@@ -0,0 +1,48 @@
+---
```
alexbrand: Calling this file `packages-repo-custom` or something makes more sense to me. Using a custom repo doesn't necessarily mean there's no internet.

ansible/roles/packages-repo/tasks/main.yaml
```diff
@@ -1,54 +1,53 @@
 ---
   # YUM
-  - name: ping Kismatic yum repository
```
alexbrand: YES!

pkg/install/plan.go
```diff
@@ -194,6 +194,7 @@ func generateAlphaNumericPassword() (string, error) {
 var commentMap = map[string]string{
 	"admin_password":             "This password is used to login to the Kubernetes Dashboard and can also be used for administration without a security certificate",
 	"allow_package_installation": "When false, installation will not occur if any node is missing the correct deb/rpm packages. When true, the installer will attempt to install missing packages for you.",
+	"package_repository_urls":    "Comma-separated list of URLs of the repositories that should be used during installation. These repository must contain the kismatic packages and all their transitive dependencies.",
```
alexbrand: s/These repository/These repositories

ansible/_packages-repo.yaml
```diff
@@ -8,3 +8,9 @@
 
     roles:
       - role: packages-repo
+        when: >
+              disconnected_installation|bool != true and
```
alexbrand: Addressed via docs.

ansible/kubernetes.yaml
```diff
@@ -13,10 +13,8 @@
   - include: _etcd-networking.yaml
   # docker
   - include: _docker.yaml
-  - include: _docker-registry-container.yaml
+  - include: _docker-registry.yaml
     when: setup_internal_docker_registry|bool == true
```
alexbrand: The docker registry images will only be pushed if we are setting up an internal repo, but not if the user has pointed us to their own repo. This seems like an issue?

ansible/roles/packages-etcd/tasks/main.yaml
```diff
@@ -21,9 +38,13 @@
...
-  - name: install curl package
-    package: name=curl state=present
-    register: curl_installation_rpm
-    until: curl_installation_rpm|success
+  - name: install curl deb package
+    apt:
+      name: curl
+      allow_unauthenticated: '{% if package_repository_urls is defined and package_repository_urls != "" %}yes{% else %}no{% endif %}'
+      state: present
+    register: etcd_installation_deb
```
alexbrand: typo

ansible/roles/packages-repo-custom/tasks/main.yaml
```diff
@@ -0,0 +1,69 @@
...
+      file: "/tmp/kismatic.key"
+      validate_certs: no
+    when: ansible_os_family == 'Debian'
+
+  - name: add Kismatic deb repository
+    apt_repository:
+      repo: 'deb {{ item.1 }} kismatic-xenial main'
+      validate_certs: no
+    when: ansible_os_family == 'Debian'
+    with_indexed_items: "package_repository_urls.split(',')"
```
alexbrand: Are we missing the use of the `index` anywhere? If not, seems like we can use `with_items` here?

ansible/roles/packages-transition-etcd/tasks/main.yaml
```diff
@@ -1,7 +1,23 @@
...
+    delay: 3
+    when: ansible_os_family == 'RedHat'
+
+  - name: install curl yum package
+    yum:
+      name: curl
+      state: present
+      disablerepo: '{% if package_repository_urls is defined and package_repository_urls != "" %}*{% endif %}'
+      enablerepo: '{% if package_repository_urls is defined and package_repository_urls != "" %}kismatic-*{% endif %}'
     register: etcd_installation_rpm
```
alexbrand: typo

ansible/roles/packages-transition-etcd/tasks/main.yaml
```diff
@@ -13,17 +29,21 @@
...
-  - name: install curl package
-    package: name=curl state=present
-    register: curl_installation_rpm
-    until: curl_installation_rpm|success
+  - name: install curl deb package
+    apt:
+      name: curl
+      allow_unauthenticated: '{% if package_repository_urls is defined and package_repository_urls != "" %}yes{% else %}no{% endif %}'
+      state: present
+    register: etcd_installation_deb
```
alexbrand: typo

ansible/roles/packages-transition-etcd/tasks/main.yaml
```diff
@@ -13,17 +29,21 @@
...
       state: present
+      allow_unauthenticated: '{% if package_repository_urls is defined and package_repository_urls != "" %}yes{% else %}no{% endif %}'
       default_release: kismatic-xenial
     register: etcd_installation_deb
     until: etcd_installation_deb|success
     retries: 3
     delay: 3
     when: ansible_os_family == 'Debian'
 
-  # required until https://github.com/ansible/ansible/issues/14804
```
alexbrand: Can we maintain this comment? Or maybe a new comment that says why we need curl?

ansible/upgrade-docker-registry.yaml
```diff
@@ -2,9 +2,7 @@
   - include: _packages-repo.yaml
     when: allow_package_installation|bool == true
 
-  # - include: _docker-registry-container.yaml play_name="Upgrade Internal Docker Registry"
-  #   when: setup_internal_docker_registry|bool == true
-  - include: _docker-registry-images.yaml  # only load images if offline and package upgraded or allow_package_installation == false
+  - include: _docker-registry.yaml upgrading=true  # only load images if offline and package upgraded or allow_package_installation == false
     when: >
```
alexbrand: Should any of the "when" conditions change here, now that the `_docker-registry` is doing multiple things?

alexbrand:
I think this looks good to me overall. Seems like all we need is a passing regression test. 

Automated testing for this feature will be tackled in the future. It was tested manually using https://github.com/alexbrand/kismatic-disconnected-install

PR #503, opened 2017-04-18
Support a signed docker registry

dkoshkin:


alexbrand:
How does this fail if the provided registry has a self-signed cert, or a signing cert that is not trusted by the machine?

alexbrand:
I am wondering if there's a preflight check we add for this

dkoshkin:
I was trying to think of one too but docker wont be installed at that point.

alexbrand:
Might be overkill, but maybe we can hit the registry over HTTPS at `/v2/`. I am not sure if the `v1` API is still in use though. In any case, if we get a 200 or 401 it should be OK.

alexbrand:
Decided to punt on the preflight check. We will add one if users run into issues.

PR #505, opened 2017-04-18
Don't include node's internal IP in the cert if it's unset

### LINKED ISSUE #498: generated certs have an empty DNS SAN

alexbrand:
The problem was that the internal IP was always being included in the node's certificates. In the case where the internal IP was not provided, we would add the empty string as a SAN in the cert.

Fixes #498

### DIFF
--- a/pkg/install/pki.go
+++ b/pkg/install/pki.go
@@ -202,7 +202,10 @@ func (lp *LocalPKI) GenerateNodeCertificate(plan *Plan, node Node, ca *tls.CA) e
 	if err != nil {
 		return err
 	}
-	nodeSANs := append(clusterSANs, node.Host, node.IP, node.InternalIP)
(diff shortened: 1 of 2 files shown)


dkoshkin:
LGTM!

PR #518, opened 2017-04-26
Add Helm proposal

dkoshkin:


proposals/helm.md
```diff
@@ -0,0 +1,22 @@
+# Helm
+
+Status: Proposal
+
+Helm is a tool for managing Kubernetes charts. Charts are packages of pre-configured Kubernetes resources.
+
+Helm is in the kubernetes Github repository, has a large community support and seems to be the de facto tool to deploy complex common applications in a repeated way.  
+
+Helm has two parts: a client (`helm`) and a server (`tiller`).
+Helm can be installed by running `helm init`, this deploys `tiller` in the cluster and sets up the local `~/.helm/` directory on the machine where the command was run on.
```
alexbrand: What is the `.helm` directory used for?

proposals/helm.md
```diff
@@ -0,0 +1,22 @@
...
+
+Although current k8s clusters built with KET _support_ Helm and can be configured to run `tiller` with `helm init` there are benefits of configuring Helm as part of the initial cluster installation:
+* Use helm to install monitoring and logging charts
+* Allow a user to have a cluster with working helm already installed
+* Deploy a predictable and tested version of helm on the cluster
+
+# Required Changes
+* Include the `helm` binary as part of the KET tar ball
+* Create a new _phase_ to deploy Helm
```
alexbrand: Not sure what you mean by a `phase`?

proposals/helm.md
```diff
@@ -0,0 +1,22 @@
...
+  * Use the included binary to run `helm init` (explore including it as a Go dependency instead)
```
alexbrand: I think running `helm init` should be fine.

proposals/helm.md
```diff
@@ -0,0 +1,22 @@
...
+* Include `tiller` docker images in the offline package
+* Plan file option to disable *NOTE* This will block installation of any charts that KET would configure, ie logging and monitoring and any others in the future
```
alexbrand: We need to define what this would look like in the plan file as part of this proposal. Might be worth discussing the "features" investigation I did a while back on a call.

proposals/helm.md
```diff
@@ -0,0 +1,22 @@
...
+Helm has two parts: a client (`helm`) and a server (`tiller`).
+Helm can be installed by running `helm init`, this deploys `tiller` in the cluster and sets up the local `~/.helm/` directory on the machine where the command was run on.
+
+Although current k8s clusters built with KET _support_ Helm and can be configured to run `tiller` with `helm init` there are benefits of configuring Helm as part of the initial cluster installation:
+* Use helm to install monitoring and logging charts
+* Allow a user to have a cluster with working helm already installed
+* Deploy a predictable and tested version of helm on the cluster
+
+# Required Changes
+* Include the `helm` binary as part of the KET tar ball
```
alexbrand: I am not against this, but want to think of alternatives:
- Use a container on one of the master nodes
- Any other?

proposals/helm.md
dkoshkin: Looks like to store the chart repositories and to cache some data.

```
➜  .helm ll
total 0
drwxr-xr-x  2 dkoshkin  APPRENDA\Domain Users    68B Apr 15 18:40 plugins
drwxr-xr-x  5 dkoshkin  APPRENDA\Domain Users   170B Oct 20  2016 repository
drwxr-xr-x  2 dkoshkin  APPRENDA\Domain Users    68B Apr 15 18:40 starters
➜  .helm cd repository
➜  repository ll
total 8
drwxr-xr-x  4 dkoshkin  APPRENDA\Domain Users   136B Oct 20  2016 cache
drwxr-xr-x  3 dkoshkin  APPRENDA\Domain Users   102B Oct 20  2016 local
-rw-r--r--  1 dkoshkin  APPRENDA\Domain Users   237B Apr 15 18:40 repositories.yaml
➜  repository cat repositories.yaml
apiVersion: v1
generated: 2017-04-15T18:40:42.880746976-04:00
repositories:
- caFile: ""
  cache: kubernetes-charts-index.yaml
  certFile: ""
  keyFile: ""
  name: kubernetes-charts
  url: http://storage.googleapis.com/kubernetes-charts
```

proposals/helm.md
dkoshkin: Similar to:
```
Installing Cluster==================================================================
Generating Kubeconfig File==========================================================
Installing Helm on the Cluster======================================================
This does bring up a question on the order, should the smoketest be the last thing? But if the cluster is broken we would be deploying tiller on a broken cluster.

File: proposals/helm.md (hunk 5 above)
Comment (dkoshkin) [inline proposals/helm.md]: I've been messing around how this would work, I'm not a fan of how the `.helm` directory gets overwritten, a user could already have helm setup and configured in a custom way.

Running it on one of the nodes might be a better solution.

Comment (dkoshkin) [inline proposals/helm.md]: There is no official docker image.

Comment (alexbrand) [inline proposals/helm.md]: Cool. I'm not terribly worried about the `.helm` directory, as long as it's clear that it will get overwritten.

File: proposals/helm.md (hunk 2 above)
Comment (alexbrand) [inline proposals/helm.md]: Got it. My initial feeling on this is that we would treat `tiller` similar to how we treat DNS, Dashboard, etc. We would add a new role and play in `kubernetes.yaml`. Do you see a reason not to?

File: proposals/helm.md (hunk 4 above)
Comment (alexbrand) [inline proposals/helm.md]: I thought about this a bit more. Some options that came to mind:

New `map[string]string` under cluster:
```
cluster:
    add_ons:
        networking: calico|weave
        package_manager: helm|none
        monitoring: prometheus|none (how do we represent prom + heapster?)
        logging: fluentd|none
```
The problem here is that some of these are really more than just one thing (e.g. monitoring: prometheus + heapster + data store)

New `map[string]bool` under cluster:
```
cluster:
    add_ons:
        calico: true|false
        weave: true|false
        helm: true|false
        prometheus: true|false
        heapster: true|false
        fluentd: true|false
```

Another where we abstract the implementation from the feature. This limits the ability to choose.
```
cluster:
    add_ons:
        package_manager: true|false
        monitoring: true|false
        logging: true|false
```

Or... Graduate some of these add-ons into top level configuration options:
```
cluster:
    monitoring:
        enabled: true|false
        ... any other required config here ...
    logging:
        enabled: true|false
        ... any other config here (such as specifying a sink for the logs) ...
```

Comment (ikester) [inline proposals/helm.md]: I'm inclined to go with the latter option since it provides the greatest flexibility to configure each "feature".

```
cluster:
  features:
    networking:
      enabled: true|false
      provider: calico|weave
      ...
    storage:
      enabled: true|false
      provider: gluster|ceph
      ...
    ingress:
      enabled: true|false
      provider: nginx|haproxy|traefik
      ...
    monitoring:
      enabled: true|false
      provider: prometheus
      options: ...
      ...
    logging:
      enabled: true|false
      provider: elk|xxx
      ...
```

Comment (alexbrand) [inline proposals/helm.md]: I like this. I wonder if "features" should live under cluster, or beside it.

Also, where do you see `helm` fitting in here? `package_manager:`?

Comment (ikester) [inline proposals/helm.md]: I'm inclined to go with the latter option since it provides the greatest flexibility to configure each "feature".

```
cluster:
  features:
    networking:
      enabled: true|false
      provider: calico|weave
      ...
    storage:
      enabled: true|false
      provider: gluster|ceph
      ...
    ingress:
      enabled: true|false
      provider: nginx|haproxy|traefik
      ...
    monitoring:
      enabled: true|false
      provider: prometheus
      options: ...
      ...
    logging:
      enabled: true|false
      provider: elk|xxx
      ...
```

alexbrand: I like this. I wonder if `features` should be under `cluster`, or beside it.

Also, where do you see `helm` fitting in there? `package_manager: `?

ikester: I would group helm with things like dns. More of a core add-on on which higher-level features depend. It's something we need to configure other things so it shouldn't be optional.

alexbrand:
How does this work when there is no internet connectivity? Do we have to stand up a Helm chart store somewhere?

dkoshkin:
With that we start going down a rabbit hole of do we also provide a mechanism to get the required docker images for those charts?

ikester:
You can configure helm with custom repos. See here: https://github.com/kubernetes/helm/blob/master/docs/using_helm.md#helm-repo-working-with-repositories

alexbrand:
Thanks for the link @ikester 

If we use helm as our way to deploy higher level "add-ons" such as logging/monitoring, it seems like we would need to ship the charts that we depend on and also setup a custom helm repo to support a disconnected install.

alexbrand:
Looks like helm is capable of installing charts from the local filesystem, and not necessarily a repo.

dkoshkin:
Having a `helm` package could work for this, this would include the helm binary + the charts as tars we want to include.

alexbrand:
Hmmm, let's think more about this. If we are thinking about moving away from packages, then we should probably not introduce a new one.

alexbrand:
I think we've reached consensus on this proposal. We can probably merge it and start working on implementation. I guess the last thing I am unsure about is upgrades.

PR #520, opened 2017-04-27
Add upgrade test from 1.3.1

alexbrand:
Adding a new test to cover upgrades from v1.3.1

### DIFF
--- a/integration/upgrade_test.go
+++ b/integration/upgrade_test.go
@@ -25,15 +25,31 @@ var _ = Describe("Upgrade", func() {
 			})
 		})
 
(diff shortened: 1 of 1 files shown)


dkoshkin:
LGTM

dkoshkin:
Can you take a look at https://github.com/apprenda/kismatic/blob/master/integration/upgrade_test.go#L33, this should be CentOS, we are using Ubuntu for the large test already.

alexbrand:
Good catch. Done.

dkoshkin:
Looks good.

PR #525, opened 2017-05-01
Rev ansible version. Clean up preflight tasks

### LINKED ISSUE #466: Upgrade fails at kismatic-inspector
Attempting to upgrade a cluster created with v.1.2.2 using v1.3.0:

`kismatic upgrade offline --dry-run`

Every master, etcd, worker node fails with :

`Error: error running inspector against remote node: server responded with non-successful status: "400 Bad Request"`

running:

`<path to>/kismatic-inspector client <etcd-node-ip>:8888 --upgrade --node-roles etcd`

produces the same error.

From the logs on the node:

`kismatic-inspector: 2017/04/03 17:36:19 error unmarshaling rules from JSON: rule with kind "freespace" is not supported`

Looks to be an incompatibility between versions of inspector, I expect this could be worked around with a custom rules file, but I cannot find an example of this.

Any hints?

alexbrand:
Fix #466 

Using the latest version of ansible fixes the issue we have seen in the
past where the inspector is left running on the node if any of the
checks fail.

TBD if there are any other changes required in the ansible plays. The long tests will tell us.


alexbrand:
There seems to be a regression in ansible-playbook, which I have documented here: https://github.com/ansible/ansible/issues/24674

For now, we'll have to force fact gathering somehow.

pkg/install/explain/updating.go
```diff
@@ -37,8 +38,14 @@ func (e *updatingExplainer) ExplainEvent(ansibleEvent ansible.Event) {
 
 	case *ansible.PlayStartEvent:
 		if e.currentPlayName != "" {
-			util.PrettyPrintOk(e.out.Bypass(), "%s", e.currentPlayName)
+			// If tasks ran, print OK. Otherwise, print SKIPPED
```
dkoshkin: Interesting I though we already print "SKIPPED" for tasks.

alexbrand: This will print skipped for play if no tasks ran inside the play.

dkoshkin: Yep I see that now, awesome.
With the new version of Ansible the plays are always printed even when all the tasks are skipped correct?

alexbrand: Unfortunately yes. Although I think it's a bug, and I have started a discussion with the ansible team at https://github.com/ansible/ansible/issues/16740#issuecomment-301922406

alexbrand:
I have opened a second ansible bug here: https://github.com/ansible/ansible/issues/24788

ansible/roles/calico/tasks/main.yaml
```diff
@@ -39,7 +39,7 @@
 
   - name: delete calico pod running on this node
     command: kubectl delete pod {{ calico_pod_name.stdout }} -n kube-system --now
-    when: calico_pod_name is defined and calico_pod_name.stdout != ""
```
dkoshkin: Is this only required with strings, we have a few occurrences of 
```
out.stdout|int == otherout.stdout|int
```

alexbrand: I actually don't remember why I changed this. It must have been because I got a "undefined variable" error when testing it out.

dkoshkin:
LGTM

Issue #509, opened 2017-04-21
Install Helm by default


swade1987:
It would be useful to have Helm pre-installed by default on all clusters.

alexbrand:
Thanks @swade1987. We will most likely address this in the short term.

PR #528, opened 2017-05-01
Install Helm

dkoshkin:
Fixes: #509 

Added new plan file options:
```
cluster...
features:
  package_manager:
    enabled: true
    provider: helm
```

Modified output:
```
Validating SSH connectivity to nodes                                            [OK]

Generating Kubeconfig File==========================================================
Generated kubeconfig file in the "generated" directory                          [OK]

Configuring Certificates============================================================
Found valid key and certificate for node "etcd001"                              [OK]
Found valid key and certificate for node "master001"                            [OK]
Found valid key and certificate for node "worker001"                            [OK]
Found certificate for docker registry                                           [OK]
Found key and certificate for service accounts                                  [OK]
Found key and certificate for user "admin"                                      [OK]
Cluster certificates can be found in the "generated" directory                  [OK]

Installing Cluster==================================================================
...

Installing Helm on the Cluster======================================================
Backed up "/Users/dkoshkin/.helm" directory                                     [OK]
Installed Tiller (the helm server side component)                               [OK]
Create Required RBAC Role for Helm
- Running task:
  master001                                                                     [OK]
- Running task: Create new serviceaccount for tiller
  master001                                                                     [OK]
- Running task: Attach cluster-admin role to the service account tiller
  master001                                                                     [OK]
- Running task: Patch the tiller spec to use the new serviceaccount
  master001                                                                     [OK]
Create Required RBAC Role for Helm Finished

Smoke Test Master Node
- Running task:
  master001                                                                     [OK]
- Running task: copy Kuberang to node
  master001                                                                     [OK]
- Running task: run smoke test checks using Kuberang
  master001                                                                     [OK]
Smoke Test Master Node Finished

The cluster was installed successfully!

- To use the generated kubeconfig file with kubectl:
    * use "kubectl --kubeconfig generated/kubeconfig"
    * or copy the config file "cp generated/kubeconfig ~/.kube/config"
- To view the Kubernetes dashboard: "./kismatic dashboard"
- To SSH into a cluster node: "./kismatic ssh etcd|master|worker|storage|$node.host"
```


eicnix:
Will this work with RBAC enabled cluster? AFAIK you don't create a serviceaccount and rolebinding for tiller. [Example](https://github.com/mgoodness/prometheus-operator/blob/d765e5651fa7acad51b236971d20cb32d9675a35/helm/prometheus-operator/README.md#rbac)

dkoshkin:
Good question @eicnix, I will take a look at that, thanks.

pkg/cli/apply.go
```diff
@@ -88,34 +89,62 @@ func (c *applyCmd) run() error {
...
+		// Run 'helm init'
+		if err := helm.Init(); err != nil {
+			util.PrettyPrintErr(c.out, "Installed Helm on the cluster")
+			return fmt.Errorf("error installing Helm on the cluster: %v", err)
+		}
+		util.PrettyPrintOk(c.out, "Installed Tiller (the helm server side component)")
+
+		// HelmRBAC will create a new role RBAC manually
+		// TODO remove when https://github.com/kubernetes/helm/issues/2224 gets fully fixed
+		if err := c.executor.RunPlay("_helm-rbac.yaml", plan); err != nil {
```
alexbrand: Would it be better to run this as part of the overall kubernetes play? We can have a new ansible var that will control it.

alexbrand: I guess the play waits until tiller is up and running. Hmmm.... This makes me wonder if calling `helm init` should be done from ansible as well.

pkg/install/execute.go
```diff
@@ -681,21 +680,17 @@ func (ae *ansibleExecutor) buildClusterCatalog(p *Plan) (*ansible.ClusterCatalog
...
+		cc.DockerCAPath = filepath.Join(tlsDir, "ca.pem")
+		cc.DockerRegistryPort = strconv.Itoa(p.DockerRegistry.Port)
+	}
+	if p.DockerRegistry.Address != "" { // Use external registry
 		cc.DockerCAPath = p.DockerRegistry.CAPath
 		cc.DockerRegistryAddress = p.DockerRegistry.Address
-		cc.DockerRegistryPort = strconv.Itoa(p.DockerRegistry.Port)
-	} else if p.DockerRegistry.SetupInternal {
+	} else if p.DockerRegistry.SetupInternal { // Use registry on master[0]
 		cc.EnableInternalDockerRegistry = true
```
alexbrand: It's somewhat unclear what the difference between `EnablePrivateDockerRegistry` and `EnableInternalDockerRegistry` is. Do we need both flags? If so, does `DeployInternalDockerRegistry` make more sense?

pkg/install/execute.go
```diff
@@ -681,21 +680,17 @@ func (ae *ansibleExecutor) buildClusterCatalog(p *Plan) (*ansible.ClusterCatalog
 		cc.LoadBalancedFQDN = p.Master.Nodes[0].InternalIP
 	}
 
-	if p.DockerRegistry.Address != "" {
-		cc.EnableInternalDockerRegistry = false
+	if p.ConfigureDockerRegistry() {
 		cc.EnablePrivateDockerRegistry = true
+		cc.DockerRegistryAddress = p.DockerRegistryAddress()
+		cc.DockerCAPath = filepath.Join(tlsDir, "ca.pem")
+		cc.DockerRegistryPort = strconv.Itoa(p.DockerRegistry.Port)
```
alexbrand: Do we respect this port when deploying the internal registry? Seems like this is changing behavior from the old default (`8443`)

pkg/install/plan_types.go
```diff
@@ -289,3 +298,12 @@ func hasIP(nodes *[]Node, ip string) bool {
 func (p Plan) ConfigureDockerRegistry() bool {
 	return p.DockerRegistry.Address != "" || p.DockerRegistry.SetupInternal
 }
+
+func (p Plan) DockerRegistryAddress() string {
+	address := p.DockerRegistry.Address
+	// If external is not set usem master[0]
+	if address == "" {
+		address = p.Master.Nodes[0].InternalIP
```
alexbrand: `InternalIP` can be empty

pkg/install/execute.go
```diff
@@ -751,6 +746,15 @@ func (ae *ansibleExecutor) generateTLSAssets(p *Plan) error {
 	if err != nil {
 		return fmt.Errorf("error generating certificates for the cluster: %v", err)
 	}
+
+	// Generate kubeconfig
+	err = GenerateKubeconfig(p, ae.options.GeneratedAssetsDirectory)
```
alexbrand: This seems unrelated to TLS assets. Was there a reason for moving this into here?

In general it seems like a good idea to move the kubeconfig generation so that it happens before the installation. I think we can achieve this by moving the call to this func higher up in `cli/apply.go`

pkg/install/helm.go
```diff
@@ -0,0 +1,79 @@
...
+	Binary          string
+	ClientDirectory string
+	TillerImage     string
+	ServiceAccount  string
+	Kubeconfig      string
+	Stdout          io.Writer
+	Stderr          io.Writer
+}
+
+func DefaultHelmClient() (*HelmConfig, error) {
```
alexbrand: Seems like this should return a `HelmClient`?

pkg/install/helm.go
```diff
@@ -0,0 +1,79 @@
...
+	home, err := homedir.Dir()
+	if err != nil {
+		return nil, fmt.Errorf("Could not determing $HOME directory: %v", err)
```
alexbrand: typo: determine

pkg/install/helm.go
```diff
@@ -0,0 +1,79 @@
...
+	_, err := os.Stat(h.ClientDirectory)
+	var backedup bool
+	// Directory exists
+	if err == nil {
+		if err = os.Rename(h.ClientDirectory, h.BackupDirectory()); err != nil {
+			return true, fmt.Errorf("Could not back up %q directory: %v", h.ClientDirectory, err)
+		}
+		return true, nil
+	} else if !os.IsNotExist(err) { // Directory does not exist but got some other error
+		return backedup, fmt.Errorf("Could not determing if $HOME/.helm directory exists: %v", err)
```
alexbrand: typo: determine

pkg/cli/apply.go
```diff
@@ -88,34 +89,62 @@ func (c *applyCmd) run() error {
...
+		util.PrintHeader(c.out, "Installing Helm on the Cluster", '=')
+		helm, err := install.DefaultHelmClient()
+		if err != nil {
+			return fmt.Errorf("error getting a valid Helm client: %v", err)
+		}
+		helm.Kubeconfig = filepath.Join(c.generatedAssetsDir, "kubeconfig")
+		// On a disconnected install set tiller image with the correct tag
+		// Prepend the custom registry address and port
+		if plan.ConfigureDockerRegistry() && plan.Cluster.DisconnectedInstallation {
+			// TODO how to cleanly get the tiller tag?
```
alexbrand: Is this still a TODO?

pkg/install/helm.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	// Directory does not already exist, nothing to do
+	return backedup, nil
+}
+
+// Init exectues 'helm init' using the local binary
+func (h *HelmConfig) Init() error {
```
alexbrand: This is minor, but what do you think of this method taking the tiller image and kubeconfig as params. That way we don't have to set them on the struct.

pkg/cli/upgrade.go
```diff
@@ -178,7 +178,7 @@ func doUpgrade(out io.Writer, opts *upgradeOpts) error {
 		fmt.Fprintln(out)
 	}
 
-	if plan.ConfigureDockerRegistry() && plan.Cluster.DisconnectedInstallation {
+	if plan.ConfgiureDockerWithPrivateRegistry() && plan.Cluster.DisconnectedInstallation {
```
alexbrand: Typo

pkg/install/plan_types.go
```diff
@@ -285,7 +295,29 @@ func hasIP(nodes *[]Node, ip string) bool {
...
+		if p.Master.Nodes[0].InternalIP != "" {
+			address = p.Master.Nodes[0].InternalIP
+		}
+	}
+	return address
+}
+
+func (p Plan) DockerRegistryPort() string {
+	port := 8443
+	// If external is not set usem master[0]
```
alexbrand: Comment seems irrelevant to code

pkg/cli/apply.go
```diff
@@ -88,34 +91,61 @@ func (c *applyCmd) run() error {
 	if err != nil {
 		return fmt.Errorf("error validating plan: %v", err)
 	}
-	plan, err := c.planner.Read()
+	plan, _ := c.planner.Read()
```
alexbrand: Chances of this failing are slim b/c we've read the plan already in validate, but I'd rather handle this even if it's redundant. If this code moves around, or we change how we do validation, we'll most likely forget about this.

pkg/util/file.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// BackupDirectory checks for existance of the $sourceDir and backs it up to backupDir
+func BackupDirectory(sourceDir string, backupDir string) (bool, error) {
+	_, err := os.Stat(sourceDir)
+	var backedup bool
+	// Directory exists
+	if err == nil {
+		if err = os.Rename(sourceDir, backupDir); err != nil {
+			return true, fmt.Errorf("Could not back up %q directory: %v", sourceDir, err)
```
alexbrand: Should this be `false` as nothing was backed up?

dkoshkin:
Need to investigate further https://github.com/kubernetes/charts/issues/1014

alexbrand:
This LGTM

PR #531, opened 2017-05-02
Fix certificate generation, and stop adding "Apprenda" subject

### LINKED ISSUE #530: Generated admin certificate is not part of the `system:masters` group
In order for the `admin` user to be able to bootstrap RBAC roles and bindings, the user needs to be bound to the `cluster-admin` role. As part of the RBAC work, this was done in the token based authentication, but we missed a small change in the cert generation to make it work in the cert-based auth.

alexbrand:
Fixes #530 

With this change, we stop adding the unnecessary subject fields on the generated certificates. This also fixes a bug where the organizations passed to the `generateCert` method were not being added to the certificate.

### DIFF
--- a/pkg/install/pki.go
+++ b/pkg/install/pki.go
@@ -11,11 +11,6 @@ import (
 )
 
 const (
-	certOrganization             = "Apprenda"
-	certOrgUnit                  = "Kismatic"
-	certCountry                  = "US"
-	certState                    = "NY"
-	certLocality                 = "Troy"
 	adminUser                    = "admin"
 	adminGroup                   = "system:masters"
 	dockerRegistryCertFilename   = "docker"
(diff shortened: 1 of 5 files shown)


pkg/install/pki.go
```diff
@@ -397,6 +397,12 @@ func generateCert(ca *tls.CA, commonName string, hostList []string, organization
 			},
 		},
 	}
+
+	for _, org := range organizations {
+		name := csr.Name{O: org}
+		req.Names = append(req.Names, name)
+	}
+
```
dkoshkin: Just adding this snippet did not work when trying to create [promethues](https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus), I also had to remove a line a few lines above:
```
O:  certOrganization,
```

alexbrand: Thanks @dkoshkin 
I'll have to look into why the group is not being picked up properly by the API server.

If only this was implemented: https://github.com/kubernetes/kubernetes/issues/30784

alexbrand: We figured this issue out and has been documented in #530

dkoshkin:
I see that you removed the fields from the CA also, for upgrades removing just the admin cert/key and letting the installer regenerate it should all still work without having to restart any of the services.

alexbrand:
Good call. That should work.

dkoshkin:
Cool yeah thats what I though, just wanted to think through it. LGTM

alexbrand:
Tests passed, but a pipeline stage further down the line failed. Squash + merging anyways.

PR #536, opened 2017-05-05
Added heapster to installations

dkoshkin:
This PR enables `heapster` installation, backed with influxdb, on all installations.
Heapster is a core component in a k8s cluster that enables numerous features, including:
* `kubectl top`
* graphs in the dashboard
* pod autoscaling

The pod specs are based on those in the heapster [repo](https://github.com/kubernetes/heapster/tree/master/deploy/kube-config/influxdb)

Plan file changes:
```
cluster:
...
add_ons:
  heapster:
    disabled: false
    options:
      heapster_replicas: 2
  package_manager:
    disabled: false
    provider: helm
```

There are a few options for the storage backing influxdb:
* default: `emptyDir{}`
* ~~`persistent_volume_enabled: true`: use NFS volume if provided, use Gluster volume, else fails validation~~
* `heapster.options.influxdb_pvc_name: "heapster-influxdb"`: user will to create the PVC/PV after install, the pod will wait until the PVC is created

**TODO**: 
[ ] add the needed images in the `kismatic-offline` package to support disconnected installations
[ ] ~~Use a persistent volume instead of using `emptyDir{}`~~


alexbrand:
Talked to @dkoshkin about this out of band, but including it here to not forget what we talked about.

My initial concern when looking at the plan file changes was that if the user is asking us to create volumes on their behalf, there is really no way for them to tell us _which_ NFS volume to use, or details about the desired Gluster volume.

We somewhat concluded that this use case would be covered by the user being able to provide a PV Claim name, and they can fulfill it however they want after the installation. 

But then, where does that leave the intermediate use case where kismatic is responsible for provisioning the volumes? If using gluster, what replication/distribution factors do we use? What storage clase do we set? If using NFS, do we choose one randomly from the list of shares? If using both NFS and gluster, and both prometheus and heapster are enabled, do we use NFS for prometheus and gluster for heapster, or the other way around?

integration/plan_patterns.go
```diff
@@ -61,6 +61,10 @@ features:
   package_manager:
     enabled: true
     provider: helm
+  heapster:
+    enabled: true
+    storage:
+      persistent_volume_enabled: ""
```
alexbrand: Looks like we need an update here?

pkg/install/plan.go
```diff
@@ -9,13 +9,45 @@ import (
...
 
 	yaml "gopkg.in/yaml.v2"
 )
 
+type stack struct {
+	lock sync.Mutex
+	s    []string
+}
+
+func NewStack() *stack {
```
alexbrand: Seems like we can keep this function unexported from the package

ansible/group_vars/all.yaml
```diff
@@ -140,7 +142,9 @@ kubernetes_certificates_service_account_key_path: "{{ kubernetes_install_dir }}/
 # etcd IPs
 etcd_networking_cluster_ip_list: "{% for host in groups['etcd'] %}https://{{ host }}:{{ etcd_networking_client_port }}{% if not loop.last %},{% endif %}{% endfor %}"
 etcd_k8s_cluster_ip_list: "{% for host in groups['etcd'] %}https://{{ host }}:{{ etcd_k8s_client_port }}{% if not loop.last %},{% endif %}{% endfor %}"
-
+#heapster
+heapster_replcias: 2
```
alexbrand: typo

ansible/group_vars/all.yaml
```diff
@@ -140,7 +142,9 @@ kubernetes_certificates_service_account_key_path: "{{ kubernetes_install_dir }}/
...
+monitoring_influxdb_replicas: 1
```
alexbrand: Can influxdb run with multiple instances?

ansible/roles/heapster/templates/influxdb.yaml
```diff
@@ -0,0 +1,44 @@
...
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: heapster-influxdb
+  namespace: kube-system
+  annotations:
+    kismatic/version: "{{ kismatic_short_version }}"
+spec:
+  replicas: 1
```
alexbrand: Should this be `{{ monitoring_influxdb_replicas }}`?

pkg/install/plan.go
```diff
@@ -194,33 +242,34 @@ func generateAlphaNumericPassword() (string, error) {
...
+	"docker_registry.address":                            "IP or hostname for your Docker registry. An internal registry will NOT be setup when this field is provided. Must be accessible from all the nodes in the cluster.",
+	"docker_registry.port":                               "Port for your Docker registry.",
+	"docker_registry.CA":                                 "Absolute path to the CA that was used when starting your Docker registry. The docker daemons on all nodes in the cluster will be configured with this CA.",
+	"nfs":                                                "A set of NFS volumes for use by on-cluster persistent workloads, managed by Kismatic.",
+	"nfs.nfs_host":                                       "The host name or ip address of an NFS server.",
+	"nfs.mount_path":                                     "The mount path of an NFS share. Must start with /",
+	"features.heapster.storage.persistent_volume_claim":  "By default an emptyDir{} will be used, to persist the data on pod restarts provide the name of the persistent volume claim to be created by you after the installation.",
```
alexbrand: Let's use this as a call to action from the user by inverting the sentence.

My first attempt:

```
Provide the name of the persistent volume claim that you will create after installation. 
If not specified, the data will be stored in ephemeral storage.
```

What do you think?

pkg/install/plan.go
```diff
@@ -63,12 +95,26 @@ func (fp *FilePlanner) Write(p *Plan) error {
...
 
+	s := NewStack()
 	scanner := bufio.NewScanner(bytes.NewReader(bytez))
+	prevIndent := -1
 	for scanner.Scan() {
 		text := scanner.Text()
 		matched := yamlKeyRE.FindStringSubmatch(text)
-
 		if matched != nil && len(matched) > 1 {
+			indent := strings.Count(matched[0], " ") / 2
```
dkoshkin: `go-yaml` uses 2 spaces by default (this is the only option atm)

ansible/group_vars/all.yaml
dkoshkin: Not anymore :( https://docs.influxdata.com/influxdb/v0.9/guides/clustering/

pkg/install/plan.go
alexbrand: Cool! Was mainly curious.

alexbrand:
What new docs or doc updates do we need for this?

dkoshkin:
I think having in the release notes the new options will be the most visible place to put document this.

pkg/ansible/clustercatalog.go
```diff
@@ -71,6 +71,8 @@ type ClusterCatalog struct {
 	LocalKubeconfigDirectory string `yaml:"local_kubeconfig_directory"`
 
 	EnableHelm bool `yaml:"enable_helm"`
+
+	AdditionalOptions map[string]string `yaml:",inline"`
```
alexbrand: Is this going to be a bag of random options? If so, I'd like to discuss this more, and understand the need for it.

Background: Before having the cluster catalog, we used to have a `map[string]string` for all the ansible vars.

alexbrand:
Did we reach a conclusion on making add-ons (all, or some) optional? My gut feeling is telling me that they should all be optional (even if we don't expose the flag).

@ikester @dkoshkin

ansible/upgrade-cluster-services.yaml
```diff
@@ -3,5 +3,6 @@
   - include: _kube-dns.yaml play_name="Upgrade Kubernetes DNS" upgrading=true
   - include: _kube-ingress.yaml play_name="Upgrade Kubernetes Ingress" upgrading=true
     when: configure_ingress|bool == true
-  - include: _heapster.yaml play_name="Upgrade Heapster Cluster Monitoring" upgrading=true
   - include: _kube-dashboard.yaml play_name="Upgrade Kubernetes Dashboard" upgrading=true
+  - include: _heapster.yaml play_name="Upgrade Heapster Cluster Monitoring" upgrading=true
+    when: heapster.enabled|bool == true
```
alexbrand: Do we need the `is defined` check here?

integration/install.go
```diff
@@ -65,11 +64,11 @@ func installKismatic(nodes provisionedNodes, installOpts installOptions, sshKey
 func buildPlan(nodes provisionedNodes, installOpts installOptions, sshKey string) PlanAWS {
 	sshUser := nodes.master[0].SSHUser
 	masterDNS := nodes.master[0].PublicIP
-	enableHelm := true
+	disableHelm := false
 	if nodes.dnsRecord != nil && nodes.dnsRecord.Name != "" {
 		masterDNS = nodes.dnsRecord.Name
 		// disable helm if using Route53
```
alexbrand: Why are we disabling helm if we are using Route53?

ansible/kubernetes.yaml
```diff
@@ -29,9 +29,11 @@
   - include: _calico-validate.yaml
   - include: _calico-network-policy.yaml
   - include: _kube-dns.yaml
+  - include: _heapster.yaml
+    when: heapster.enabled|bool == true
```
alexbrand: Do we need the `is defined` checks here?

ansible/upgrade-cluster-services.yaml
```diff
@@ -4,3 +4,5 @@
   - include: _kube-ingress.yaml play_name="Upgrade Kubernetes Ingress" upgrading=true
     when: configure_ingress|bool == true
   - include: _kube-dashboard.yaml play_name="Upgrade Kubernetes Dashboard" upgrading=true
+  - include: _heapster.yaml play_name="Upgrade Heapster Cluster Monitoring" upgrading=true
+    when: heapster.enabled|bool == true
```
alexbrand: Do we need the `is defined` check here?

alexbrand:
For the record: We had a discussion offline about making all add-ons optional and we agreed that indeed all should be optional, as we have already seen cases where we would need to disable some (e.g. contiv CNI plugin does not work with kubedns).

alexbrand:
@dkoshkin do you mind updating the PR description with the plan file changes?

ansible/upgrade-cluster-services.yaml
dkoshkin: I will add for consistency but _should_ not be required as it defaults to false. We can rework all of these in a single PR at some time in the future.

integration/install.go
dkoshkin: `helm init` relies on the `fqdn` address and the Route53 address is not reachable from outside the cluster on the test machine.

pkg/install/plan.go
```diff
@@ -178,6 +178,13 @@ func WritePlanTemplate(p *Plan, w PlanReadWriter) error {
 	// Set DockerRegistry defaults
 	p.DockerRegistry.Port = 8443
 
+	// Add-Ons
+	// Heapster
+	p.AddOns.HeapsterMonitoring.Options = make(map[string]string)
+	p.AddOns.HeapsterMonitoring.Options["influxdb_pvc_name"] = ""
```
alexbrand: I guess this is one downside of not having a proper struct for the options?

pkg/install/execute.go
```diff
@@ -756,12 +756,27 @@ func (ae *ansibleExecutor) buildClusterCatalog(p *Plan) (*ansible.ClusterCatalog
...
 	cc.EnableGluster = p.Storage.Nodes != nil && len(p.Storage.Nodes) > 0
-	// add_ons.package_manager
-	if p.AddOns.PackageManager.Enabled {
+
+	// add_ons
+	// heapster
+	if !p.AddOns.HeapsterMonitoring.Disabled {
+		cc.Heapster.Enabled = true
+		if p.AddOns.HeapsterMonitoring.Options != nil {
+			cc.Heapster.Options = make(map[string]string)
```
dkoshkin: @alexbrand would it make sense to just assign the existing map or iterate over it?

alexbrand:
Looks great! I think one thing we can do is expose the number of replicas in the options, so that an upgrade doesn't "reset" the instance count if the user has scaled it.

dkoshkin:
What would that option look like?
* Is there some validation for >0?
* If we dont validate and the section is missing from the yaml the default value will be 0, is that fine?

ansible/group_vars/all.yaml
```diff
@@ -239,6 +241,14 @@ helm_orig_img: gcr.io/kubernetes-helm/tiller:{{ helm_version }}
 helm_priv_img: "{{ docker_registry_address }}:{{ docker_registry_port }}/gcr.io/kubernetes-helm/tiller:{{ helm_version }}"
 helm_img: "{% if load_private_images|bool == true %}{{ helm_priv_img }}{% else %}{{ helm_orig_img }}{% endif %}"
 
+heapster_orig_img: gcr.io/google_containers/heapster-amd64:{{ heapster_version }}
+heapster_priv_img: "{{ docker_registry_address }}:{{ docker_registry_port }}/gcr.io/google_containers/heapster-amd64:{{ heapster_version }}"
```
alexbrand: Needs change to consistent name after #596

integration/heapster.go
```diff
@@ -0,0 +1,53 @@
...
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/apprenda/kismatic/pkg/retry"
+)
+
+func verifyHeapster(master NodeDeets, sshKey string) error {
+	// create volumes for alertmanager, prometheus-server and grafana
+	for n := 0; n < 1; n++ {
```
alexbrand: Do we need this loop?

pkg/install/plan_types.go
```diff
@@ -147,20 +147,35 @@ type SSHConnection struct {
...
+}
+
+type HeapsterMonitoring struct {
+	Disabled bool
+	Options  HeapsterOptions `yaml:"options"`
+}
+
+type HeapsterOptions struct {
+	HeapsterReplicas int    `yaml:"heapster_replicas"`
+	InfluxDBPVName   string `yaml:"influxdb_pvc_name"`
```
alexbrand: Typo: `InfluxDBPVName` Should be PVC instead of PV?

PR #537, opened 2017-05-06
Enable calico network policy on all installs

dkoshkin:
Currently installing the calico network policy controller during an install is optional. 
**NOTE** running the policy controller does not automatically enable any policy, by default all policies are disabled and it is only enable when users start creating network policies.

However, this complicates the install for a few reasons:
* Adding the policy controller to the cluster after the install is difficult
* Other network solutions, ie weave include the network controller as part of its deployment

This also simplifies the plan file options.

**TODO**: remove the `policy_enabled` option from the provisioner


swade1987:
1. Examples should be provided with the release notes after this change becomes effective. 
2. Calicoctl should be provided out the above within the Kismatic binary to allow for instant use.

alexbrand:
Hmm. I worry that we are going to start including a bunch of binaries in the tarball and it's just going to blow up in size.

swade1987:
@alexbrand agreed but if a given version of a cluster requires a specific calicoctl version to interact with then to me that's a dependency that is required.

dkoshkin:
One thing we can do is have 2 different builds:
1) With just the bits needed to install, ie `kismatic`
2) With all the tools, `helm`, `kubectl`, `calicoctl`, `etcdctl`, `provisioner`

alexbrand:
Let's merge this in. We can figure out packaging of calicoctl after.

Furthermore, there's also a `calicoctl` docker container that I think is becoming the defacto way of using it.

PR #538, opened 2017-05-08
Certificate generation revamp

### LINKED ISSUE #535: Use x509 client auth for kubelet, kube-proxy and control plane components
Currently, the kubelet, kube-proxy and the control plane components use token-based authentication with the admin's token. This means that this admin token is spread across all the nodes of the cluster.

Now that we are moving towards RBAC, we should instead use x509 certificate authentication with least-privileged accounts that are specific to each component.

The following are users/groups that should be used for each of the components:

Component | User | Group
------------|-------|-------
Scheduler| system:kube-scheduler| |
Controller Manager |system:kube-controller-manager| |
Kubelets | system:node:$hostname | system:nodes
Kube-proxy | system:kube-proxy | |

These certificates will be generated during our cert generation phase. Once generated, they will be copied over to the respective nodes. These certificates will be used in `kubeconfig` files that are specific to each component.

**Impact on upgrade**
It seems like as long as the existing CA is available, we can generate these new certificates, and replace them during the upgrade process. It's unclear to me at this time if we can ensure that the certs are re-generated, as it will be up to the user to not include them in the `generated` dir.

alexbrand:
Fixes #535 

This is a rather large PR, so I have split the work into meaningful commits that can be reviewed separately.

TODO:
- [x] Update kubedns to use a kubeconfig that doesn't have the admin password: See https://github.com/apprenda/kismatic/issues/535#issuecomment-300243621
- [x] Update calico to use a kubeconfig that doesn't have the admin password: 
Don't use a kubeconfig at all. Instead, use another configuration parameter in the CNI config to set the API server location.
- [x] Fix message that shows up when validation finds an existing cert
- [x] Fix so that duplicate cert specs are not requested twice
- [x] Perform cleanup during upgrade
- [x] Do we want to put all the certs in a directory? (e.g. `/etc/kubernetes/pki`)

Out of scope:
- CLI command for generating certs
- Using different security domains for etcd and kubernetes


pkg/install/pki.go
```diff
@@ -37,6 +46,142 @@ type LocalPKI struct {
...
+		})
+
+		m = append(m, certificateSpec{
+			description: "kube-proxy",
+			filename:    kubeProxyCertFilenamePrefix,
+			commonName:  kubeProxyUser,
+		})
+		// etcd client certificate
+		// all nodes need to be able to talk to etcd b/c of calico
+		m = append(m, certificateSpec{
```
dkoshkin: Do _etcd_ nodes also need the client cert for when we run `etcdctl`

pkg/install/pki.go
```diff
@@ -37,6 +46,142 @@ type LocalPKI struct {
...
+			filename:    "etcd-client",
+			commonName:  "etcd-client",
+		})
+	}
+
+	return m, nil
+}
+
+// returns a list of cert specs for the cluster described in the plan file
+func certManifestForCluster(plan Plan) ([]certificateSpec, error) {
```
dkoshkin: Do we need a service account cert?

ansible/group_vars/all.yaml
```diff
@@ -126,17 +126,27 @@ local_kubernetes_master_ip: https://127.0.0.1:{{ kubernetes_master_secure_port }
...
-kubernetes_certificates_key_file_name: kubenode-key.pem
-kubernetes_certificates_service_account_cert_file_name: service-account.pem
-kubernetes_certificates_service_account_key_file_name: service-account-key.pem
-kubernetes_certificates_ca_path: "{{ kubernetes_install_dir }}/{{ kubernetes_certificates_ca_file_name }}"
-kubernetes_certificates_cert_path: "{{ kubernetes_install_dir }}/{{ kubernetes_certificates_cert_file_name }}"
-kubernetes_certificates_key_path: "{{ kubernetes_install_dir }}/{{ kubernetes_certificates_key_file_name }}"
-kubernetes_certificates_service_account_cert_path: "{{ kubernetes_install_dir }}/{{ kubernetes_certificates_service_account_cert_file_name }}"
-kubernetes_certificates_service_account_key_path: "{{ kubernetes_install_dir }}/{{ kubernetes_certificates_service_account_key_file_name }}"
+# TODO: Do we want to change this?
+kubernetes_certificates_dir: "{{ kubernetes_install_dir }}"
```
dkoshkin: I think moving the certs to a `pki` subdirectory in `/etc/kubernetes` would be a good idea. We already have a few other dirs there.

Obviously would need testing and verification that the variable is respected in the components and pod specs for calico and DNS.(Not sure if it currently is)

ansible/roles/kubelet/templates/10-calico.conf
```diff
@@ -15,6 +15,6 @@
     },
 {% endif %}
     "kubernetes": {
-        "kubeconfig": "{{ kubernetes_services_kubeconfig_path }}"
+        "k8s_api_root": "{% if 'master' in group_names %}{{ local_kubernetes_master_ip }}{% else %}{{ kubernetes_master_ip }}{% endif %}"
```
dkoshkin: Good catch!

pkg/install/pki.go
```diff
@@ -54,6 +54,34 @@ type certificateSpec struct {
...
+	prelimEqual := s.description == other.description &&
+		s.filename == other.filename &&
+		s.commonName == other.commonName &&
+		len(s.subjectAlternateNames) == len(other.subjectAlternateNames) &&
+		len(s.organizations) == len(other.organizations)
+	if !prelimEqual {
+		return false
+	}
+	// Compare subject alt names. Sensitive to ordering.
+	for _, x := range s.subjectAlternateNames {
```
dkoshkin: Should we order this before comparing?

pkg/install/pki.go
alexbrand: Can the unsecured endpoint be used if you are on the etcd node?

pkg/install/pki.go
alexbrand: We have the key/cert pair that is used to sign/verify service accounts. It's only needed by master nodes.

ansible/group_vars/all.yaml
alexbrand: Yeah. I agree. The `/etc/kubernetes` directory is messy right now.

pkg/install/pki.go
alexbrand: Yes. Thinking about this more, order shouldn't impact equality.

pkg/install/pki.go
dkoshkin: Looks like we are only listen with TLS enabled when starting up the etcd services?
```
--listen-client-urls=https://${IP}:${CLIENT_PORT},https://127.0.0.1:${CLIENT_PORT}
```
Have you ran into any issues during upgrade tests?

alexbrand: Hmmm. I did not run into issues. I'll have to take a closer look at our `etcdctl` usage in the plays. Thanks!

dkoshkin:
Looks great, Im excited for this change. Lets merge it when we can.

alexbrand:
Yep! I want to look into the etcd thing you pointed out before merging. Also want to run tests after merging #546

alexbrand:
Manually verified that the current usage of etcdctl was not impacted by the cert work. This is because we are still able to use the certificate that etcd itself is using for HTTPS. The location of these certificates did not change, so existing usage of etcdctl continues to work.

I also ran the diagnostics command to verify that it still works, and successfully got etcdctl output:
```
# cat etcd_k8s_health
2017-05-17 15:38:50.513337 I | warning: ignoring ServerName for user-provided CA for backwards compatibility is deprecated
2017-05-17 15:38:50.513869 I | warning: ignoring ServerName for user-provided CA for backwards compatibility is deprecated
member bb02d6fe92882ad8 is healthy: got healthy result from https://192.168.42.2:2379
cluster is healthy
```

alexbrand:
@dkoshkin Can you take a look at the doc change? Want to make sure I didn't miss anything. After that I think we can merge this in.

docs/CERTIFICATES.md
```diff
@@ -0,0 +1,67 @@
...
+Kismatic will generate certificates for all the components in the cluster. Furthermore,
+it will also generate an admin certificate that can be used to access the API server
+as a cluster admin.
+
+### Generated Certificates
+
+| Certificate | Purpose | Filename |
+|---|---|---|
+| Self-Signed CA | Sign generated certificates |  ca.pem |
+| Etcd Server Cert | Serving API over HTTPS, performing peer-authentication | $nodeName-etcd.pem |
```
dkoshkin: Is there a reason why some cert names have `$nodeName` and some don't?

alexbrand: Certificates that have the node's hostname or IP in the CN or SAN list will need a different filename. These are the etcd server certificate, the API server certificate and kubelet client certificate. The rest of the client certs are the same regardless of the node that has them.

dkoshkin:
LGTM!

integration/upgrade_test.go
```diff
@@ -148,6 +148,14 @@ var _ = Describe("Upgrade", func() {
 						By("Removing old packages")
 						RemoveKismaticPackages()
 
+						// KET v1.3.4 introduced a new cert generation process that invalidated
+						// existing admin certificates. For this reason, we need to delete the
+						// certificate before running the upgrade.
+						err = os.Remove("generated/keys/admin.pem")
```
dkoshkin: Why do you need to delete it here if its being deleted during the upgrade with [this](https://github.com/apprenda/kismatic/pull/538/commits/b62cd7376d0c4a56e69c765cb1262bcaf59adc0a)

pkg/install/add_worker.go
```diff
@@ -24,7 +24,7 @@ func (ae *ansibleExecutor) AddWorker(originalPlan *Plan, newWorker Node) (*Plan,
 	if err != nil {
 		return nil, err
 	}
-	if err = ae.pki.GenerateNodeCertificate(originalPlan, newWorker, ca); err != nil {
+	if err = ae.pki.GenerateNodeCertificate(&updatedPlan, newWorker, ca); err != nil {
```
dkoshkin: Interesting how did this work before :)

integration/upgrade_test.go
alexbrand: This is a different test that does not use the `upgradeCluster` function.

pkg/install/add_worker.go
alexbrand: The GenerateNodeCertificate was purposefully built for add-worker, and did not expect to find the new worker in the plan file. Now that GenerateNodeCertificate is for any node, we need to figure out the node's roles from looking at a valid plan.

integration/upgrade_test.go
dkoshkin: My bad, I thought I was looking at the actual code that would delete the cert during upgrades, this is just integration tests, the user is responsible for deleting that node as we discussed previously.

dkoshkin:
LGTM!

integration/upgrade_test.go
alexbrand: Yeah... We can revisit that if it becomes painful, but I think we can move on with this.

alexbrand:
Failing tests was a flake/timeout.

PR #546, opened 2017-05-12
Use official calico CNI install script

alexbrand:
Historically, we have controlled the deployment of the calico policy controller using a flag in the plan file. Now that we deploy the controller out of the box, and we removed the flag, there is no need to use a custom calico CNI configuration file. For this reason, we can rely on the upstream installation script maintained by the calico team.

To avoid having to create a secret for a certificate that is already on the node, we configure the location of the etcd certificates manually (in the cni_network_config), instead of relying on the template support provided by the calico script.


ansible/roles/calico/templates/calico.yaml
```diff
@@ -11,6 +11,28 @@ data:
...
+        "etcd_key_file": "{{ kubernetes_certificates_key_path }}",
+        "etcd_cert_file": "{{ kubernetes_certificates_cert_path }}",
+        "etcd_ca_cert_file": "{{ kubernetes_certificates_ca_path }}",
+        "log_level": "info",
+        "ipam": {
+            "type": "calico-ipam"
+        },
+        "policy": {
+            "type": "k8s",
+            "k8s_api_root": "https://__KUBERNETES_SERVICE_HOST__:__KUBERNETES_SERVICE_PORT__",
```
dkoshkin: Can you remind me how the var substitution magic works?
Also looks like we didnt have `k8s_api_root` defined before, do you know why it this needed?

dkoshkin:
No tests :(

ansible/roles/calico/templates/calico.yaml
alexbrand: The install script maintained by the calico team runs `sed` against this config file with values it gets from environment variables. The script can be found at https://github.com/projectcalico/cni-plugin/blob/master/k8s-install/scripts/install-cni.sh and more docs on that are here: http://docs.projectcalico.org/v2.2/getting-started/kubernetes/installation/hosted/ 

The `k8s_api_root` is in the upstream config map. It is a bit strange to have to define the k8s endpoint in two places (here and in the kubeconfig). I am going to ask in the calico chat for more info on this.

alexbrand: Just confirmed with calico team that it is not necessary to specify the `k8s_api_root`.

Adding conversation for the record:
```
alexbrand [5:11 PM] 
Thanks @gunjan5 . Does the `policy` object need the `k8s_api_root` if the `kubeconfig` is being provided. It's not entirely clear from the docs
gunjan5 [5:27 PM] 
@alexbrand no. You can just use kubeconfig. So server address in `kubernetes` overrides the one in `policy` which overrides the one from `kubeconfig` (if any)
we,ve been using `k8s_api_root` in policy to template out and override the API server address in calico manifests
```

alexbrand: We actually do need to specify `k8s_api_root` because the generated kubeconfig file does not include the address to the server. Given that the `policy` one is preferred over the `kubernetes` one, I am going to set it again.

dkoshkin: Logs from the CNI container, 
- looks like `k8s_api_root` is not being set
- the cni generates a `kubeconfig` file, are we using that?

```
Wrote Calico CNI binaries to /host/opt/cni/bin/
CNI plugin version: v1.8.3
Wrote CNI config: {
    "name": "k8s-pod-network",
    "type": "calico",
    "etcd_endpoints": "https://ip-10-0-3-220:6666",
    "etcd_key_file": "/etc/kubernetes/kubenode-key.pem",
    "etcd_cert_file": "/etc/kubernetes/kubenode.pem",
    "etcd_ca_cert_file": "/etc/kubernetes/ca.pem",
    "log_level": "info",
    "ipam": {
        "type": "calico-ipam"
    },
    "policy": {
        "type": "k8s",
        "k8s_auth_token": "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9uYW1lc3BhY2UiOiJrdWJlLXN5c3RlbSIsImt1YmVybmV0ZXMuaW8vc2VydmljZWFjY291bnQvc2VjcmV0Lm5hbWUiOiJkZWZhdWx0LXRva2VuLWRoN3M2Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZXJ2aWNlLWFjY291bnQubmFtZSI6ImRlZmF1bHQiLCJrdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3NlcnZpY2UtYWNjb3VudC51aWQiOiJlZDQ0MDJkZi0zYTgxLTExZTctYWY0Ni0xMjAzYzMxMGRiMzgiLCJzdWIiOiJzeXN0ZW06c2VydmljZWFjY291bnQ6a3ViZS1zeXN0ZW06ZGVmYXVsdCJ9.mXKjsl7oMj4kdeczA8BjUj2SynWJ08hm5OZ0qBvdZnTOaTSxMABK-eCwnbpxT18UgkKHuHhQd-8kNCjBz-pycIjctvoMMyjdvh8YUhrTARz6G2V0wkhmAqatJCQd0Qq-np0Nv4CM_6za4sm-hPD836caPoo-Ab09b-IdOMlfWM0wqu9qoiXgeOBmN5Yzq0NFyDLtP7uN3lH1mAqSVzxF3lVbZ_oRLhByROZLK4stOyf5lyGN4kcr_ywDLFbDYQwfpnZweVzdu_SjRuAfiPZLcJV14kkw5ko1gxTfSJobtCTNfiUdpSsGdQhzpXKWumkTFqY-y5zNAbazljBN5k0Zzw"
    },
    "kubernetes": {
        "kubeconfig": "/etc/cni/net.d/calico-kubeconfig"
    }
}
```

alexbrand: We do use the generated kubeconfig, but it does not include the API server address. This results in the kubelet being unable to setup a pod network:
```
May 16 22:02:09 ip-10-0-3-220 kubelet[13887]: E0516 22:02:09.764755   13887 cni.go:257] 
Error adding network: invalid configuration: no server found for cluster "local"
```

`k8s_api_root` was not set in the test run because I removed it yesterday. Just added it again.

alexbrand:
Tests are looking good now. The one failure is a flake.

dkoshkin:
LGTM

PR #552, opened 2017-05-18
Cert CLI proposal

alexbrand:
Proposal for adding certificate management related commands to the kismatic binary.

### DIFF
--- a/proposals/certs-cli.md
+++ b/proposals/certs-cli.md
@@ -1,0 +1,89 @@
+# Certificates CLI
+
+Status: Proposal
+
+Certificates are a requirement for running secure Kubernetes clusters. In a cluster bootstrapped by kismatic, all component endpoints are secured using TLS. Furthermore, certificates are also used for user authentication when there is no other authentication mechanism configured.
+
+In order to facilitate the generation and management of certificates, a new command in the kismatic CLI is proposed.
+
+## Use Cases
+
+* As an operator, I want to grant cluster access to a new user by generating a client certificate
+* As an operator, I want to grant cluster access to an external system by generating a client certificate
+* As an operator, I want to check the validity of all certificates currently in use on the cluster
+* As an operator, I want to redeploy certificates that are nearing their expiration date
+* As an operator, I want to redeploy certificates with a new CA without having to reinstall my cluster
+
+# Design
+
+The following CLI commands are proposed for introduction to the `kismatic` binary:
+
+## Certificate Generation
+A generic, "swiss-army" style command is proposed for generating certificates:
+
+```
+kismatic certificates generate <name> [options]
+```
+
+Output: Generated key and certificate is placed in the generated directory. Key's filename is `<name>-key.pem`, and certificate's filename is `<name>.pem`.
+
+Pre-conditions:
+* CA is in the generated directory
+
+Options:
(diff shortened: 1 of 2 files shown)


jaycoon:
@alexbrand I really like this CLI proposal and agree with most of it. Some feedback:

**Do not set a default for `--expiry-days`, let the CA decide.** 
Validity period will be dependent on the CA being used, so KET should assume the CA is already in compliance for the environment being deployed to. If the CA default is higher than recommended, we can give a warning during validation, and at that time recommend the value be set to 365 days. Even then, the CA may not deploy a cert with a 365 validity period (e.g. the CA's own cert validity period expires in less than 365 days).

**Rename `--expiry-days` to `--validity-period`.**
For language consistency. 

**Add validation checks for validity periods.**
The current limit for a validity period is 39 months ([here's a post from the CA Council](https://casecurity.org/2015/02/19/ssl-certificate-validity-periods-limited-to-39-months-starting-in-april/)). The limit will be reduced to 825 days in 2018 ([CA/Browser ballot 193](https://cabforum.org/2017/03/17/ballot-193-825-day-certificate-lifetimes/)); KET can stay on top of these changes, and give recommendations during validation. 
Proposed validation checks:
1. Error: `--validity-period` value is greater than 39 months
2. Warning: the specified CA has a default validity period greater than 39 months (out of compliance)
3. Warning: `--validity-period` (or CA default) is greater than 825 days 


>  I think service accounts would have to be recreated if the CA changes, or if the service account signing cert changes.

Service accounts shouldn't need to be recreated, but they probably need to have their own cert remapped; these scenarios can be tested.

alexbrand:
Thanks @jaycoon 

I'll update the proposal and ping you for review.

alexbrand:
I will break this up into two proposals, as we will be pushing out the cert rotation feature to another release.

alexbrand:
Discussed with @jaycoon about using the CA's expiration policy by default, but looking at openssl and other tools, it doesn't seem like it is possible to do from just the CA certificate. The default expiration policy is configured in the tools, and not the CA cert. For this reason, I've kept the 365 days default.

@jaycoon, @dkoshkin Updated the proposal. PTAL.

jaycoon:
If the validity-period isn't set, it makes sense to set to 365 days. We can talk around this setting, and CA defaults per site, when doing installation and gathering requirements.

All other changes look good as well @alexbrand

dkoshkin:
How does the expiration work with the existing flag:
```
cluster:
  certificates:
    expiry: 17520h 
```

alexbrand:
Hmm. Good point @dkoshkin. I don't think this cert generation command will need access to the plan file. Do you think we should use the plan file as the default instead?

dkoshkin:
No I dont think we should use the plan file, just have the defaults we recommend to be the same in both the plan and the CLI.

alexbrand:
Got it. We started with 2 years. I wonder if we should drop that down to 1 year, or bring up the default in the new CLI command. Seems like we can drop the default to 1 year in the plan file, but we will need an answer for cert rotation sooner in time. What do you think?

dkoshkin:
I like 1 year, gives us a target date :)

alexbrand:
Works for me. I'll merge this then, and create an issue for the CLI piece.

PR #553, opened 2017-05-18
Remove unused helm variables from the plan

dkoshkin:
Currently we have a `provider` field for `Features.PackageManager`, there is only one provider, lets not pollute the plan file now and add it later when we actually need it.


alexbrand:
LGTM. We'll know more about this towards the end of the release when this, monitoring, logging, etc are integrated.

PR #554, opened 2017-05-19
Upgrade helm to v2.4.2

dkoshkin:
Helm released a new version and recommend upgrading from `v2.4.1`, we haven't released yet an upgrade path is not yet required.


alexbrand:
LGTM!

PR #558, opened 2017-05-22
Add centralized logging proposal

alexbrand:


alexbrand:
Some discussion out of band around log rotation: We came to the conclusion that we don't have to expose the log rotation settings given that the logs are being shipped out to a central store.

PR #574, opened 2017-06-12
Set version file when adding a worker

### LINKED ISSUE #570: Version is not being set with add-worker command

dkoshkin:
Fixes #570


alexbrand:
LGTM

PR #575, opened 2017-06-12
Fix typo in the end construct

prat0318:
`end` should be replaced by `endif`.


dkoshkin:
Thanks @prat0318 , did you notice a failure with this? Strange that Ansible doesn't throw a templating error here.

prat0318:
Yeah, saw a parse error during `upgrade`, which got fixed when i changed to `endif`.

dkoshkin:
@alexbrand do you see a problem merging this?

alexbrand:
LGTM! Although it would be nice to have a post-mortem on why this passed tests.

dkoshkin:
We have an integration test that only does this upgrade on Ubuntu and not on CentOS where this would impact it.

PR #576, opened 2017-06-12
Upgrade helm

dkoshkin:
`helm init --upgrade` is idempotent, the cluster will be initialized or `upgraded` only if tiller is already running.


alexbrand:
LGTM

PR #577, opened 2017-06-13
Transitive ansible dependency now requires make

alexbrand:
Ansible does not seem to lock down dependency versions. Specifically, a
new version of Paramiko was released, which itself brings in pynacl.
There is an open issue in the pynacl repo with no resolution yet, but
the recommended workaround is to install make.

See https://github.com/pyca/pynacl/issues/298 for more details.


dkoshkin:
Tested manually, LGTM

PR #579, opened 2017-06-13
Don't run a version check after each install test

dkoshkin:
We use internal IPs in our tests and those won't be reachable from the test machine.

This issue was introduced in https://github.com/apprenda/kismatic/pull/574


alexbrand:
LGTM

Issue #569, opened 2017-06-04
Unable to start calico-policy-controller pods after upgrading from v1.3.2 to v1.3.3


hreidar:
Hi, I'm getting an error after upgrading from kismatic1.3.2 to kismatic1.3.3. I'm running a three node dev cluster on Ubuntu 16.04. I have already upgraded once before on the same cluster, from 1.2.2 to 1.3.2 without any problems.

The error:
```
2017-06-02T15:52:17.317154788Z 2017-06-02 15:52:17,316 7 INFO Configuring /etc/hosts
2017-06-02T15:52:17.318115415Z 2017-06-02 15:52:17,317 7 INFO Appended 'kubernetes.default  -> 172.17.0.1' to /etc/hosts
2017-06-02T15:52:17.318142521Z 2017-06-02 15:52:17,317 7 INFO Beginning execution
2017-06-02T15:52:17.319288644Z 2017-06-02 15:52:17,318 7 INFO Leader election enabled? False
2017-06-02T15:52:17.348649154Z Traceback (most recent call last):
2017-06-02T15:52:17.348721247Z   File "<string>", line 562, in <module>
2017-06-02T15:52:17.349044126Z   File "<string>", line 159, in run
2017-06-02T15:52:17.349078054Z   File "site-packages/pycalico/datastore.py", line 901, in delete_policy_tier
2017-06-02T15:52:17.349098896Z   File "site-packages/etcd/client.py", line 610, in delete
2017-06-02T15:52:17.349115926Z   File "site-packages/etcd/client.py", line 820, in wrapper
2017-06-02T15:52:17.349131653Z   File "site-packages/etcd/client.py", line 890, in api_execute
2017-06-02T15:52:17.34914836Z   File "site-packages/urllib3/request.py", line 66, in request
2017-06-02T15:52:17.349775254Z   File "site-packages/urllib3/request.py", line 87, in request_encode_url
2017-06-02T15:52:17.349850291Z   File "site-packages/urllib3/poolmanager.py", line 244, in urlopen
2017-06-02T15:52:17.349861178Z   File "site-packages/urllib3/connectionpool.py", line 600, in urlopen
2017-06-02T15:52:17.349868216Z   File "site-packages/urllib3/connectionpool.py", line 345, in _make_request
2017-06-02T15:52:17.349874438Z   File "site-packages/urllib3/connectionpool.py", line 844, in _validate_conn
[...]

PR #582, opened 2017-06-13
Remove old policy-controller rc after upgrade

dkoshkin:
During upgrades we should remove old network-policy RC since it was replaced by a Deployment. We have a similar process for DNS and Dashboard.


alexbrand:
LGTM

dkoshkin:
According to their [docs](http://docs.projectcalico.org/v2.3/getting-started/kubernetes/upgrade) the old policy needs to be removed prior to starting a new one.

Needs more investigation.

dkoshkin:
Neither this nor https://github.com/apprenda/kismatic/pull/595 fix https://github.com/apprenda/kismatic/issues/569 but I wasnt really expecting it to.

I tested this manually to verify that it actually removes the old RC.

PR #591, opened 2017-06-15
Fix tasks that will fail with custom yum repos

### LINKED ISSUE #587: Disconnected installation fails on CentOS
Using kismatic 1.3.3, I'm trying to perform a disconnected installation on CentOS 7.2.1511, and the installation is failing.  It looks as though the checks in certain cases will only succeed on a Debian server.

```
- Running task: clean kismatic yum cache
  kismatic001                                                                   [OK]
- Running task: stat /etc/apt/sources.list
  kismatic001                                                                   [SKIPPED]
- Running task: rename /etc/apt/sources.list
  kismatic001 The conditional check 'apt_sources.stat.exists and ansible_os_family == 'Debian'' failed. The error was: error while evaluating conditional (apt_sources.stat.exists and ansible_os_family == 'Debian'): 'dict object' has no attribute 'stat'

The error appears to have been in '/home/kismaticuser/ansible/playbooks/roles/packages-repo-custom/tasks/main.yaml': line 34, column 5, but may

  - name: rename /etc/apt/sources.list
    ^ here
                                                                                [ERROR]
Configure Kismatic Package Repos Finished
```

dkoshkin:
Fixes #587 

Validate `ansible_os_family` before any OS specific checks.

### DIFF
--- a/ansible/roles/packages-repo-custom/tasks/main.yaml
+++ b/ansible/roles/packages-repo-custom/tasks/main.yaml
@@ -33,7 +33,7 @@
 
   - name: rename /etc/apt/sources.list
     command: mv /etc/apt/sources.list /etc/apt/sources.list.backup
-    when: apt_sources.stat.exists and ansible_os_family == 'Debian'
+    when: ansible_os_family == 'Debian' and apt_sources.stat.exists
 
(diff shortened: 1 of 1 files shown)


alexbrand:
LGTM

PR #592, opened 2017-06-15
Fallback to InternalIP/ExternalIP when using hosts file

### LINKED ISSUE #504: API server unable to contact new worker node
@xetys ran into an issue after adding a new worker. The worker was added using `apply`, but any attempts to get logs from containers running on that node would fail with:
```
Get https://minion03:10250/containerLogs/kloud/uaa-2391481347-9pn66/uaa-app?timestamps=true: 
dial tcp: lookup minion03 on 185.48.116.10:53: no such host
```

dkoshkin:
Fixes #504 

The reported issue occurs when `update_hosts_files` is enabled. When adding a new worker `/etc/hosts` gets modified but those changes aren't picked up the kube-api server until a restart.

Use new `--kubelet-preferred-address-types` flag in the API server. 

To minimize impact use the default `Hostname,InternalDNS,InternalIP,ExternalDNS,ExternalIP,LegacyHostIP` unless modifying hostnames.

This is the same config kubeadm uses.


alexbrand:
LGTM

Issue #594, opened 2017-06-16
CNI fails to create pods when a master node goes down


dkoshkin:
After https://github.com/apprenda/kismatic/pull/546 the script uses the internal `kubernetes` service endpoint instead of the FQDN specified in the plan file.

For a core component like CNI and DNS KET should probably not use the `kubernetes` endpoint as that does not respect the liveliness probes.

PR #595, opened 2017-06-19
Use kubeconfig with FQDN in calico conf file

dkoshkin:
Fixe #594 

Use the same cni conf file that was used prior to v1.3.3.
The conf file contains the `kubeconfig` that points to the FQDN instead of the local service IP.

### DIFF
--- a/ansible/roles/calico/templates/calico.yaml
+++ b/ansible/roles/calico/templates/calico.yaml
@@ -25,7 +25,7 @@ data:
         },
         "policy": {
             "type": "k8s",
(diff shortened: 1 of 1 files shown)


alexbrand:
One of the reasons for the changes in #546 was to use service accounts for calico, instead of using password authentication, which we'll eventually want to move away from. Reverting the change will brings us back to password auth. I think we can use the load balanced FQDN in the `k8s_api_root` field of the config?

dkoshkin:
Modified the PR with your suggestion and the HA test passed!

alexbrand:
LGTM

PR #597, opened 2017-06-19
Add flag to control private registry seeding

### LINKED ISSUE #589: Add ability to enable/disable seeding of images to private container image registry
Currently, the `disconnected_installation` flag controls both the source of the container images (private registry or docker hub) and whether KET should seed the private container registry. This means that it is currently not possible to perform a disconnected install using a registry that requires authentication, or a registry that is not writable by Kismatic nodes.

dkoshkin:
Fixes #589 

There is a new plan option `disallow_registry_seeding`:
```
cluster:
  name: kubernetes
  admin_password: MeHpoTijxinubci31      # This password is used to login to the Kubernetes Dashboard and can also be used for administration without a security certificate.
  allow_package_installation: true       # When false, installation will not occur if any node is missing the correct deb/rpm packages. When true, the installer will attempt to install missing packages for you.
  package_repository_urls: ""            # Comma-separated list of URLs of the repositories that should be used during installation. These repositories must contain the kismatic packages and all their transitive dependencies.
  disconnected_installation: false       # Set to true if you have local package and Docker repositories seeded with Kismatic binaries.
  disallow_registry_seeding: false       # Set to true if you have seeded your registry with the required images for the installation. Only has an effect when disconnected_installation is true.
```

I used *disallow* instead of disable to be consistent with *allow*_package_installation


alexbrand:
Great. I think the `disconnected_installation` inline documentation might need an update? I'm drawing a blank on what it should be though.

dkoshkin:
Good catch:
```
disconnected_installation: false       # Set to true if you have already installed the required packages on the nodes or provided a local URL in package_repository_urls containing those packages. With this flag set to true, the installer will also seed the needed container images into the available registry. 
```

alexbrand:
I think we can drop the second sentence? Does the `disconnected_install` still control the seeding of the registry?

dkoshkin:
You are right, I was thinking with the default case with `disallow_registry_seeding`, will remove that sentence. 
Also do you think `Only has an effect when disconnected_installation is true.` is needed?

alexbrand:
I think it looks good. Do we need an e2e test for this?

ansible/_docker-registry.yaml
```diff
@@ -7,8 +7,8 @@
 
     roles:
       - role: packages-offline
-        when: allow_package_installation|bool == true and disconnected_installation|bool == true  # install when allowed to install and disconnected
+        when: allow_package_installation|bool == true and disconnected_installation|bool == true and seed_registry|bool == true # install when allowed to install and disconnected
       - role: docker-registry-container
         when: deploy_internal_docker_registry|bool == true and (upgrading is not defined or upgrading|bool == false) # setup a registry running on master[0]
```
alexbrand: Is there a configuration where we would be unable to deploy the internal registry? I think no internet access + seed set to false would be one? I am not sure if we should try to prevent that, but still worth bringing it up.

pkg/install/validate.go
```diff
@@ -491,14 +491,20 @@ func validateAllowedAddress(address string) bool {
...
+	registry DockerRegistry
 }
 
 func (l disconnectedInstallation) validate() (bool, []error) {
 	v := newValidator()
-	if l.cluster.DisconnectedInstallation && !l.registryProvided {
-		v.addError(fmt.Errorf("A docker registry is required when disconnected_installation is true"))
+	if l.cluster.DisconnectedInstallation {
+		if !l.registry.ConfigureDockerWithPrivateRegistry() {
+			v.addError(fmt.Errorf("A docker registry is required when disconnected_installation is true"))
```
alexbrand: I think we should use "container image registry" instead of "docker registry" in this error message.

ansible/_docker-registry.yaml
alexbrand: Looks like you already covered this in validation 👍

dkoshkin:
As part of this PR we are renaming the `cluster.allow_package_installation` to `cluster.disable_package_installation`.

We want the zero values to reflect the plan file default values, hence the switch to **dis**able.

Also all plan values will use `enable` or `disable` instead of having a mix of `allow/disallow` and `enable/disable`

alexbrand:
LGTM!

PR #606, opened 2017-06-22
Rename add-ons disabled to disable for consistency

dkoshkin:
Renamed `disabled` to `disable`
```
add_ons:
  heapster:
    disable: false
    options:
      heapster_replicas: 2
  package_manager:
    disable: false
    provider: helm                       # Options: 'helm'
```


alexbrand:
LGTM!

PR #608, opened 2017-06-22
Include kubectl binary in KET binary

### LINKED ISSUE #516: Have kubectl inside the Kismatic binary ready to use
The user should be able to interact with the cluster straight away after it's been provisioned.

They shouldn't need to download and install Kubectl as a separate step.

We can also pin the Kubectl version to the version of Kubernetes being used for the release.

dkoshkin:
Fixes #516


alexbrand:
LGTM!

PR #610, opened 2017-06-23
Use service account for kubedns, delete unused kubeconfig file on upgrade

### LINKED ISSUE #535: Use x509 client auth for kubelet, kube-proxy and control plane components

dkoshkin:
Fixes #535 

Using `--kube-master-url` does not work because with this option the client doesn't mount the required CA.
Instead use the `default` service account and the CA thats is stored as a secret by setting `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` to point the LB fqdn.


alexbrand:
LGTM

PR #612, opened 2017-06-27
Add flags to the add volume command

### LINKED ISSUE #609: Configurable reclaimPolicy for adding a persistent-volume
The `pv.yaml` template has persistentVolumeReclaimPolicy hardcoded to 'Retain'. (see below)

https://github.com/apprenda/kismatic/blob/master/ansible/roles/persistent-volume/templates/pv.yaml#L12

I would recommend adding a flag to `kismatic storage add` to override this default.

Scenario: 
As every Persistent Volume (PV) I create is set to _Retain_ this means I have to manually wipe the contents of the PV every time I want to redeploy.

alexbrand:
Fixes #609 

Broken up into two commits to separate auto-generated documentation changes.


ansible/roles/persistent-volume/templates/pv.yaml
```diff
@@ -9,7 +9,7 @@ spec:
     storage: {{ volume_quota_gb }}Gi
   accessModes:
     - ReadWriteMany
```
dkoshkin: While we are in here do you think its worth externalizing this option also? A mismatch in PV/PVC here will cause it not to mount.

alexbrand: Yes, definitely. Good catch!

dkoshkin:
LGTM

alexbrand:
The storage test I wanted to see pass has passed in CI. Merging this in.

PR #614, opened 2017-06-27
Update upgrade tests to run from base cluster v1.3.3

alexbrand:
Following the upgrade guidelines discussed:
1. KET supports upgrades from the same minor version and any/-3(?) previous patch version
1. KET supports upgrades from the previous minor version last patch version

In preparation for the v1.4.0 release, update the upgrade tests to only test from the previous minor, last patch version (v1.3.3)


dkoshkin:
LGTM, lets merge it.

PR #617, opened 2017-06-29
Default heapster settings when not provided

### LINKED ISSUE #615: Plan validation should not fail validation with old plan
- [ ] Fix `heapster` section

dkoshkin:
Fixes #615 

When using an old plan file that is missing the `add_ons` section the validation should not fail and instead use default values.


alexbrand:
LGTM

PR #618, opened 2017-06-29
Backup admin cert, and regenerate kubeconfig during upgrade

alexbrand:
Pre-existing admin certificates (< v1.3.2) are not valid because they are missing the user required for RBAC. For this reason, we backup pre-existing admin certs if we detect that they were generated by us. 

If we regenerate the admin cert, we need to regenerate the kubeconfig file because it embeds the admin certificate. The previous kubeconfig file is backed up if it exists, and the user is informed about it.


dkoshkin:
LGTM

PR #620, opened 2017-06-29
Fix nil panics with heapster

dkoshkin:
This fixed failing tests after #617.

It wouldn't panic during normal execution now, only during tests but this is a good example of why using pointers here is not ideal.


alexbrand:
LGTM!

PR #622, opened 2017-07-03
Transition to CircleCI 

alexbrand:
To do:
- [x] Remove snap environment variables
- [x] Figure out way to trigger slow tests
- [x] Go back to using ansible version v2.3.0.0


dkoshkin:
Should we add the PR trigger.
```
deployment:
  fake_deploy_for_cci2:
    tag: /.*/
    commands:
      - echo "make tags run in 2.0"
```

Can we have make `local` that includes the `docker run ...` for `make test` and `make dist` since those are used to often.

dkoshkin:
LGTM!

PR #635, opened 2017-07-07
Fix issue where a long line logged by ansible would result in KET getting stuck

### LINKED ISSUE #633: Execution gets stuck gathering facts for hosts
With certain nodes, the execution gets stuck gathering facts when running `./kismatic install apply`:
```
root@apprk8s:/home/vagrant# ./kismatic install apply

Validating SSH connectivity to nodes                                            [OK]
Configure Cluster Prerequisites                                                 [OK]
Gather Node Facts
- Task: Gathering Facts
```

alexbrand:
The problem here was that ansible would, under certain (unknown) conditions, log a line that was larger than the Scanner's default buffer size. This resulted in the Scanner giving up, which in turn resulted in ansible getting blocked writing to the FIFO pipe.

Fixes #633

### DIFF
--- a/pkg/ansible/event_stream.go
+++ b/pkg/ansible/event_stream.go
@@ -1,26 +1,35 @@
 package ansible
 
 import (
-	"bufio"
 	"encoding/json"
 	"fmt"
 	"io"
+
+	"github.com/apprenda/kismatic/pkg/util"
 )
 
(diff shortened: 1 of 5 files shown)


dkoshkin:
Nice LGTM!

PR #638, opened 2017-07-07
Respect the certificate expiration policy set in plan file

alexbrand:
The certificate expiration option in the plan file is now respected when we generate certificates. 

I also propose we add a new field to the plan file to control the CA certificate expiration separate from the expiration of generated certificates. The reason for this is that CA certs are usually longer lived than regular certs.


ikester:
I agree with your assessment that CAs may be long-lived, at least as compared to component Certs. So it makes sense to have the flexibility of a different parameter for its expiry.

Makefile
```diff
@@ -96,7 +96,7 @@ test: vendor
 	    make bare-test
 
 bare-test: vendor
-	go test -v ./cmd/... ./pkg/... $(TEST_OPTS)
```
dkoshkin: why no `-v`?

alexbrand: I think I was trying to find a failing test, and having -v made it difficult to look at the output. But yeah, I think we should have it by default.

PR #641, opened 2017-07-11
Set hostname to lower when using nodenames and static pods

### LINKED ISSUE #634: Uppercase hostnames cause an issue
A user reported an issue when installing with a plan file that contained an uppercased hostnames.

When doing k8s pod validation, the installer is looking for `kube-apiserver-K8SMASTER`, but the name is actually `kube-apiserver-k8smaster`.

dkoshkin:
The kubelet uses the nodes hostname in a few places, however it sets any uppercase to lowercase. We need to follow the same strategy when interacting with the cluster. Specifically the nodename and the name of a static pod.

Also modified an existing integration test to cover this scenario.

Fixes #634

### DIFF
--- a/ansible/roles/calico/tasks/main.yaml
+++ b/ansible/roles/calico/tasks/main.yaml
@@ -24,13 +24,13 @@
       mode: "{{ network_environment_mode }}"
 
   - name: get the name of the calico pod running on this node
-    command: kubectl get pods -l=k8s-app=calico-node --template {%raw%}'{{range .items}}{{if eq .spec.nodeName{%endraw%} "{{ inventory_hostname }}"{%raw%}}}{{.metadata.name}}{{"\n"}}{{end}}{{end}}'{%endraw%} -n kube-system
+    command: kubectl get pods -l=k8s-app=calico-node --template {%raw%}'{{range .items}}{{if eq .spec.nodeName{%endraw%} "{{ inventory_hostname|lower }}"{%raw%}}}{{.metadata.name}}{{"\n"}}{{end}}{{end}}'{%endraw%} -n kube-system
     register: calico_pod_name
     when: upgrading is defined and upgrading|bool == true
 
     # label nodes that will have have calico isntalled
   - name: label nodes for calico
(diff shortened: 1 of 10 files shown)


ansible/roles/calico/tasks/main.yaml
```diff
@@ -30,7 +30,7 @@
 
     # label nodes that will have have calico isntalled
   - name: label nodes for calico
-    command: kubectl label nodes {% for host in play_hosts %}{{ host }} {% endfor %} kismatic/calico=true --overwrite --kubeconfig {{ kubernetes_kubeconfig_path }}
+    command: kubectl label nodes {% for host in play_hosts %}{{ host|lower }} {% endfor %} kismatic/calico=true --overwrite --kubeconfig {{ kubernetes_kubeconfig_path }}
```
alexbrand: Do we need to this as well on line 27 of this file?

ansible/roles/kube-proxy/tasks/main.yaml
```diff
@@ -15,4 +15,4 @@
       group: "{{ kubernetes_group }}"
       mode: "{{ kubernetes_service_mode }}"
 
-  - include: ../validate-pod/tasks/validate-pod.yaml item="kube-proxy-{{ inventory_hostname }}"
+  - include: ../validate-pod/tasks/validate-pod.yaml item="kube-proxy-{{ inventory_hostname|lower }}"
```
alexbrand: I wonder if it makes sense to call `lower` inside `validate-pod.yaml`, instead of sprinkling it all over. Not sure if it works.

dkoshkin: I thought of that but I wanted to keep that `validate-pod` generic and I dont know if any upper case is valid. Plus its already being `|lower` in a bunch of places so really not good way to abstract that.

ansible/roles/calico/tasks/main.yaml
dkoshkin: Modified the rest of the occurrences.

ansible/roles/kube-proxy/tasks/main.yaml
alexbrand: Yeah, I thought about it more and it actually doesn't make sense, as we also validate pods that don't have the hostname.

alexbrand:
LGTM!

PR #652, opened 2017-07-17
Add cni proposal

dkoshkin:


proposals/pluggable-cni.md
```diff
@@ -0,0 +1,44 @@
...
+    provider: helm                       
+```
+
+The `disabled:` field is valid, a user can choose not to install a CNI plugin.
+When `disabled:` is `true` the CNI flags in the Kubernetes components, however all pod validation for the other add-ons and the smoketest will be skipped.
+
+When `provider: custom` is set, the CNI flags will be set in the Kubernetes components, however all pod validation for the other add-ons and the smoketest will be skipped.
+
+## Other Considerations
+* A mechanism to install required [CNI binaries](https://github.com/containernetworking/cni) is required for certain network solutions. This can be a Docker container that contains the required binaries, contains a shared volume on the host machine and copies them on the machine. (Similar to Calico's approach)
```
alexbrand: I am not sure what this means for us, specifically. Do we need to do anything here?

proposals/pluggable-cni.md
```diff
@@ -0,0 +1,44 @@
...
+
+* Plan file will need to change, specifically:
```
alexbrand: Seems like this could go into the section that discusses the plan file changes. 

Can you be more specific on the required change? Is this about removing the networking block, or moving the `type: overlay`?

proposals/pluggable-cni.md
```diff
@@ -0,0 +1,44 @@
+# Pluggable CNI
+
+To support a larger set of environments and networks KET needs to support different CNI solutions.
+
+Currently only Calico is supported. All clusters setup with KET with <`v1.4.x` used Calico as its CNI add-on, users were not given an option.
+
+# Motivation 
+
+Initially support for an additional CNI plugins will be added to ket. [Weave Net](https://github.com/weaveworks/weave) to support additional functionality use cases as describe below.
```
alexbrand: s/Initially/Initial
s/ket/KET

I think we can remove the second sentence in this line? Or maybe mention that we will focus on Weave Net as the next CNI plugin that KET will support out of the box?

proposals/pluggable-cni.md
```diff
@@ -0,0 +1,44 @@
...
+
+Weave Net: 
+* adds support for the [Microsoft Azure](https://azure.microsoft.com/)
```
alexbrand: Multicast could be included in this list as well

proposals/pluggable-cni.md
```diff
@@ -0,0 +1,44 @@
...
+    options:
+      heapster_replicas: 2
+      influxdb_pvc_name: ""            
+  package_manager:
+    disable: false
+    provider: helm                       
+```
+
+The `disabled:` field is valid, a user can choose not to install a CNI plugin.
+When `disabled:` is `true` the CNI flags in the Kubernetes components, however all pod validation for the other add-ons and the smoketest will be skipped.
```
alexbrand: The CNI flags _are set_ in the Kubernetes components?
Can you elaborate no these flags? Are these binary flags that basically enable CNI, or do they take a path a location on disk? (I think it may be a combination?)

alexbrand:
Let's move forward with this. LGTM!

PR #654, opened 2017-07-17
Remove linkerd link from readme

alexbrand:


dkoshkin:
Should we remove `docs/LINKERD.md` also?

alexbrand:
We can tackle the linkerd documentation when we look at docs in general.

PR #656, opened 2017-07-18
Fix undefined variable in validate-pod.yaml

### LINKED ISSUE #655: Undefined ansible variable when validating pods
```
Start Kubernetes Proxy                                                          [ERROR]
- Task: get docker logs for kube-proxy-master1
  worker1: the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'container_ID' is undefined

The error appears to have been in '/Users/StevenWade/src/github.com/swade1987/pds-kubernetes-brown-bag/ansible/playbooks/roles/validate-pod/tasks/validate-pod.yaml': line 24, column 5, but may

    #failed_when: true to always print stderr and stdout
  - name: get docker logs for {{ item }}
    ^ here
We could be wrong, but this one looks like it might be an issue with
missing quotes.  Always quote template expression brackets when they
start a value. For instance:

    with_items:
      - {{ foo }}

Should be written as:

    with_items:
      - "{{ foo }}"
                                                                                [ERROR]
```

alexbrand:
Fixes #655

### DIFF
--- a/ansible/roles/validate-pod/tasks/validate-pod.yaml
+++ b/ansible/roles/validate-pod/tasks/validate-pod.yaml
@@ -22,7 +22,7 @@
 
     #failed_when: true to always print stderr and stdout
   - name: get docker logs for {{ item }}
-    command: docker logs {{ container_ID.stdout }}
+    command: docker logs {{ containerID.stdout }}
     register: docker_logs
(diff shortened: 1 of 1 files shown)


dkoshkin:
LGTM

PR #659, opened 2017-07-19
Add GOOS and versions to kubectl and helm makefile targets

alexbrand:
Before this change, if the vendor-helm/out and vendor-kubectl/out
directories existed, the make target would not run. We are making this
change because we ran into an issue with our CI build where the resulting tarballs contained the linux
kubectl and helm binaries in the GOOS=darwin package.

The new make targets have the version and GOOS to ensure that the they
run if the files are missing.


dkoshkin:
LGTM

PR #667, opened 2017-07-19
Addon CNI

### LINKED ISSUE #593: Treat networking as an add-on and support multiple CNI providers
Support for WeaveNet and Contiv in addition to Calico is desired.

dkoshkin:
Fixes #593 

Make Calico as a CNI add-on
```
add_ons:
  cni:
    disable: false
    provider: calico
    options:
      calico:
        mode: overlay
```
When `disable` is `true` the installer will skip all pod validations and the smoketest.

To retain backwards compatibility if  `add_ons.cni` is *nil* KET will default to Calico and the `cluster.networking.type` option will be used.

### DIFF
--- a/ansible/_helm.yaml
+++ b/ansible/_helm.yaml
@@ -24,14 +24,17 @@
         local_action: command ../../helm repo add kismatic https://apprenda.github.io/kismatic-charts/
         become: no
         when: disconnected_installation|bool != true
-      - name: wait until the tiller pod is ready
-        command: kubectl get deployment tiller-deploy -n kube-system -o jsonpath='{.status.availableReplicas}'
-        register: readyReplicas
-        until: readyReplicas.stdout|int == 1
-        retries: 24
-        delay: 10
-        failed_when: false # We don't want this task to actually fail (We catch the failure with a custom msg in the next task)
-      - name: fail if the tiller pod is not ready
-        fail:
-          msg: "Timed out waiting for the tiller pod to be in the ready state."
-        when: readyReplicas.stdout|int != 1
+
+      - block:
+        - name: wait until the tiller pod is ready
+          command: kubectl get deployment tiller-deploy -n kube-system -o jsonpath='{.status.availableReplicas}'
+          register: readyReplicas
+          until: readyReplicas.stdout|int == 1
+          retries: 24
+          delay: 10
+          failed_when: false # We don't want this task to actually fail (We catch the failure with a custom msg in the next task)
+        - name: fail if the tiller pod is not ready
+          fail:
+            msg: "Timed out waiting for the tiller pod to be in the ready state."
+          when: readyReplicas.stdout|int != 1
+        when: run_validation|bool == true 
--- a/ansible/_helm.yaml
+++ b/ansible/_helm.yaml
@@ -37,4 +37,4 @@
           fail:
             msg: "Timed out waiting for the tiller pod to be in the ready state."
           when: readyReplicas.stdout|int != 1
-        when: run_validation|bool == true 
+        when: run_pod_validation|bool == true 
--- a/integration/install_test.go
+++ b/integration/install_test.go
@@ -90,6 +90,18 @@ var _ = Describe("kismatic", func() {
 			})
 		})
 
+		Context("when deploying a cluster with all node roles and disabled CNI", func() {
+			installOpts := installOptions{
+				disableCNI: true,
+			}
+			ItOnAWS("should install successfully [slow]", func(aws infrastructureProvisioner) {
+				WithInfrastructure(NodeCount{1, 1, 1, 1, 1}, Ubuntu1604LTS, aws, func(nodes provisionedNodes, sshKey string) {
+					err := installKismatic(nodes, installOpts, sshKey)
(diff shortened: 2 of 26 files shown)


ansible/_helm.yaml
```diff
@@ -25,14 +25,17 @@
...
+          register: readyReplicas
+          until: readyReplicas.stdout|int == 1
+          retries: 24
+          delay: 10
+          failed_when: false # We don't want this task to actually fail (We catch the failure with a custom msg in the next task)
+        - name: fail if the tiller pod is not ready
+          fail:
+            msg: "Timed out waiting for the tiller pod to be in the ready state."
+          when: readyReplicas.stdout|int != 1
+        when: run_validation|bool == true
```
alexbrand: Can we include the word `pods` or in there.. something like` run_pod_validation` or `validate_pods`? Just to avoid confusion with preflight validation.

integration/install_test.go
```diff
@@ -90,6 +90,18 @@ var _ = Describe("kismatic", func() {
 			})
 		})
 
+		Context("when deploying a cluster with all node roles and disabled CNI", func() {
+			installOpts := installOptions{
+				disableCNI: true,
+			}
+			ItOnAWS("should install successfully [slow]", func(aws infrastructureProvisioner) {
```
alexbrand: What are we testing here in the end? The cluster won't be functional, right? 

I guess the install should succeed because we won't attempt validation / smoke testing?

pkg/install/plan_types.go
```diff
@@ -356,3 +378,8 @@ func (p Plan) DockerRegistryPort() string {
 	}
 	return strconv.Itoa(port)
 }
+
+// CanValidatePods returns true if pod validation/smoketest should run
+func (p Plan) CanValidatePods() bool {
```
alexbrand: Would it make more sense to name this something like `NetworkConfigured` or `NetworkDeployed`? In this way, whoever is the client of this method can decide whether to run validation or not.

pkg/install/plan_types.go
```diff
@@ -356,3 +378,8 @@ func (p Plan) DockerRegistryPort() string {
...
+	return p.AddOns.CNI == nil || !p.AddOns.CNI.Disable
```
alexbrand: Is `p.AddOns.CNI = nil` a valid configuration? Seems like we could remove it?

pkg/install/validate.go
```diff
@@ -224,11 +218,28 @@ func (s *SSHConfig) validate() (bool, []error) {
...
 	v := newValidator()
+	v.validate(f.CNI)
 	v.validate(f.HeapsterMonitoring)
 	v.validate(&f.PackageManager)
 	return v.valid()
 }
 
+func (n *CNI) validate() (bool, []error) {
+	v := newValidator()
+	if n != nil && !n.Disable {
```
alexbrand: I think at this point n should never be nil, given that we remain backwards compatible.

pkg/install/validate.go
```diff
@@ -224,11 +218,28 @@ func (s *SSHConfig) validate() (bool, []error) {
...
+		if !util.Contains(n.Provider, CNIProviders()) {
+			v.addError(fmt.Errorf("CNI %q is not a valid option %v", n.Provider, CNIProviders()))
```
alexbrand: What do you think about this wording? 
`%q is not a valid CNI provider. Valid providers are %v.`

pkg/install/validate.go
```diff
@@ -224,11 +218,28 @@ func (s *SSHConfig) validate() (bool, []error) {
...
+		}
+		if n.Options.Calico.Mode == "" {
+			v.addError(errors.New("Calico mode cannot be empty"))
+		}
+		if !util.Contains(n.Options.Calico.Mode, CalicoMode()) {
+			v.addError(fmt.Errorf("Calico mode %q is not a valid option %v", n.Options.Calico.Mode, CalicoMode()))
```
alexbrand: Same suggestion as above.

integration/install_test.go
dkoshkin: Yep thats exactly what it tests, I dont want us to add something in the future and forget to turn off validation.

pkg/install/plan_types.go
dkoshkin: This fails in unit tests, will look how to fix the test instead.

pkg/install/validate.go
dkoshkin: This fails in unit tests, will look how to fix the test instead.

pkg/install/validate.go
dkoshkin: I like it

integration/install_test.go
alexbrand: Makes sense. Do we need the 5 machines?

alexbrand:
Open review items are nitpicks. In the interest of unblocking other CNI-related work, I think we can merge this. @dkoshkin wdyt?

dkoshkin:
Going to merge. For the integration test we currently don't have a good way to use a 1 node cluster with custom options, and the other comment I would like to keep the `nil` check.

PR #668, opened 2017-07-19
Ansible become

### LINKED ISSUE #661: Inconsistent user elevation  in Ansible
Currently we use both `become` and `become_method` in different Ansible roles, moving to `become` will allow a user to overwrite the `become_method` in the Ansible config file.

Let's discuss if we also want to expose the `become_method` as an option in the plan file.

dkoshkin:
Fixes #661 

* `remote_user` is default to `root` removing that flag allows the user to overwrite a global option ins `ansible.cfg`
* use `become` instead of `become_method` to allow the user to overwrite a global option ins `ansible.cfg`

### DIFF
--- a/ansible/_addon-logging.yaml
+++ b/ansible/_addon-logging.yaml
@@ -3,7 +3,7 @@
     any_errors_fatal: true
     name: "Add-On Logging"
     remote_user: root
-    become_method: sudo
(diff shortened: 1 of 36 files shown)


alexbrand:
LGTM

PR #669, opened 2017-07-20
CNI Weave

dkoshkin:
Add new CNI add-on option `weave`

TODO:
- [x] Figure out how to get the CNI binaries on the nodes
- [x] Fix NetworkPolicy not working. Traffic is being blocked even after creating a resource to allow. Test can be seen failing in the `when deploying a skunkworks cluster and Weave` integration test. - https://github.com/weaveworks/weave/issues/3072

### DIFF
--- a/ansible/_cni.yaml
+++ b/ansible/_cni.yaml
@@ -1,0 +1,18 @@
+---
+  - hosts: master:worker:ingress:storage
+    any_errors_fatal: true
+    name: "{{ play_name | default('Configure Kubernetes CNI') }}"
+    serial: "{{ serial_count | default('100%') }}"
+    become: yes
+    vars_files:
+      - group_vars/all.yaml
+
+    tasks:
+      - name: create /opt/cni/bin directory
+        file:
+          path: "/opt/cni/bin"
+          state: directory
+      - name: download cni-bin image
+        command: docker pull {{ cni_bin_img }}
+      - name: install cni binaries
+        command: docker run -e OVERWRITE=true -v /opt/cni/bin:/host/opt/cni/bin {{ cni_bin_img }}
--- a/ansible/roles/weave/templates/weave.yaml
+++ b/ansible/roles/weave/templates/weave.yaml
@@ -1,0 +1,153 @@
+apiVersion: v1
+kind: List
+items:
+  - apiVersion: v1
+    kind: ServiceAccount
+    metadata:
+      name: weave-net
+      labels:
+        tier: control-plane
+        name: weave-net
+      namespace: kube-system
+  - apiVersion: rbac.authorization.k8s.io/v1beta1
+    kind: ClusterRole
+    metadata:
+      name: weave-net
+      labels:
+        tier: control-plane
+        name: weave-net
+    rules:
+      - apiGroups:
+          - ''
+        resources:
+          - pods
+          - namespaces
+          - nodes
+        verbs:
+          - get
+          - list
+          - watch
+      - apiGroups:
+          - extensions
+        resources:
+          - networkpolicies
+        verbs:
+          - get
+          - list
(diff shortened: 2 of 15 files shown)


ansible/roles/weave/templates/weave.yaml
```diff
@@ -0,0 +1,153 @@
...
+              hostPath:
+                path: /etc
+            - name: dbus
+              hostPath:
+                path: /var/lib/dbus
+            - name: lib-modules
+              hostPath:
+                path: /lib/modules
+      updateStrategy:
+        type: RollingUpdate
```
alexbrand: I think we need to use the original (pre 1.7 or pre 1.6) update strategy here to stay in line with our upgrades.

integration/install_test.go
```diff
@@ -241,6 +241,59 @@ var _ = Describe("kismatic", func() {
...
+					// Use master[0] public IP
+					sub.It("should have an accessible dashboard", func() error {
+						return canAccessDashboard(fmt.Sprintf("https://admin:abbazabba@%s:6443/ui", nodes.master[0].PublicIP))
+					})
+
+					sub.It("should respect network policies", func() error {
+						return verifyNetworkPolicy(nodes.master[0], sshKey)
+					})
+
+					// sub.It("should allow creating RBAC policy", func() error {
```
alexbrand: I think we can get rid of this test? Our clusters already depend on working RBAC.

alexbrand:
Looks like we are missing upgrade changes here?

ansible/upgrade-nodes.yaml
```diff
@@ -32,11 +32,17 @@
   - include: _kube-controller-manager.yaml play_name="Upgrade Kubernetes Controller Manager" upgrading=true
   - include: _validate-control-plane-node.yaml serial_count="1" upgrading=true
   - include: _kube-proxy.yaml play_name="Upgrade Kubernetes Proxy" upgrading=true
-  - include: _calico.yaml play_name="Upgrade Cluster Network" upgrading=true
+  - include: _cni.yaml upgrading=true
```
alexbrand: Do we need to set the play name here?

ansible/_cni.yaml
```diff
@@ -0,0 +1,20 @@
...
+    become_method: sudo
+    vars_files:
+      - group_vars/all.yaml
+
+    tasks:
+      - name: create /opt/cni/bin directory
+        file:
+          path: "/opt/cni/bin"
+          state: directory
+      # TODO figure out how to do CNI
```
dkoshkin: Remove this was figured out.

pkg/install/validate_test.go
```diff
@@ -996,16 +996,16 @@ func TestCNIAddOn(t *testing.T) {
...
-		// 				Mode: "overlay"},
-		// 		},
-		// 	},
-		// 	valid: true,
-		// },
+		{
+			n: CNI{
+				Provider: "weave",
+				Options: CNIOptions{
+					Calico: CalicoOptions{
```
alexbrand: If this is the only test for the weave provider, I would remove the calico options.

This also makes me wonder if we should consider this a valid CNI configuration, as the user is setting options for a different provider.

dkoshkin: The downside of validating something like that is that a user would need to remove options that are already set by default.

alexbrand: Are we setting calico options when the provider is not calico?

dkoshkin: Do you mean in the plan file or when passing to Ansible?
In the plan file its always set, since it defaults to `calico`. All options for all cni plugins (currently only calico mode) are also always set when passing to Ansible.

alexbrand:
LGTM!

Issue #650, opened 2017-07-17
Heapster: ServiceType option


dkoshkin:
Expose an option to change from the default `ClusterIP` ServiceType.

Issue #649, opened 2017-07-17
Heapster: Sink option


dkoshkin:
Expose an option to use a different sink instead of the on cluster InfluxDB

PR #676, opened 2017-07-24
Add serviceType and sink options to heapster

### LINKED ISSUE #648: Additions to Heapster installation
WIth Kismatic 1.4.0, Heapster is now installable, which is much appreciated.  However, it would be nice to have two additions to the current installation:

1. Add an option to also install Grafana, so that the metrics become even more readily available. 
2. Add an option to specify a different sink than the deployed InfluxDB (which would also cause there not to be a local InfluxDB instance) so that folks who have an external InfluxDB installation can collect there data there.

dkoshkin:
Fixes #648, #649, #650.

Add additional options in heapster add-on
```
add_ons:
  dns:
    disable: false
  heapster:
    disable: false
    options:
      heapster_replicas: 2
      heapster_service_type: ClusterIP   # Specify kubernetes ServiceType; default 'ClusterIP'
      heapster_sink: influxdb:http://heapster-influxdb.kube-system.svc:8086 # Specify the sink to store heapster data; default to a pod running on cluster.
```

Added `heapster_service_type` and `heapster_sink`. Both of these are optional and will have defaults if empty.

Another thing to consider is should the options have structs, ie.
```
add_ons:
  dns:
    disable: false
  heapster:
    disable: false
    options:
      heapster:
        replicas: 2
        service_type: ClusterIP
        sink: influxdb:http://heapster-influxdb.kube-system.svc:8086
      influxdb:
        pvc_name:
```

TODO:
- [ ]  Modify provisioner to include these options

### DIFF
--- a/pkg/install/plan.go
+++ b/pkg/install/plan.go
@@ -117,8 +117,24 @@ func setDefaults(p *Plan) {
 	}
 	if p.AddOns.HeapsterMonitoring == nil {
 		p.AddOns.HeapsterMonitoring = &HeapsterMonitoring{}
-		p.AddOns.HeapsterMonitoring.Options.HeapsterReplicas = 2
 	}
+	if p.AddOns.HeapsterMonitoring.Options.Heapster.Replicas == 0 {
+		p.AddOns.HeapsterMonitoring.Options.Heapster.Replicas = 2
+	}
+	// read field from KET < v1.5.0
+	if p.AddOns.HeapsterMonitoring.Options.HeapsterReplicas != 0 {
+		p.AddOns.HeapsterMonitoring.Options.Heapster.Replicas = p.AddOns.HeapsterMonitoring.Options.HeapsterReplicas
+	}
+	if p.AddOns.HeapsterMonitoring.Options.Heapster.Sink == "" {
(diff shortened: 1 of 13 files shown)


alexbrand:
I think having structs for options makes sense in the case where we have multiple providers.

pkg/install/plan.go
```diff
@@ -203,6 +210,8 @@ func WritePlanTemplate(p *Plan, w PlanReadWriter) error {
 	// Heapster
 	p.AddOns.HeapsterMonitoring = &HeapsterMonitoring{}
 	p.AddOns.HeapsterMonitoring.Options.HeapsterReplicas = 2
+	p.AddOns.HeapsterMonitoring.Options.HeapsterServiceType = "ClusterIP"
```
alexbrand: How do you feel about leaving these empty instead of filling them in with the default value? I don't have strong feelings either way, but wanted to bounce the idea with you.

dkoshkin: As a user having it pre filled would help see what the value is supposed to be and what the current value is, especially the `sink: influxdb:http://heapster-influxdb.kube-system.svc:8086` option.

This is also inline with the rest of the plan.

alexbrand: Make sense. Thanks!

PR #677, opened 2017-07-25
Introduce contiv as a CNI provider

### LINKED ISSUE #671: Add Contiv as a CNI provider
Introduce support for Contiv as a CNI provider.

- [x] Figure out if we can use etcd v3
- [x] Figure out if we can use secure etcd endpoints (how to pass certificates to wherever they are needed): 
Doesn't look like contiv currently supports an etcd cluster over HTTPS,
 as it uses the default HTTP transport in the etcd client,
 and there is not way to override it.
 (https://github.com/contiv/netplugin/blob/master/state/etcdstatedriver.go#L62).
- [x] Figure out if contiv supports multiple etcd nodes
Contiv currently doesn't support multiple etcd nodes in the backend. It only sets one URL when building the etcd client:
 (https://github.com/contiv/netplugin/blob/master/state/etcdstatedriver.go#L63).
- [ ] Translate installation script into an ansible play
- [x] Figure out what to do about taints - the taint is used for a replicaset that should only run on master nodes. We can probably rework this to use a daemonset.
- [x] "Own" the resources that get deployed
- [ ] How to ship the contiv binary (`netctl`) - probably use a container similar to what we'll do with CNI

alexbrand:
Fixes #671 

To do:
- [x] Move cni installation to container
- [x] Figure out how to distribute netctl
- [x] Add integration test coverage
- [x] Figure out issue with kuberang
- [x] Move netmaster to daemonset that gets deployed on master nodes

We'll do this in follow up PRs:
- [ ] Review contiv options we want to expose in plan file
- [ ] Figure out upgrades with contiv

### DIFF
--- a/ansible/_contiv.yaml
+++ b/ansible/_contiv.yaml
@@ -1,0 +1,11 @@
+---
+  - hosts: master:worker:ingress:storage
+    any_errors_fatal: true
+    name: "{{ play_name | default('Start Contiv Network Components') }}"
+    serial: "{{ serial_count | default('100%') }}"
+    become: yes
+    vars_files:
+      - group_vars/all.yaml
+
+    roles:
+      - contiv
--- a/ansible/roles/contiv/tasks/main.yaml
+++ b/ansible/roles/contiv/tasks/main.yaml
@@ -1,0 +1,141 @@
+---
+  # Pre-download contiv images
+  - name: download contiv container images
+    command: docker pull {{ contiv_netplugin_img }}
+
+  # Create a secret that contains the contiv certificates.
+  # Runs on the first master node if we find that the secret does 
+  # not exist
+  - name: check if contiv proxy server certs secert exists
+    command: kubectl -n kube-system get secret {{ contiv.proxy_server_certs_secret_name }} -o jsonpath='{.metadata.name}' --ignore-not-found 
+    register: certs_check_result
+
+  - block:
+    - name: create tmp dir for contiv certs
+      file:
+        path: "/tmp/contiv-certs"
+        state: directory
+
+    - name: copy contiv proxy server certificates
+      copy:
+        src: "{{ tls_directory }}/{{ item }}"
+        dest: "/tmp/contiv-certs/{{ item }}"
+        owner: "{{ kubernetes_owner }}"
+        group: "{{ kubernetes_group }}"
+        mode: "{{ kubernetes_service_mode }}"
+      with_items:
+        - "{{ contiv.certs.proxy_server_key_filename }}"
+        - "{{ contiv.certs.proxy_server_cert_filename }}"
+
+    - name: create secret for contiv proxy server certificates
+      command: kubectl -n kube-system create secret tls {{ contiv.proxy_server_certs_secret_name }} --key=/tmp/contiv-certs/{{ contiv.certs.proxy_server_key_filename }} --cert=/tmp/contiv-certs/{{ contiv.certs.proxy_server_cert_filename }}
+
+    - name: delete temp contiv certs
+      file:
+        path: "/tmp/contiv-certs"
+        state: absent
+    when: inventory_hostname == groups['master'][0] and certs_check_result.stdout != contiv.proxy_server_certs_secret_name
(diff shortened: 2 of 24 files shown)


ansible/_contiv.yaml
```diff
@@ -0,0 +1,12 @@
+---
+  - hosts: master:worker:ingress:storage
+    any_errors_fatal: true
+    name: "{{ play_name | default('Start Contiv Network Components') }}"
+    serial: "{{ serial_count | default('100%') }}"
+    remote_user: root
+    become_method: sudo
```
dkoshkin: We want to use become: yes after #668

ansible/roles/contiv/templates/configmap.yaml
```diff
@@ -0,0 +1,26 @@
...
+      "cniVersion": "0.1.0",
+      "name": "contiv-net",
+      "type": "contivk8s"
+    }
+  config: |-
+    {
+       "K8S_API_SERVER": "https://{{ kubernetes_load_balanced_fqdn }}:6443",
+       "K8S_CA": "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+       "K8S_KEY": "",
+       "K8S_CERT": "",
```
dkoshkin: Do we not need `K8S_KEY` and `K8S_CERT`?

alexbrand: It uses the default locations for the key and cert.

dkoshkin:
Should we pre-pull docker images similar to how we do it for Calico? That was originally added because we were seeing timeout of the pods starting on slow networks.

alexbrand:
Definitely. Thanks @dkoshkin!

ansible/roles/contiv/tasks/main.yaml
```diff
@@ -0,0 +1,141 @@
...
+      command: docker run --net host --rm --entrypoint /contiv/bin/netctl {{ contiv_netplugin_img }} --netmaster http://localhost:9999 net ls
+      register: contiv_networks
+      run_once: true
+
+    - name: create infra network if missing
+      command: docker run --net host --rm --entrypoint /contiv/bin/netctl {{ contiv_netplugin_img }} --netmaster http://localhost:9999 net create -n infra -s 132.1.1.0/24 -g 132.1.1.1 contivh1 # using defaults from installation script
+      when: "'contivh1' not in contiv_networks.stdout"
+      run_once: true
+
+    - name: restart netplugin to propagate forwarding mode change
```
dkoshkin: This seems crazy, but if thats what it takes :(

ansible/roles/kubelet/templates/kubelet.service
```diff
@@ -21,6 +21,9 @@ ExecStart=/usr/bin/kubelet \
   --require-kubeconfig=true \
   --kubeconfig={{ kubernetes_kubeconfig.kubelet }} \
   --node-ip={{ internal_ipv4 }} \
+{% if inventory_hostname in groups['master'] %}
+  --node-labels=node-role.kubernetes.io/master= \
```
dkoshkin: Is this a common label?

pkg/install/validate_test.go
```diff
@@ -1005,17 +1005,12 @@ func TestCNIAddOn(t *testing.T) {
 			},
 			valid: true,
 		},
-		// {
```
dkoshkin: Can remove.

ansible/roles/contiv/tasks/main.yaml
alexbrand: Yup, that's what the official installation script does as well.

ansible/roles/kubelet/templates/kubelet.service
alexbrand: There is still some discussion around this label, but both kubeadm and kargo are using it. Discussion here: https://groups.google.com/forum/#!topic/kubernetes-sig-cluster-lifecycle/_Ys5ABKgcbM

dkoshkin:
LGTM!

PR #679, opened 2017-07-26
Better error messages during online upgrades

### LINKED ISSUE #670: Using [SKIPPED] for un-supported component upgrades
I feel we should be using `[SKIPPED]` instead of `[ERROR]` when KET is upgrading components we don't support, see below ...

```
ingress1 [ingress]                                                              [ERROR]
- Performing an online upgrade of ingress nodes is not supported.
storage1 [storage]                                                              [ERROR]
- Performing an online upgrade of storage nodes is not supported.
storage2 [storage]                                                              [ERROR]
- Performing an online upgrade of storage nodes is not supported. 
```

The user should only be seeing the word `[ERROR]` when something has gone wrong with the process being executed.

dkoshkin:
Fixes #670 

```
Validate Online Upgrade=============================================================
etcd001 [etcd]                                                                  [WARNING]
- This node is part of an etcd cluster that has less than 3 members. Upgrading it may make the cluster unavailable.
master001 [master]                                                              [WARNING]
- This is the only master node in the cluster. Upgrading it may make the cluster unavailable.
- This node is acting as the load balanced endpoint for the master nodes. Upgrading it may make the cluster unavailable
worker001 [worker ingress storage]                                              [WARNING]
- This is the only worker node in the cluster. Upgrading it may make cluster features unavailable.
- Pod "kube-system/calico-node-nfm61" is using HostPath volume "lib-modules", which is unsafe for upgrades.
- Pod "kube-system/calico-node-nfm61" is using HostPath volume "var-run-calico", which is unsafe for upgrades.
- Pod "kube-system/calico-node-nfm61" is using HostPath volume "cni-bin-dir", which is unsafe for upgrades.
- Pod "kube-system/calico-node-nfm61" is using HostPath volume "cni-net-dir", which is unsafe for upgrades.
- Pod "kube-system/calico-node-nfm61" is using HostPath volume "etcd-certs", which is unsafe for upgrades.
- Pod "kube-system/calico-policy-controller-4277936060-hm2z9" is using HostPath volume "ca", which is unsafe for upgrades.
- Pod "kube-system/calico-policy-controller-4277936060-hm2z9" is using HostPath volume "cert", which is unsafe for upgrades.
- Pod "kube-system/calico-policy-controller-4277936060-hm2z9" is using HostPath volume "key", which is unsafe for upgrades.
- Pod "kube-system/kube-proxy-worker001" is using HostPath volume "ssl-certs", which is unsafe for upgrades.
- Pod "kube-system/kube-proxy-worker001" is using HostPath volume "kubeconfig", which is unsafe for upgrades.
- Pod "kube-system/kube-proxy-worker001" is using HostPath volume "etc-kube-ssl", which is unsafe for upgrades.
- Pod managed by ReplicaSet "kube-system/calico-policy-controller-4277936060" is running on this node, and the ReplicaSet does not have a replica count greater than 1.
- All the replicas that belong to the ReplicaSet "kube-system/calico-policy-controller-4277936060" are running on this node.
- Pod managed by DaemonSet "kube-system/default-http-backend" is running on this node, and no other nodes are capable of hosting this daemon. Upgrading it may make the daemon unavailable.
- Pod managed by DaemonSet "kube-system/gluster-healthz" is running on this node, and no other nodes are capable of hosting this daemon. Upgrading it may make the daemon unavailable.
- Pod managed by ReplicaSet "kube-system/heapster-622576002" is running on this node, and the ReplicaSet does not have a replica count greater than 1.
- All the replicas that belong to the ReplicaSet "kube-system/heapster-622576002" are running on this node.
- Pod managed by ReplicaSet "kube-system/heapster-influxdb-1807514822" is running on this node, and the ReplicaSet does not have a replica count greater than 1.
- All the replicas that belong to the ReplicaSet "kube-system/heapster-influxdb-1807514822" are running on this node.
- Pod managed by DaemonSet "kube-system/ingress" is running on this node, and no other nodes are capable of hosting this daemon. Upgrading it may make the daemon unavailable.
- Pod managed by ReplicaSet "kube-system/kube-dns-3793137827" is running on this node, and the ReplicaSet does not have a replica count greater than 1.
- All the replicas that belong to the ReplicaSet "kube-system/kube-dns-3793137827" are running on this node.
- The pod "kube-system/kube-proxy-worker001" is not being managed by a controller. Upgrading this node might result in data or availability loss.
- Pod managed by ReplicaSet "kube-system/kubernetes-dashboard-4218587249" is running on this node, and the ReplicaSet does not have a replica count greater than 1.
- All the replicas that belong to the ReplicaSet "kube-system/kubernetes-dashboard-4218587249" are running on this node.
- Pod managed by ReplicaSet "kube-system/tiller-deploy-4160234829" is running on this node, and the ReplicaSet does not have a replica count greater than 1.
- All the replicas that belong to the ReplicaSet "kube-system/tiller-deploy-4160234829" are running on this node.
- Performing an upgrade of ingress nodes may cause it to be temporary offline.
- Performing an upgrade of storage nodes may cause it to be temporary offline.

Ignoring safety checks and continuing with the upgrade                          [WARNING]
```

### DIFF
--- a/pkg/cli/upgrade.go
+++ b/pkg/cli/upgrade.go
@@ -265,7 +265,11 @@ func upgradeNodes(in io.Reader, out io.Writer, plan install.Plan, opts upgradeOp
 			util.PrettyPrint(out, "%s %v", node.Node.Host, node.Roles)
 			errs := install.DetectNodeUpgradeSafety(plan, node.Node, kubeClient)
 			if len(errs) != 0 {
-				util.PrintError(out)
+				if opts.ignoreSafetyChecks {
+					util.PrintWarn(out)
+				} else {
+					util.PrintError(out)
+				}
 				fmt.Fprintln(out)
 				for _, err := range errs {
 					fmt.Println("-", err.Error())
--- a/pkg/install/upgrade.go
+++ b/pkg/install/upgrade.go
@@ -24,40 +24,40 @@ type etcdNodeCountErr struct{}
 
 func (e etcdNodeCountErr) Error() string {
 	return "This node is part of an etcd cluster that has less than 3 members. " +
-		"Upgrading it will make the cluster unavailable."
+		"Upgrading it may make the cluster unavailable."
 }
 
 type masterNodeCountErr struct{}
 
 func (e masterNodeCountErr) Error() string {
 	return "This is the only master node in the cluster. " +
-		"Upgrading it will make the cluster unavailable."
+		"Upgrading it may make the cluster unavailable."
 }
 
 type masterNodeLoadBalancingErr struct{}
 
 func (e masterNodeLoadBalancingErr) Error() string {
 	return "This node is acting as the load balanced endpoint for the master nodes. " +
-		"Upgrading it will make the cluster unavailable"
+		"Upgrading it may make the cluster unavailable"
 }
 
 type ingressNotSupportedErr struct{}
 
 func (e ingressNotSupportedErr) Error() string {
-	return "Performing an online upgrade of ingress nodes is not supported."
+	return "Upgrading this node may result in storage volumes becoming temporarily unavailable."
 }
 
 type storageNotSupportedErr struct{}
 
 func (e storageNotSupportedErr) Error() string {
-	return "Performing an online upgrade of storage nodes is not supported."
+	return "Upgrading this node may result in service unavailability if clients are accessing services directly through this ingress point."
 }
 
 type workerNodeCountErr struct{}
 
 func (e workerNodeCountErr) Error() string {
 	return "This is the only worker node in the cluster. " +
-		"Upgrading it will make cluster features unavailable."
+		"Upgrading it may make cluster features unavailable."
 }
 
 type podUnsafeVolumeErr struct {
@@ -90,7 +90,7 @@ type podUnsafeDaemonErr struct {
 
 func (e podUnsafeDaemonErr) Error() string {
 	return fmt.Sprintf(`Pod managed by DaemonSet "%s/%s" is running on this node, and no other nodes `+
-		"are capable of hosting this daemon. Upgrading it will make the daemon unavailable.", e.dsNamespace, e.dsName)
+		"are capable of hosting this daemon. Upgrading it may make the daemon unavailable.", e.dsNamespace, e.dsName)
 }
 
 type unmanagedPodErr struct {


alexbrand:
I think we should be more specific in the warnings for ingress and storage nodes.

For storage, the problem is that we currently don't check if there is a brick on the node that would cause a volume to become unavailable. With that said, we haven't upgraded gluster yet, but I think we should still make the user aware of potential issues so that the validation is already there when we actually do upgrade gluster. I think a more specific warning could say:
```
Upgrading this node may result in storage volumes becoming unavailable.
```

For ingress, the problem is that we don't know if there is a load balancer in front of all the ingress nodes. Furthermore, if there is a single ingress node, taking down the ingress controller *will* cause downtime, as clients won't be able to access services exposed via ingress while the ingress controller is down. A more specific warning for ingress could be something along the lines of:
```
Upgrading this node may result in service unavailability if clients are accessing services directly through this ingress point.
```

swade1987:
@alexbrand if I have multiple ingress nodes will any message be displayed during the upgrade process, if so what is it?

alexbrand:
That is up for discussion. I think we should still warn given that we are not certain there is a load balancer in front of the ingress nodes.

The warnings are attached to a specific node, so I think we can keep the same message, regardless of the number of ingress nodes.

swade1987:
I question why we wouldn't have the following setup in the `kismatic-cluster.yaml` file.

```
ingress: 
  expected_count: 
  load_balanced_fqdn: 
  load_balanced_short_name: 
  nodes:
```

This means that we can have the same error message setup as we do for master nodes. As these are the two component types which are likely to have load balancers in front of them, this consistency makes sense to me.

alexbrand:
That is an excellent point. We use those fields in the master nodes for the certificates that we generate. 

Given that ingress certificates are a responsibility of the user, we wouldn't use those fields for anything during installation.

Do you think we should add them just for the purposes of this validation?

swade1987:
Yes as we should be validating that the ingress nodes can talk to the ingress load balancer during both installation and pre and post upgrade.

alexbrand:
We currently don't perform validation of load balancers, as they could be set up after the fact. We'd need to think more about this one, but probably under a new issue as it's not strictly related to this PR.

swade1987:
So, currently what checks are done pre-upgrade of Ingress and Storage nodes?

alexbrand:
We don't have any checks specific to ingress or storage nodes, which is why we had the message saying "online upgrades are not supported for ingress/storage" originally.

swade1987:
If we aren't doing any checks when an upgrade takes place are we simply just skipping upgrading any components currently running on these node types?

dkoshkin:
No we don't skip components, there are some checks that run during/after an upgrade. Ingress for ex. we validate the ingress pods came up correctly.

alexbrand:
@swade1987 @dkoshkin I think it's too close to the release to introduce new APIs and checks. I also want to discuss this more offline. How do you guys feel about improving the warning message for now, and revisiting this down the line?

swade1987:
@alexbrand sounds sensible, let's discuss this more and have a good implementation for the release after this one coming up thoughts?

alexbrand:
Sounds like a plan @swade1987!

dkoshkin:
Merging, @alexbrand and @swade1987 you guys ok with that?

alexbrand:
LGTM!

swade1987:
I am not 100% convinced but due to time constraints it LGTM

PR #680, opened 2017-07-26
Simplify calico deployment and validation

dkoshkin:
- Validate calico nodes after upgrades in the same way as during install
- Don't use NodeSelector when deploying calico pods
  - This was originally added when upgrading from KET <`v1.3.0` but no longer required
- Label nodes with `kismatic/cni-provider=calico` instead of `kismatic/calico=true`


alexbrand:
LGTM

PR #682, opened 2017-07-26
Add support to configure any CNI after install

dkoshkin:
Let's talk about what word we want to use if not `custom`


alexbrand:
I am fine with `custom`. I can't think of a better alternative at this point.

PR #683, opened 2017-07-26
Update k8s to v1.7.2, heapster, helm, registry, dashboard, etcd

dkoshkin:
As part of this, removing the `"Installing with previous version of Kismatic"` test. This was initially added after the package repo was broken, however the `upgrade` tests are also testing this.
It is also cumbersome to maintain by manually changing the previous version.


integration/plan_patterns.go
```diff
@@ -44,6 +44,7 @@ const planAWSOverlay = `cluster:
   disconnected_installation: {{.DisconnectedInstallation}}
   disable_registry_seeding: {{.DisableRegistrySeeding}}
   networking:
+    type: overlay
```
alexbrand: can we add a comment on these that they are here because of backwards compat.

alexbrand:
LGTM

swade1987:
@alexbrand @dkoshkin are we going to be using Helm v2.5.1 in the v1.5.0 release of KET?

alexbrand:
@swade1987 Yep! Helm v2.5.1

PR #684, opened 2017-07-26
Preflight throw up "device" attribute not found

prat0318:
I was facing the following error:
```
Run Cluster Pre-Flight Checks                                                   [ERROR]
- Task: fail if the block device is already mounted
  kubemedium02: The conditional check 'item.device == "{{ docker_direct_lvm_block_device_path }}"' failed. The error was: error while evaluating conditional (item.device == "{{ docker_direct_lvm_block_device_path }}"): 'ansible.vars.unsafe_proxy.AnsibleUnsafeText object' has no attribute 'device'

The error appears to have been in '/home/prateek.agarwal/kismatic-v1.4.1/ansible/playbooks/roles/preflight/tasks/direct_lvm_preflight.yaml': line 19, column 5, but may

    with_items: ansible_mounts
  - name: fail if the block device is already mounted
    ^ here
                                                                                [ERROR]
```

Issue was `ansible_mounts` was taken as a literal list and was not expanded.

This patch fixes it.

kismatic version: v1.4.1

### DIFF
--- a/ansible/roles/preflight/tasks/direct_lvm_preflight.yaml
+++ b/ansible/roles/preflight/tasks/direct_lvm_preflight.yaml
@@ -14,5 +14,5 @@
   - name: fail if the block device is already mounted
     fail:
       msg: "Block deviced specified for docker storage is currently mounted. This should be an unmounted, unused device"
-    with_items: "(( ansible_mounts }}"
+    with_items: "{{ ansible_mounts }}"
     when: item.device == "{{ docker_direct_lvm_block_device_path }}"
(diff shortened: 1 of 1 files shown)


ansible/roles/preflight/tasks/direct_lvm_preflight.yaml
```diff
@@ -14,5 +14,5 @@
   - name: fail if the block device is already mounted
     fail:
       msg: "Block deviced specified for docker storage is currently mounted. This should be an unmounted, unused device"
-    with_items: ansible_mounts
+    with_items: "(( ansible_mounts }}"
```
alexbrand: Should this start with `{{` instead of `((`?

prat0318: My bad. fixed.

dkoshkin:
Thanks @prat0318 LGTM. @alexbrand feel free to merge if you think its good. 
We do a similar task in `volume-add` role.

prat0318:
Thanks for merging. Any ETAs of a patch release cut with this commit anytime soon?

alexbrand:
We plan to release next week.

PR #698, opened 2017-08-01
Set environment on commands that actually need it

### LINKED ISSUE #697: Installation fails on packet when helm is enabled
The problem is that we don't specify the kubeconfig file, which means it is going to try accessing `localhost`. On Ubuntu running on packet, `localhost` does not resolve to `127.0.0.1`. (Have not tried CentOS).

```
Initialize Helm and Start Tiller                                                [ERROR]
- Task: create new serviceaccount for tiller
  kismatic-master-0-1501615941:                                                 [ERROR]
---- STDERR ----
The connection to the server localhost:8080 was refused - did you specify the right host or port?
---------------
error installing: error running playbook: error running ansible: exit status 2
```

alexbrand:
This makes it more apparent 1) why that environment variable is being set and 2) which commands actually need it.

Fixes #697

### DIFF
--- a/ansible/_helm.yaml
+++ b/ansible/_helm.yaml
@@ -5,9 +5,6 @@
     become: yes
     run_once: true
 
-    environment:
-      KUBECONFIG: "{{ local_kubeconfig_directory }}"
-
     tasks:
       - name: "create new serviceaccount for tiller"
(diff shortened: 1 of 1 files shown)


dkoshkin:
LGTM!

PR #699, opened 2017-08-01
Set timeout on the drain command

### LINKED ISSUE #695: Kubectl gets stuck draining nodes during an upgrade
I have seen tests that are not making forward progress because kubectl is stuck draining one of the nodes during an upgrade. I have not been able to track down the issue yet, but it seems like it could be an issue in kubectl per se.

alexbrand:
Attempts to improve / fix #695


dkoshkin:
LGTM!

PR #703, opened 2017-08-02
Components on master node should talk to local API

alexbrand:
Components running on master nodes that require access to the Kubernetes
API server should use the local API server, instead of going out to the
load balanced endpoint.

Fix #703


dkoshkin:
LGTM!

PR #708, opened 2017-08-03
Improve the diagnostics UX

### LINKED ISSUE #706: Diagnose UX is misleading
We have seen many occasions where the user expects the diagnose process to actually interpret the data that is being collected, and tell them if everything is "OK". 

Furthermore, the fact that we do print "OK" makes the user think KET has evaluated the data and concluded that there is no issue.

The diagnose UX needs to communicate that all we are doing is getting diagnostic information that can then be used by the user to further investigate the issue, or to send it to the KET team.

alexbrand:
Fix #706 

New output:
```
✗ ./kismatic diagnose -f kismatic-cluster-1.yaml

Gathering Diagnostic Data===========================================================
Validate SSH connectivity to nodes                                              [OK]
Gather Node Diagnostics                                                         [OK]

Finished gathering diagnostic information from all nodes in the cluster.
You may find the diagnostic data in the ./diagnostics directory.
```

### DIFF
--- a/ansible/_diagnose-nodes.yaml
+++ b/ansible/_diagnose-nodes.yaml
@@ -1,7 +1,6 @@
 ---
   - hosts: all
-    name: "Diagnose Nodes"
-    #become: yes
+    name: "Gather Node Diagnostics"
     vars_files:
       - group_vars/all.yaml
       - group_vars/diagnostics.yaml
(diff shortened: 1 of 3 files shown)


dkoshkin:
LGTM

alexbrand:
@dkoshkin How do you feel about adding `.log` extensions to all the collected files? As a user, it tells me what these files are and it makes it easier to just double click on them to open them up.

dkoshkin:
Yeah that makes sense to me.

PR #711, opened 2017-08-03
Improved the commenting of the plan file template

### LINKED ISSUE #696: Commented plan file is hard to read
The current commenting strategy that has comments in line with the actual fields has grown a bit unreadable. Some of the lines wrap around and make it even harder to understand what is going on.

I think we should template out a plan file in the following manner:
* Add comments above each field that requires a comment. Break the line into multiple lines if it is longer than 80 characters.
* Options should go on the next line, after explaining the field.

For example:
```
add_ons:
  cni:
    disable: false
    # The provider that should be used for the CNI add-on. The 'custom' 
    # option will result in a CNI ready cluster, however it is up to you to 
    # configure a CNI plugin after the installation.
    # Options: 'calico','weave','contiv','custom'. 
    provider: calico                                        
    options:
      calico:
        # The calico data path mode. Routed pods can be addressed from outside
        # the Kubernetes cluster; Overlay pods can only address each other.
        # Options: 'overlay','routed'. 
        mode: overlay                                                    
```

alexbrand:
Fix #696 

Comments now appear above the line that is being commented. The golden files used for testing are examples of what the templates look like now. Looking for feedback @dkoshkin @swade1987 @ikester 

[Sample without storage](https://github.com/apprenda/kismatic/pull/711/files#diff-da188fc8492246f027b2d7d11de02481)

[Sample with storage nodes/nfs](https://github.com/apprenda/kismatic/pull/711/files#diff-f1f0da9e15b0a411f3736f1871297191)

The empty string is still being included in the template `""`. We could remove them by running 
```
strings.Replace(line, `""`, ``, 1)
``` 
on all lines, but not sure about it.

### DIFF
--- a/pkg/cli/plan.go
+++ b/pkg/cli/plan.go
@@ -87,64 +87,18 @@ func doPlan(in io.Reader, out io.Writer, planner install.Planner, planFile strin
 	fmt.Fprintf(out, "- %d nfs volumes\n", nfsVolumes)
 	fmt.Fprintln(out)
 
-	template := planTemplate{
-		etcdNodes:    etcdNodes,
-		masterNodes:  masterNodes,
-		workerNodes:  workerNodes,
-		ingressNodes: ingressNodes,
-		storageNodes: storageNodes,
-		nfsVolumes:   nfsVolumes,
+	planTemplate := install.PlanTemplateOptions{
+		EtcdNodes:    etcdNodes,
+		MasterNodes:  masterNodes,
+		WorkerNodes:  workerNodes,
+		IngressNodes: ingressNodes,
+		StorageNodes: storageNodes,
+		NFSVolumes:   nfsVolumes,
 	}
-
-	plan := buildPlan(template)
-	// Write out the plan
-	if err = install.WritePlanTemplate(plan, planner); err != nil {
+	if err = install.WritePlanTemplate(planTemplate, planner); err != nil {
 		return fmt.Errorf("error planning installation: %v", err)
 	}
 	fmt.Fprintf(out, "Wrote plan file template to %q\n", planFile)
 	fmt.Fprintf(out, "Edit the plan file to further describe your cluster. Once ready, execute the \"install validate\" command to proceed.\n")
 	return nil
-}
-
-type planTemplate struct {
-	etcdNodes    int
-	masterNodes  int
-	workerNodes  int
-	ingressNodes int
-	storageNodes int
-	nfsVolumes   int
-}
-
-func buildPlan(template planTemplate) *install.Plan {
-	// Create a plan
-	masterNodeGroup := install.MasterNodeGroup{}
-	masterNodeGroup.ExpectedCount = template.masterNodes
-	plan := install.Plan{
-		Etcd: install.NodeGroup{
-			ExpectedCount: template.etcdNodes,
-		},
-		Master: masterNodeGroup,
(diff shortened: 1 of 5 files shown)


swade1987:
@alexbrand please remove the `""` i have to remove them every time I created a plan file. They don't add any value being there.

dkoshkin:
@swade1987 why do you have to remove it?

swade1987:
@dkoshkin see example below, I remember before when I used the quotes execution failed.

```
worker:
  expected_count: 3
  nodes:
  - host: worker1
    ip: 46.101.87.132
  - host: worker2
    ip: 46.101.87.171
  - host: worker3
    ip: 139.59.172.127
```

dkoshkin:
2 things:
1. Would it be better no to wrap the comments and just let the users editor wrap the line if required.
2. What if we add an optional flag to not print out the comments?

dkoshkin:
@swade1987 that should not be the case, if it still is thats an issue.

ikester:
I would add a newline before every comment section and I would keep the hard wrapping at column 80. This is pretty standard UNIX/Linux practice and makes it readable even on a small TTY connection.

Having a flag to skip the comments is a nice idea. I wonder how much it will be used.

Example:

```
name: kubernetes

# This password is used to login to the Kubernetes Dashboard and can also be
# used for administration without a security certificate.
admin_password: password

# When true, installation will not occur if any node is missing the correct
# deb/rpm packages. When false, the installer will attempt to install missing
# packages for you.
disable_package_installation: false
```

swade1987:
The YAML file is very long and its many because of the comments, the more and more add-ons and functionality we add the longer that file is going to become, as we need to comment those features as well. I would advise the following:

- Keep the YAML file short and sweet reference versioned documentation.
- Have clear versioning of the docs describing each line of the plan file per release.

One of the main reasons for this apart from having tidy YAML files is it's then easier to look at the documentation for older versions of the plan file without having to download an old version of Kismatic and execute ./kismatic install plan and also vice-versa for newer releases.

My suggestion layout would be:

```
# ================
# For more information on the structure of this YAML file please see the link below
# https://github.com/apprenda/kismatic/blob/master/docs/plan-file/v1.3.0/breakdown.md
# ================
cluster:
  name: kubernetes
```

That way all we would need to do is replace `v1.3.0` with the version of kismatic constructing the plan file.

alexbrand:
I really like the idea of linking the documentation at the top of the file.

In terms of the inline docs, I think there's value in having them around, so that I don't have to switch contexts when building my plan file. I am sure there are two camps to this, so maybe having a flag to disable commenting is the way to go.

swade1987:
Versionable documentation is extremely important as we add more and more functionality and increase the matrix of possible Kismatic versions we need to support. 

We want to have this outside the YAML file as it should be allowed to be maintained and updated out-of-band.

mammerman:
FWIW I agree that linking out to separately maintained docs is the right way to do this.

alexbrand:
Thanks @mammerman and @swade1987. 

To be clear, I am 100% for having separate, versioned reference documentation for the configuration file, which is why I have already started working on that. You may find that work in #712.

I think the discussion in this PR is around the commenting of the YAML file itself. We can probably take this offline for a more sync conversation.

dkoshkin:
Rebase, but besides that LGTM, lets merge this in after.

PR #712, opened 2017-08-03
Add reference documentation for plan file fields [skip ci]

alexbrand:
Added new reference documentation for the plan file. We currently don't have this and it's a problem. I took a first stab with a format that I thought works, but want to see what others think.

Also added a table of contents to our documentation to improve navigation and discoverability.

PTAL @dkoshkin @ikester @swade1987 @a-mccarthy


dkoshkin:
Did you manually have to generate the `plan_file`?!

alexbrand:
Hehe. I did this time, but started looking into automating it. Still want to get eyes on this ASAP. 😛

docs/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Reference
+- [Plan File](plan_file.md)
+- [Certificate Generation](certificates.md)
+- [Docker Configuration](docker.md)
+- [Troubleshooting](troubleshooting.md)
+- [Troubleshooting Calico](troubleshooting-calico.md)
+
+## Upgrade Notes
+### v1.3.0
```
alexbrand: I should probably put the latest versions on the top?

swade1987:
I am going to say something some what controversial here.

The lack of standardised documentation means we are having to document our YAML file like this.

If we had a documented plan file  (per release) as part of our repo and reference the URL in the YAML file I would suggest we could remove the comments in the plan file altogether.

This means we would only need to do two changes per release.

1. Create a new version specific plan file breakdown document in our repo.
2. Pin the YAML URL to be specific to the version of the document in the repo.

Thoughts?

dkoshkin:
LGTM for now! I do think we will want an automated way to at least verify the fields are all there.

alexbrand:
@swade1987 I am not sure why you think your comment is controversial. 

In any case, I am not sure I fully understand your suggestion. Do you think we should only have a commented YAML file checked into the repo somewhere, and that should be our documentation for the plan file?

swade1987:
I feel the YAML file should simply contain the following:

```
# ================
# For more information on the structure of this YAML file please see the link below
# https://github.com/apprenda/kismatic/blob/master/docs/plan-file/v1.3.0/breakdown.md
# ================
cluster:
  name: kubernetes
```

That way all we would need to do is replace `v1.3.0` with the version of kismatic constructing the plan file.

alexbrand:
Okay. I think that is a separate discussion from this. This PR is not about the YAML plan file or the comments in it, but about introducing reference documentation for each field in the plan file.

ikester:
I think it's useful to have inline comments (most Linux config files do). But having the option to generate the plan file without them (for Pros) may be nice option.

alexbrand:
I have updated the title of this PR to better reflect the content. I think there is some confusion with this PR and #711

swade1987:
The YAML file is very long and its many because of the comments, the more and more add-ons and functionality we add the longer that file is going to become, as we need to comment those features as well. I would advise the following:

- Keep the YAML file short and sweet reference versioned documentation.
- Have clear versioning of the docs describing each line of the plan file per release.

One of the main reasons for this apart from having tidy YAML files is it's then easier to look at the documentation for older versions of the plan file without having to download an old version of Kismatic and execute `./kismatic install plan` and also vice-versa for newer releases.

dkoshkin:
Would it make more sense for the index to use the same case as what it is in the plan file.

* cluster
   * name
   * ...
   * networking
        * pod_cidr_block

docs/plan_file.md
```diff
@@ -0,0 +1,227 @@
...
+    * [direct_lvm](#docker_storage_direct_lvm_enabled)
+      * [enabled](#docker_storage_direct_lvm_enabled)
+      * [block_device](#docker_storage_direct_lvm_block_device)
+      * [enable_deferred_deletion](#docker_storage_direct_lvm_enable_deferred_deletion)
+* [docker_registry](#docker_registry)
+  * [setup_internal](#docker_registry_setup_internal)
+  * [address](#docker_registry_address)
+  * [port](#docker_registry_port)
+  * [ca](#docker_registry_ca)
+* [add-ons](#add-ons)
```
dkoshkin: should be `add_ons` also update the link `#add_ons`

dkoshkin:
Two things we talked about:
* add the breadcrumb links
* add a task in the makefile to run the whole command

Besides that LGTM, lets merge this in after that.

PR #728, opened 2017-08-09
Fix dashboard when generating plan file

### LINKED ISSUE #727: Duplicate "dashboard" flag in 1.5.1 kismatic-cluster.yaml
After downloading kismatic v.1.5.1 and executing `./kismatic install plan` the follow is shown in the plan file.

```
dashboard: null
dashbard: null
```

alexbrand:
Fixes #727

### DIFF
--- a/pkg/install/plan.go
+++ b/pkg/install/plan.go
@@ -155,7 +155,10 @@ var yamlKeyRE = regexp.MustCompile(`[^a-zA-Z]*([a-z_\-A-Z]+)[ ]*:`)
 
 // Write the plan to the file system
 func (fp *FilePlanner) Write(p *Plan) error {
-	oneTimeComments := commentMap
+	oneTimeComments := map[string]string{}
+	for k, v := range commentMap {
+		oneTimeComments[k] = v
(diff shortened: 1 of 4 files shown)


pkg/install/plan.go
```diff
@@ -155,7 +155,10 @@ var yamlKeyRE = regexp.MustCompile(`[^a-zA-Z]*([a-z_\-A-Z]+)[ ]*:`)
 
 // Write the plan to the file system
 func (fp *FilePlanner) Write(p *Plan) error {
-	oneTimeComments := commentMap
+	oneTimeComments := map[string]string{}
```
dkoshkin: What was the reason to do this?

alexbrand: The new unit test surfaced a bug. The original intent here was to create a copy of comment map because we delete keys from it, but we were only making a copy of the reference to the map, and not the map itself.

With this change, we create a copy of the comment map.

dkoshkin:
LGTM. The new test should prevent these kind of bugs in the future.

PR #732, opened 2017-08-09
Fix dns pod validation issue

### LINKED ISSUE #731: DNS pod validation attempts to print logs when they were not collected
When the get logs is skipped, we are still trying to log them. We should not try to print the messages if they are not there.

```
- Running task: get the logs of the first DNS pod that did not start up in time
  ip-10-0-3-245                                                                 [SKIPPED]
- Running task: fail if DNS pods are not ready
  ip-10-0-3-245 the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'stdout'

The error appears to have been in '/tmp/kismatic-041795636/tests/test-377618310/ansible/playbooks/roles/kube-dns/tasks/main.yaml': line 37, column 7, but may

    - name: fail if DNS pods are not ready
      ^ here
                                                                                [ERROR]
```

alexbrand:
Fixes #731

### DIFF
--- a/ansible/roles/kube-dns/tasks/main.yaml
+++ b/ansible/roles/kube-dns/tasks/main.yaml
@@ -20,6 +20,12 @@
       delay: 10
       failed_when: false # We don't want this task to actually fail (We catch the failure with a custom msg in the next task)
     
+    - name: fail if DNS pod validation command returned an error
+      fail:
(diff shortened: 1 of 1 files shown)


dkoshkin:
LGTM

PR #734, opened 2017-08-10
Updating Slack URL

jaycoon:
Now references the subdomain: [http://slack.kismatic.com/](http://slack.kismatic.com/)
(can also use [http://slack-signup.kismatic.com/](http://slack-signup.kismatic.com/))


alexbrand:
Thanks @jaycoon!

PR #743, opened 2017-08-17
Add cloud-provider proposal

dkoshkin:


proposals/cloud-provider.md
```diff
@@ -0,0 +1,135 @@
...
+
+More detail [here](https://kubernetes.io/docs/getting-started-guides/scratch/#cloud-providers)
+
+### Plan File Changes
+These options will be exposed in the plan file as follows: 
+```
+cluster:
+	kube_apiserver:
+    option_overrides: {}
+	cloud_provider:
```
alexbrand: Is this a first class option under `cluster`? It's unclear right now as it seems there might be an indentation problem?

proposals/cloud-provider.md
```diff
@@ -0,0 +1,135 @@
...
+		provider:
+		config:
+```
+
+Initially we will target support(with tests) for `aws` and `openstack`, however we should not prevent the user from using any other provider.
+This can be accomplished with a warning at runtime or documentation.
+
+### aws provider
+`aws` does not require the `cloud-config` and utilizes AMI policies to interact with the API.
```
alexbrand: s/AMI/IAM 😄

proposals/cloud-provider.md
```diff
@@ -0,0 +1,135 @@
...
+metadata:
+  name: slow
+provisioner: kubernetes.io/aws-ebs
+parameters:
+  type: io1
+  zones: us-east-1a, us-east-1c
+  iopsPerGB: "10"
+```
+
+Sample poicies below:
```
alexbrand: These are IAM policies I assume?

proposals/cloud-provider.md
```diff
@@ -0,0 +1,135 @@
...
+[Global]
+auth-url = OS_AUTH_URL
+user-id = OS_USERNAME
+api-key = OS_PASSWORD
+tenant-id = OS_TENANT_ID
+tenant-name = OS_TENANT_NAME
+[LoadBalancer]
+subnet-id = 11111111-1111-1111-1111-111111111111
+```
+This will require testing and someone with opestack experience.
```
alexbrand: typo on opestack

proposals/cloud-provider.md
```diff
@@ -0,0 +1,135 @@
...
+                "elasticloadbalancing:*"
+            ],
+            "Resource": [
+                "*"
+            ]
+        }
+    ]
+}
+```
+Minion:
```
alexbrand: Can we call this worker instead?

proposals/cloud-provider.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Plan File Changes
+These options will be exposed in the plan file as follows: 
+```
+cluster:
+	kube_apiserver:
+    option_overrides: {}
+	cloud_provider:
+		provider:
+		config:
```
alexbrand: What are the mechanics for the config file? Is it an absolute path to a file on the installation machine that we then grab and copy to all the machines?

proposals/cloud-provider.md
dkoshkin: Why do you do this to me AWS?! :)

alexbrand:
This looks great to me overall! If we can avoid the absolute path requirement for the config file, that would be great.

Issue #724, opened 2017-08-08
Use official distribution packages for Kubernetes and other cluster components


alexbrand:
Given that the community has started building and maintaining official distribution packages for Kubernetes, we should transition over to using them instead of maintaining our own.

Overall, our packaging and distribution strategy should follow these principles for all components:
- Use upstream packages and bits unchanged
- Support natural way of distribution/packaging that is being created by the community
- If official distribution strategy is unavailable, prefer a container for simplification of disconnected install use case.

| Component | Distribution |
|-------------|-------------|
| Docker         | Official Docker packages for each distro |
| Etcd             | Tarball or Container image |
| Kubelet        | Official RPM/DEB (pending further investigation) |
| Control plane components | Official container images (unchanged) |
| CNI | Official distribution is not available. Using custom built container |
| CNI Add-on providers | Official container images (unchanged) |
| Other add-ons | Official container images (unchanged) |

**Upgrade Considerations**
Performing this transition will require the removal of our own packages and package repositories during the upgrade.

Open questions:
- [x] Why did kubeadm decide to use RPMs/DEBs instead of using containers for the Kubelet?

swade1987:
@alexbrand I was talking to Kris Nova about Kubeadm and see said it's extremely hard to run inside a container due to the volumes.

alexbrand:
Notes on why kubelet is not run inside container:
https://docs.google.com/document/d/1XLZR0OQckXVSB6xVcaqkSWUzufllOMRR9eeWEpGNjJY/edit

alexbrand:
Closed via #768

PR #756, opened 2017-08-23
Run etcd in container instead of systemd service

### LINKED ISSUE #568: With epel enabled on etcd nodes yum update breaks etcd
Hi,
Just a heads up for the community. If you enabled epel on the systems where etcd is built a yum update will overwrite the kismatic version of etcd with a new version which then then break kismatic etcd.
Kismatic etcd creates two files /usr/bin/etcd_k8s  /usr/bin/etcd_networking which will be overwritten if you do yum updates via cron and you have epel enabled. 
The solution is to either put an exclusion in /etc/yum.conf or disable epel. Just an fyi.

--
Ted

dkoshkin:
Fixes #568, partial work on #724 

As part of this PR, I also removed the KET `v1.3` upgrade code.


ansible/group_vars/etcd-k8s.yaml
```diff
@@ -1,7 +1,7 @@
 # etcd-k8s
 # etcd-install
 etcd_install_dir: /etc/etcd_k8s
-etcd_install_bin_name: etcd_k8s
+etcd_name: etcd_k8s
 etcdctl_install_bin_name: etcdctl
```
alexbrand: Should we clean this var up as well?

ansible/roles/etcd/tasks/main.yaml
```diff
@@ -12,47 +12,41 @@
...
+    command: "docker run --net=host --volume=/etc/ssl/certs/:/etc/ssl/certs/ --volume={{etcd_install_dir}}:{{etcd_install_dir}}:ro {{ etcd_img }} /usr/local/bin/etcdctl --endpoint='https://127.0.0.1:{{ etcd_service_client_port }}/' --cert-file={{ etcd_certificates.etcd_client }} --key-file={{ etcd_certificates.etcd_client_key }} --ca-file={{ etcd_certificates.ca }} cluster-health"
     register: result
     until: result|success
     retries: 3
     delay: 5
     when: "{{ etcd_insecure_validate|default('false')|bool == false }}"
 
-  - name: verify insecure {{ etcd_install_bin_name }} cluster health
-    command: "{{ bin_dir }}/{{ etcdctl_install_bin_name }} --endpoint='http://127.0.0.1:{{ etcd_service_client_port }}/' cluster-health"
+  - name: verify insecure {{ etcd_name }} cluster health
```
alexbrand: Can we get rid of the `insecure` here?

ansible/roles/etcd/templates/etcd.insecure.service
```diff
@@ -1,26 +1,37 @@
...
 Description=etcd key-value store
 Documentation=https://github.com/coreos/etcd
+After=docker.service
+Requires=docker.service
 
 [Service]
 User=root
-Type=notify
-ExecStart={{ bin_dir }}/{{ etcd_install_bin_name }} \
+ExecStartPre=-{{ bin_dir }}/docker stop {{ etcd_name }}
```
alexbrand: What's the `-` in front of these commands?

dkoshkin:
I figured the upgrades were a potential source of failure so did loads of testing for the upgrade from different versions of KET with different and the same `etcd` versions, all went well.

ansible/roles/etcd/templates/etcd.insecure.service
dkoshkin: Ignore failures, most of the time the docker container will not actually be running so this command will fail.

alexbrand: Got it!

ansible/roles/etcd/templates/etcd.insecure.service
```diff
@@ -1,26 +1,37 @@
...
+ExecStartPre=-{{ bin_dir }}/docker rm -f {{ etcd_name }}
+ExecStart={{ bin_dir }}/docker run \
+  -p {{ etcd_service_peer_port }}:{{ etcd_service_peer_port }} \
+  -p {{ etcd_service_client_port }}:{{ etcd_service_client_port }} \
+  --volume={{ etcd_service_data_dir }}:/etcd-data \
+  --volume={{ etcd_install_dir }}:{{ etcd_install_dir }}:ro \
+  --volume=/etc/ssl/certs/:/etc/ssl/certs/ \
```
alexbrand: Can this be set to read-only?

integration/prepare.go
```diff
@@ -47,7 +45,7 @@ type nodePrep struct {
 
 var ubuntu1604Prep = nodePrep{
 	CommandsToPrepRepo:         []string{copyKismaticKeyDeb, copyKismaticRepoDeb, updateAptGet},
-	CommandsToInstallEtcd:      []string{installCurlApt, installEtcdApt, installTransitionEtcdApt},
+	CommandsToInstallEtcd:      []string{installCurlApt, installEtcdApt},
```
alexbrand: Do we still need to install etcd using packages?

alexbrand:
Great stuff here! The last piece that I just started thinking about is whether we should run etcd in the host network namespace. I can't think of a specific reason at the moment, but I thought I would still bring it up.

dkoshkin:
From the [docs](https://coreos.com/etcd/docs/latest/v2/docker_guide.html) and other deployments I've seen it doesn't.
The one thing that is different with the docker deployment is we lost the `http` endpoint on the `127.0.0.1` IP.

alexbrand:
I think this is ready to go into the feature branch. I am going to kick off a full test run, and merge if everything looks good. 

If we find any outstanding work down the line, we can address in a follow-up issue.

alexbrand:
Most failures are related to the disconnected install, which we don't have an answer for right now.

Merging this in.

PR #758, opened 2017-08-24
Skip drain if API server is not running

### LINKED ISSUE #738: Unable to drain node during an upgrade if the API server is already down
In the case where the upgrade of a master node fails after shutting down the API server, the user has to manually restart the API server to be able to retry the upgrade. We should improve this to not require manual intervention by the user, and ideal not leave the master node in a "bad" state.

One think to keep in mind is that the cluster should still be functional at this point, assuming it has more than one master node.

alexbrand:
Fixes #738 
```
When upgrading master nodes, there is the possibility of entering a
scenario where the API server has been stopped, but the upgrade has to
be retried due to failure. For this reason, specifically in master
nodes, the drain command only runs if the API server is supposed to be
running.
```

### DIFF
--- a/ansible/_kube-drain-node.yaml
+++ b/ansible/_kube-drain-node.yaml
@@ -3,14 +3,29 @@
     hosts: master:worker:ingress:storage
     serial: 1
     tasks:
-      - name: "run kubectl drain"
+      - name: check API server pod manifest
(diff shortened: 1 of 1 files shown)


dkoshkin:
LGTM!

Issue #563, opened 2017-05-30
Unreachable node does not show up in kismatic info command output


bgregorius:
If you have a node that "goes away" the kismatic info command does not show the affected node in the output.

Discovered on an AWS cluster where a node started failing AWS status checks, couldn't ssh to the node, etc.

kismatic info stopped reporting that node as part of the cluster
kubectl get nodes however still showed the node with a NotReady status

The expectation here is that kismatic info should show all nodes that are part of the cluster until they have explicitly been removed.

alexbrand:
I tried to reproduce this issue but failed to do so. Before even displaying node information, we verify that we can SSH to all machines in the cluster.

My suspicion here is that the bad node was replaced with a new node with the same identity that was up and running. Given that we currently default the case of "no kismatic version file" to `v1.0.0`, the `kismatic info` command happily reported that.

We should improve this but I am dropping the priority for now.

PR #759, opened 2017-08-24
Stop defaulting "no ket version file" to 1.0.0

### LINKED ISSUE #716: Version gathering code does not elevate when checking /etc/kismatic-version
The version gathering code that runs during `info` or `upgrade` does not elevate when getting the contents of the `/etc/kismatic-version` file. Given that the file is created with the user's default umask, there will be cases where we will be unable to get the contents of the file.

When this happens, we default to `v1.0.0` which then causes downstream issues during an upgrade. The defaulting can now be removed, given that KET v.1.5.X does not support upgrades from version < 1.4.1. 

In the case where we are unable to read the contents of the file, we should return an error.

alexbrand:
Fixes #716, #563 

```
In the initial implementation of KET upgrades, a missing KET version
file would result in KET assuming the node was version 1.0.0. Given that KET no
longer supports upgrades from versions that don't have this file, we can
remove this special case.

Also set permissions on the KET version file marker.
```

### DIFF
--- a/ansible/roles/update_version/tasks/main.yaml
+++ b/ansible/roles/update_version/tasks/main.yaml
@@ -1,6 +1,6 @@
 ---
-  # write version file
   - name: write version file
     template:
       src: kismatic-version.j2
       dest: "/etc/kismatic-version"
+      mode: 0644
--- a/pkg/install/about.go
+++ b/pkg/install/about.go
@@ -23,11 +23,6 @@ type ListableNode struct {
 	Version semver.Version
 }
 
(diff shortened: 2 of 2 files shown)


dkoshkin:
What happens when the file can't be read or is just missing? Will it just fail?

alexbrand:
That's correct. The error is captured and displayed to the user.

With respect to #716, I decided to tackle the root cause which is that we were not setting proper permissions on the KET version file. Now that we are, we don't need to elevate to root to read the file.

dkoshkin:
LGTM!

PR #760, opened 2017-08-24
Cloud provider

### LINKED ISSUE #124: Cloud Provider Support
I created some AWS hosts and successfully used Kismatic to install Kubernetes. However, when I tried to use an EBS volume this failed. Checking /etc/systemd/system/kubelet.service on workers and masters I realised that --cloud-provider is empty. Manually setting --cloud-provider to "aws" and restarting the kubelet service then fails with:
```
Zone not specified in configuration file; querying AWS metadata service
Nov 25 06:25:04 ip-172-15-1-114.eu-west-1.compute.internal kubelet[9542]: error: failed to run Kubelet: could not init cloud provider "aws": error finding instance i-0054f74dfec8d7fc9: error listing AWS instances: NoCredentialProviders: no valid providers in chain
```

Looking again at kube-up for comparison, it seems that this is probably because Kismatic doesn't do the IAM role setup that kube-up does (see [here](https://medium.com/@canthefason/kube-up-i-know-what-you-did-on-aws-93e728d3f56a#.n80meac1x)) so it can't talk to AWS. That would mean it can't do any of the cloud integrations that Kubernetes is capable of.

Is my reasoning correct? And is AWS integration anywhere on the roadmap?

dkoshkin:
Fixes #124 

```
cluster
  kube_apiserver:
    option_overrides: {}
  cloud_provider:
    provider: ""                         # Options: 'aws','azure','cloudstack','fake','gce','mesos','openstack','ovirt','photon','rackspace','vsphere'. Leave empty for bare metal setups or other unsupported providers.
    config: ""                           # Path to the config file, leave empty if provider does not require it.
```
When using certain cloud-providers the `hostname-override` option is not always respected, k8s [issue](https://github.com/kubernetes/kubernetes/issues/48929).
For example the hostname that was used in AWS is not what is returned by `hostname`

Because of this we can no longer guarantee that the hostname provided in the plan file will be the same as the kubernetes node hostname.
We relied on knowing the hostname when looking up static pod names and labeling nodes with `cni-provider`, `ingress` and `storage`.

Fix:
* static pod names: added a `kismatic/nodeIP` label on all static pod spec files and use `--selector` when determining if the node is up
* node labels: use `--node-labels=` field in the kubelet, WARNING this is an Alpha feature and will need to be monitored with new k8s releases


ansible/roles/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -37,6 +38,14 @@ spec:
     - mountPath: /etc/ssl/certs
       name: ssl-certs-host
       readOnly: true
+    - name: usr-ca-certs-host
+      mountPath: /usr/share/ca-certificates
+      readOnly: true
+{% if cloud_provider is defined and cloud_provider == 'aws' and ansible_os_family == 'RedHat' %}
```
alexbrand: Is this requirement documented somewhere? Would be great to jot down why this is required.

ansible/roles/kubelet/templates/kubelet.service
```diff
@@ -20,6 +21,7 @@ ExecStart=/usr/bin/kubelet \
   --hostname-override={{ inventory_hostname }} \
   --require-kubeconfig=true \
   --kubeconfig={{ kubernetes_kubeconfig.kubelet }} \
+  --node-labels=kismatic/nodeIP={{ internal_ipv4 }},kismatic/cni-provider={{ cni.provider|default("")}}{% if 'ingress' in group_names%},kismatic/ingress=true{% endif %}{% if 'storage' in group_names%},kismatic/storage=true{% endif %} \
```
alexbrand: Heads up: we already use this flag in line 27

ansible/group_vars/all.yaml
```diff
@@ -157,6 +157,8 @@ local_kubernetes_master_ip: https://127.0.0.1:{{ kubernetes_master_secure_port }
 kubernetes_master_ip: https://{{ kubernetes_load_balanced_fqdn }}:{{ kubernetes_master_secure_port }}
 kubernetes_schedulable: "{% if 'worker' in group_names %}true{% else %}false{% endif %}"
 kube_dns_replicas: "{{ [2, groups['worker'] | length] | min }}"
+# cloud provider
+cloud_config: "{% if cloud_config_local is defined and cloud_config_local != '' %}{{ kubernetes_install_dir }}/cloud_config{% else %}{% endif %}"
```
alexbrand: Is this the path to the cloud config file? Wondering if we should use a file extension here. Maybe `cloud-provider.conf`? Not sure if there is an existing convention for this file in the community.

pkg/install/validate.go
```diff
@@ -216,6 +217,21 @@ func (s *SSHConfig) validate() (bool, []error) {
 	return v.valid()
 }
 
+func (c *CloudProvider) validate() (bool, []error) {
+	v := newValidator()
+	if c.Provider != "" {
+		if !util.Contains(c.Provider, cloudProviders()) {
+			v.addError(fmt.Errorf("%q is not a valid cloud provider. Optins are %v", c.Provider, cloudProviders()))
```
alexbrand: typo: Optins

ansible/roles/kube-apiserver/templates/kube-apiserver.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   labels:
     tier: control-plane
     component: kube-apiserver
+    kismatic/nodeIP: {{ internal_ipv4 }}
```
alexbrand: Was there a reason for choosing the IP here instead of the node name?

dkoshkin:
Currently in kubernetes the `hostname-override` flag is not always respected when using the `cloud-provider` flag. During the upgrade process we rely on knowing the node name from the plan file `host` flag. 

Because this is no longer guaranteed, we will need to document this behavior and recommend to users to use the fqdn in the plan file.

ansible/roles/worker-smoke-test/tasks/main.yaml
```diff
@@ -1,3 +1,16 @@
...
-    command: kubectl get nodes {{ worker_node|lower }} --kubeconfig {{ kubernetes_kubeconfig_path }}
+  - name: wait for node '{{ worker_node|lower }}' has registered with API server and is Ready
+    command: kubectl get nodes --selector kismatic/host={{ worker_node }} --kubeconfig {{ kubernetes_kubeconfig_path }}
+    register: nodeStatus
+    until: nodeStatus|success and "Ready" in nodeStatus.stdout
+    retries: 20
+    delay: 6
+
+
+  - name: fail if node '{{ worker_node|lower }}' is not Ready
```
alexbrand: I think we need to handle to possible failure scenarios here:
1. Kubectl fails for some reason, and stderr has a meaningful message
2. The node never enters the ready state, and has some other state. In this case, stderr will not have anything. Instead, it might be worth doing a kubectl describe node or something to get more data.

integration/cloud_integration.go
```diff
@@ -0,0 +1,27 @@
+package integration
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/apprenda/kismatic/pkg/retry"
+)
+
+func testLoadBalancer(node NodeDeets, sshKey string) error {
```
alexbrand: Can we include AWS in the name of this func somewhere? This is an AWS specific func.

Also seems like it tests more than just the load balancer. Maybe testAWSCloudProvider?

integration/install_test.go
```diff
@@ -349,3 +373,23 @@ var _ = Describe("kismatic", func() {
...
+
+	By("installing the cluster")
+	err := installKismatic(nodes, installOpts, sshKey)
+	Expect(err).ToNot(HaveOccurred())
+
+	By("test the LoadBalancer")
+	err = testLoadBalancer(nodes.master[0], sshKey)
+	Expect(err).ToNot(HaveOccurred())
+
+	By("adding a worker node")
```
alexbrand: What's the reason for adding a worker here? Doesn't seem like we are asserting anything cloud-provider related after adding it.

alexbrand:
Needs rebase from master.

ansible/roles/worker-smoke-test/tasks/main.yaml
```diff
@@ -1,3 +1,21 @@
...
+      msg: |
+        An error occurred trying to get the status of the node
+        
+        {{ nodeStatus.stderr }}
+    when: nodeStatus|failure
+
+  - name: fail if node '{{ worker_node|lower }}' is not Ready
+    fail:
+      msg: |
+        Node is not in the Ready state
```
alexbrand: Getting the state that the node is in would be helpful for debugging/troubleshooting.

ansible/roles/worker-smoke-test/tasks/main.yaml
```diff
@@ -1,3 +1,21 @@
 ---
-  - name: verify {{ worker_node|lower }} registered with API server
-    command: kubectl get nodes {{ worker_node|lower }} --kubeconfig {{ kubernetes_kubeconfig_path }}
+  - name: wait for node '{{ worker_node|lower }}' has registered with API server and is Ready
```
alexbrand: This sentence sounds a little weird to me. How about:
 >wait for node 'x' to register with the API server and become Ready

ansible/roles/kube-apiserver/templates/kube-apiserver.yaml
dkoshkin: Oops looks like your comment got lost with a rebase, when testing on AWS I ran into an issue of the `kube-apiserver`  not starting up correctly on CentOS and RHEL, I found both [kops](https://github.com/kubernetes/kops/issues/668) and [kubespray](https://github.com/kubernetes-incubator/kubespray/issues/1143) ran into a similar issues.

This happens because symlinked files do not get properly mounted in docker and the kube-apiserver cannot connect to the AWS API because of the missing CA.

ansible/roles/worker-smoke-test/tasks/main.yaml
dkoshkin: How does `wait for node 'x' to register with the API server and become Ready`sound?

ansible/roles/worker-smoke-test/tasks/main.yaml
dkoshkin: Getting just the state is complicated, instead I am printing the results of `kubectl get nodes --selector kismatic/host={{ worker_node }}` 

Will look something like:
```
Node is not in the Ready state

NAME        STATUS                     AGE       VERSION
master001   Ready,SchedulingDisabled   12m       v1.7.4
```

alexbrand: Even better!

alexbrand:
LGTM

PR #764, opened 2017-08-25
Use the --skip-refresh helm flag when doing a disconnected install

### LINKED ISSUE #763: Helm init may fail during disconnected installs
BUG

**What happened**:
Task `helm init` fails when trying to add the default repo `https://kubernetes-charts.storage.googleapis.com/index.yaml`

On disconnected installs `helm init` should use the `--skip-refresh` flag

Run a disconnected install with a bastion node that is also disconnected and does not have the `helm` cache locally.

dkoshkin:
Fixes #763

### DIFF
--- a/ansible/_helm.yaml
+++ b/ansible/_helm.yaml
@@ -14,14 +14,16 @@
         command: "kubectl create clusterrolebinding tiller --clusterrole cluster-admin --serviceaccount=kube-system:tiller"
         register: out
         failed_when: 'out.rc != 0 and out.stderr is defined and "Error from server (AlreadyExists)" not in out.stderr'
+        
       - name: "run helm init"
-        local_action: command ../../helm init --service-account=tiller -i="{{ helm_img }}" --upgrade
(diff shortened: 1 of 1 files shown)


alexbrand:
Does this mean the repository won't be set up locally if you are doing a disconnected install? Makes sense to me, but want to understand. I tried looking at the helm docs but it's not entirely clear to me.

```
      --skip-refresh             do not refresh (download) the local repository cache
```

dkoshkin:
Just looked into this, it sets it up as a repo, so `help repo list` will still show it, but it doesn't try to fetch the `~/helm/repository/cache/stable-index.yaml` file that contains the list of all charts in the repo.

alexbrand:
This LGTM. If there's no internet access, there is no way for us to fetch the repo information. 

I think @mssch had some thoughts on this though.

mssch:
No specific thoughts, just can confirm #763 since I ran into it during a disconnected install, this should fix it indeed.
LGTM

PR #773, opened 2017-09-07
Bump glusterfs version and respect version when installing on Ubuntu

### LINKED ISSUE #772: Glusterfs version set in ansible is not being respected during installation
The glusterfs version is not being honored when installing gluster on Ubuntu.

alexbrand:
Fix #772


alexbrand:
In terms of upgrade concerns:
1. We currently don't handle upgrades of glusterfs
2. This is a bump of the patch version, so we should be OK if we ever decide to implement glusterfs upgrades.

dkoshkin:
LGTM

PR #775, opened 2017-09-08
Remove deprecated fields from plan file.

alexbrand:
```
Remove deprecated fields, and code for handling/testing deprecated
fields.

Also made the plan file unmarshalling fail if any unknown field is
provided in the YAML file. This will prevent mis-configuration and also
let the user know when their plan file is outdated.
```


alexbrand:
I am going to close this. We can tackle it in the future when we have a way to migrate old plan files to new ones. I don't think we should remove the fields until we have that migration path.

PR #785, opened 2017-09-14
Disable preflight check for rpcbind.socket port 111

### LINKED ISSUE #784: Preflight Port 111 check fails
BUG

**What happened**:
A preflight check fails with:
```
=> The following checks failed on "ip-10-0-3-16.ec2.internal":
   - Port Available: 111
   - Port Accessible: 111: error reading from TCP socket: read tcp 172.21.0.3:37744->52.202.91.215:111: read: connection reset by peer
```
The preflight to pass

This failure can be seen when running the `Running validation with package installation disabled using a Minikube RedHat 7 layout` integration test

We have also seen this failure on certain networks when port `111` is blocked

Port 111 is a low, reserved port for `rpc`, for now we can safely assume that if this port is already being used, it is by the `rpcbind.socket` service.

**Environment**:
```
Kismatic:
  Version: 1.5.3-16-g13234654-dirty
  Built: Wed Sep 13 23:12:55 UTC 2017
  Go Version: go1.8
```
CircleCI build
RHEL 7.3

dkoshkin:
Fixes #784 

Disable the checks for port 111 in the preflight, until we have a better method of checking if ports are used by the expected service.

### DIFF
--- a/pkg/inspector/rule/rule_set.go
+++ b/pkg/inspector/rule/rule_set.go
@@ -380,13 +380,14 @@ const defaultRuleSet = `---
   timeout: 5s
 
 # Ports required for NFS
-- kind: TCPPortAvailable
-  when: ["storage"]
-  port: 111
-- kind: TCPPortAccessible
-  when: ["storage"]
-  port: 111
-  timeout: 5s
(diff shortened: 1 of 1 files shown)


alexbrand:
LGTM!

PR #786, opened 2017-09-14
Feature packages binaries

dkoshkin:


alexbrand:
This is a feature branch that got a bunch of _reviewed_ work merged into it.

PR #787, opened 2017-09-15
Support docker registry with username and password

### LINKED ISSUE #783: Support Docker registry authentication
FEATURE

**What happened**:
Currently KET does not support pulling the required installation images from a private Docker registry that requires authentication. 
Because KET installs the docker daemon, the user does not have an opportunity to run `docker login` or configure an authentication method in any other way.

Being able to use Docker registries with authentication.

**Environment**:
```
Kismatic:
  Version: 1.5.3
```

dkoshkin:
Fixes #783


docs/plan-file-reference.md
```diff
@@ -426,6 +428,26 @@
 | **Required** |  No |
 | **Default** | ` ` | 
 
+###  docker_registry.username
+
+ If the registry requires authentication provide the username and password below. Otherwise leave blank for unauthenticated access.
```
alexbrand: How about:
> The username that should be used when connecting to a registry that has authentication enabled.

docs/plan-file-reference.md
```diff
@@ -426,6 +428,26 @@
...
+
+| | |
+|----------|-----------------|
+| **Kind** |  string |
+| **Required** |  No |
+| **Default** | ` ` | 
+
+###  docker_registry.password
+
+ Password for the above username.
```
alexbrand: How about:

> The password that should be used when connecting to a registry that has authentication enabled.

alexbrand:
LGTM

PR #789, opened 2017-09-18
Expose calico log_level in the plan file

### LINKED ISSUE #779: Make calico log level parameterized
**Is this a BUG REPORT or FEATURE REQUEST?**: FEATURE REQUEST

**What happened**:
Calico info logs spam a lot of 'iptables' info filling up the disk. We would like to keep the log level to WARNING. [Currently, it is hardcoded to INFO in ansible roles](https://github.com/apprenda/kismatic/blob/18987af8e3c6844b3b3ed2fe38bc3532f6e6eb37/ansible/roles/calico/templates/calico.yaml#L22). Solution would be to make it parameterized and pick up from config yaml (if not provided, only then keep it as INFO)

Parameterized way to provided calico log level

Grep the calico logs. They all should say INFO.

dkoshkin:
Fixes #779 

Calico [doc](https://docs.projectcalico.org/v2.5/reference/cni-plugin/configuration)

### DIFF
--- a/ansible/roles/calico/templates/calico.yaml
+++ b/ansible/roles/calico/templates/calico.yaml
@@ -19,7 +19,7 @@ data:
         "etcd_key_file": "{{ kubernetes_certificates.etcd_client_key }}",
         "etcd_cert_file": "{{ kubernetes_certificates.etcd_client }}",
         "etcd_ca_cert_file": "{{ kubernetes_certificates.ca }}",
-        "log_level": "info",
+        "log_level": "{{ cni.options.calico.log_level }}",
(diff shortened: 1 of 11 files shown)


alexbrand:
LGTM!

alexbrand:
Now that I think about it more... What happens during an upgrade?

dkoshkin:
It defaults to `info` even if the field is missing in the plan file. If the user changes during upgrade, it would be set and restarted.

alexbrand:
Sweet!

prat0318:
Awesome.

PR #793, opened 2017-09-20
Add node labels option in the plan file to label nodes

### LINKED ISSUE #133: Automate the labelling of nodes
Would be useful if users had the possibility of specifying labels under each node in the YAML file.

Then KET would execute `kubectl label nodes <node-name> <label-key>=<label-value>` for them during the provisioning process.

dkoshkin:
Fixes #133 

Example, either format will work as the are both `map[string]string`
```
worker:
  expected_count: 1
  nodes:
  - host: worker001
    ip: 192.168.42.4
    labels: {"com.koshkin/worker" : "", "com.koshkin/foo": "bar"}
  - host: worker002
    ip: 192.168.42.5
    labels:
      com.koshkin/foo: bar
      com.koshkin/ssd: true 
```

### DIFF
--- a/integration/labels.go
+++ b/integration/labels.go
@@ -1,0 +1,40 @@
+package integration
+
+import (
+	"fmt"
+	"time"
+)
+
+func containsLabels(nodes provisionedNodes, sshKey string) error {
+	// labels were hardcoded in th plan pattern
+	tests := []struct {
+		nodes []NodeDeets
+		label string
+	}{
+		{
+			nodes: nodes.master,
+			label: "com.integrationtest/master:true",
+		},
+		{
+			nodes: nodes.worker,
+			label: "com.integrationtest/worker:true",
+		},
+		{
+			nodes: nodes.ingress,
+			label: "com.integrationtest/ingress:true",
+		},
+		{
+			nodes: nodes.storage,
+			label: "com.integrationtest/storage:true",
+		},
+	}
+	for _, role := range tests {
+		for _, n := range role.nodes {
+			if err := runViaSSH([]string{fmt.Sprintf("sudo kubectl get nodes %s -o jsonpath='{.metadata.labels}' | grep %q", n.Hostname, role.label)}, []NodeDeets{nodes.master[0]}, sshKey, 1*time.Minute); err != nil {
+				return fmt.Errorf("error validating node %q label: %v", n.Hostname, err)
+			}
+		}
+	}
+
+	return nil
+}
--- a/pkg/install/execute.go
+++ b/pkg/install/execute.go
@@ -787,6 +787,18 @@ func (ae *ansibleExecutor) buildClusterCatalog(p *Plan) (*ansible.ClusterCatalog
 		}
 	}
 
+	// merge node labels
+	// cannot use inventory file because nodes share roles
+	// set it to a map[host][]key=value
+	cc.NodeLabels = make(map[string][]string)
+	for _, n := range p.getAllNodes() {
+		if val, ok := cc.NodeLabels[n.Host]; ok {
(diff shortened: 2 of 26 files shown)


pkg/install/plan_types.go
```diff
@@ -411,6 +411,11 @@ type Node struct {
 	// The internal (or private) IP address of the node.
 	// If set, this IP will be used when configuring cluster components.
 	InternalIP string
+	// Labels to add when registering the node in the cluster.
+	// Labels must be key=value pairs separated by ','.
```
alexbrand: This is no longer the case, right? Given that we are using a map[string]string

alexbrand:
We should have test coverage in our integration suite. My inclination is to add a new subtest to the large clusters we currently deploy, instead of adding a new cluster deployment for testing this.

swade1987:
@alexbrand @dkoshkin was looking at the tests for this and I don't seem to see any negative tests such as if the user enters the following what happens:

- `labels: {"com.koshkin/worker" : ""}`
- `labels: {"" : "com.koshkin/worker"}`
- `labels: {"" : ""}`
- `labels: {"node-type:test" : ":test"}`

integration/labels.go
```diff
@@ -0,0 +1,22 @@
+package integration
+
+import (
+	"fmt"
+	"time"
+)
+
+func containsLabels(nodes provisionedNodes, sshKey string) error {
```
alexbrand: Curious about if there was any reason for not verifying ingress and storage nodes as well.

pkg/install/execute.go
```diff
@@ -807,6 +807,18 @@ func (ae *ansibleExecutor) buildClusterCatalog(p *Plan) (*ansible.ClusterCatalog
 		}
 	}
 
+	// merge node labels
+	// cannot use inventory file because nodes share roles
+	// set it to a map[host][]key=value
+	cc.NodeLabels = make(map[string][]string)
+	for _, n := range p.getAllNodes() {
```
alexbrand: Is this merging of labels deterministic? If so, in what order will they be merged? If not, we should probably call it out in the reference doc as a potential gotcha.

pkg/install/plan_types.go
```diff
@@ -458,14 +462,18 @@ type SSHConnection struct {
 // GetUniqueNodes returns a list of the unique nodes that are listed in the plan file.
 // That is, if a node has multiple roles, it will only appear once in the list.
 func (p *Plan) GetUniqueNodes() []Node {
-	seenNodes := map[Node]bool{}
+	seenNodes := map[string]bool{}
 	nodes := []Node{}
 	for _, node := range p.getAllNodes() {
-		if seenNodes[node] {
+		// Cannot use the Node struct directly as it contains a map
```
alexbrand: I think I am confused by this. Do you mind elaborating on why we can't use the `Host` as the identity of the node?

pkg/install/validate.go
```diff
@@ -422,6 +424,33 @@ func (n *Node) validate() (bool, []error) {
...
+		errs := validation.IsQualifiedName(key)
+		for _, err := range errs {
+			v.addError(fmt.Errorf("Node label name %q is not valid %s", key, err))
+		}
+		errs = validation.IsValidLabelValue(val)
+		for _, err := range errs {
+			v.addError(fmt.Errorf("Node label %q is not valid %s", val, err))
+		}
+
+		// // regex is used by kubernetes validation
```
alexbrand: Remove?

pkg/install/execute.go
dkoshkin: Added a line in the reference doc:
```
+	// If a label is repeated for the same node under different roles, the last label in the plan file will be used.
```

alexbrand:
Nice, LGTM!

PR #794, opened 2017-09-20
Validate all nodes to be unique

### LINKED ISSUE #792: Validate unique nodes across plan file, not just nodeGroup
BUG

**What happened**:
2 nodes with the same `host` but different `IPs` were used as master and etcd. Validation did not fail and the installation failed later when trying to verify `etcd` health. 

Validation should fail as the `hostname` was not unique to one IP.

Use the same `host` for 2 different nodes for 2 different roles.

We currently check for uniqueness, but only within a NodeGroup not across the whole plan file. 

**Environment**:
- Kernel (e.g. `uname -a`):

dkoshkin:
Fixes #792


alexbrand:
Nice, good catch!

PR #796, opened 2017-09-20
Deploy docker certificates in the docker play

alexbrand:
The docker certificates are only consumed by docker. For this reason, it seems to me that the best place for deploying them is when installing docker.


dkoshkin:
LGTM!

PR #797, opened 2017-09-21
New kismatic volume delete command

### LINKED ISSUE #542: Kismatic: function to delete volumes
there are an add and a list option, but no delete option for volumes

I suggest implementing a `kismatic volume delete` command, which has to

* delete pv with given name
* run `gluster volume stop NAME --mode=script && gluster volume delete NAME --mode=script` to delete the gluster volume

dkoshkin:
Fixes #542

### DIFF
--- a/integration/storage.go
+++ b/integration/storage.go
@@ -116,6 +116,13 @@ func createVolume(planFile *os.File, name string, replicationCount int, distribu
 	return cmd.Run()
 }
 
+func deleteVolume(planFile *os.File, name string) error {
+	cmd := exec.Command("./kismatic", "volume", "delete", "-f", planFile.Name(), name, "--force")
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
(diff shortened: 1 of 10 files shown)


integration/storage.go
```diff
@@ -238,5 +245,12 @@ func testStatefulWorkload(nodes provisionedNodes, sshKey string) error {
 	if err != nil {
 		return fmt.Errorf("Reader workload failed: %v", err)
 	}
+
+	By("Deleting the storage volume")
+	err = deleteVolume(plan, "kis-int-test")
+	if err != nil {
+		return fmt.Errorf("Failed to delete volume: %v", err)
+	}
+
```
alexbrand: We are running the command here, but not making any assertions after the fact. Seems like we should verify that the volume was indeed deleted?

dkoshkin: Added an assertion after the removing.

alexbrand:
This LGTM! Needs cli docs regen.

PR #801, opened 2017-09-21
Remove custom repo URLs feature

### LINKED ISSUE #800: Remove custom repo URLs feature
While revamping the disconnected installation (#725), we have decided to remove the custom repo URLs feature from the plan file. 

The current implementation makes assumptions that don't always hold true now that we are not using our own packages. Furthermore, the vast number of options that can be set when configuring repositories (autN/authZ, proxy, gpg validation, gpg public keys, etc) makes it difficult to support this properly.

Moving forward, the user will be responsible for configuring their nodes when they want to use a package repository other than the official ones.

alexbrand:
Fixes #800


dkoshkin:
LGTM, can't see anything else.

PR #757, opened 2017-08-23
Proxy fix

dkoshkin:
Fixes #744, fixes #745, fixes #746, fixes #187, fixes #762 

- Move `http_proxy` settings under the tasks that use them
- Run preflight checks from the cluster, not the install node
- Verify connectivity to helm repos during preflight
- Fix disconnected preflight check and run `--preflight` during tests

PR #808, opened 2017-09-25
Run upgrade preflight checks when performing upgrade

### LINKED ISSUE #807: Inspector is running installation checks during upgrade
When the execution of the inspector checks were moved into the cluster as part of #757, the upgrade play lost the `--upgrade` flag which is necessary when performing upgrade preflight checks.

alexbrand:
Fixes #807


dkoshkin:
LGTM. Good catch, this got broken in https://github.com/apprenda/kismatic/pull/757, so it was not yet released.

PR #814, opened 2017-09-26
Disconnected install

### LINKED ISSUE #725: Revamp the disconnected installation technique
As we move from custom built distribution packages to upstream packages, we will have to revamp our disconnected installation documentation / technique. The main trigger for this work is that we will most likely lose the "offline" package, which means KET will either a) stop being responsible for seeding the container registry, or b) change the seeding process.

**Container Images**
Components that are distributed as container images will need to be available in an internal container image registry. Potential solutions here, from simpler to more complex:

1. Maintain a list of container images that are required during an installation. The user is responsible for seeding their internal registry with all images.
2. Maintain a script that must be run by the user prior to the installation. This script will contain a list of the container images that are required for an installation, and will download, tag, and push each one to the internal registry.
3. Make KET compute the required container images according to the selected options, add-ons, etc. KET can then seed the registry from the installation machine. This assumes that the installation machine has internet access and a docker client that can be used to download, tag and push images.

**OS Packages**
If we are unable to shed the need for RPM/DEB packages, we can continue to accept a list of private package repositories. The user is responsible for seeding the private repository with the required packages. Some potential improvements to the current status quo:

1. Maintain a script for each supported linux distro that builds a package repository mirror.
2. Maintain a container image that has a pre-built repository mirror. This container can then be started by the user internally.

alexbrand:
Fixes #725 

This PR introduces the new method for performing disconnected installations with KET. The initial commit in this PR includes documentation on this new feature. I have attempted to maintain separate commits that tell a concise story.

Moving forward, the KET installation or upgrade process is no longer responsible for seeding the local registry with the required images. Instead, the user is responsible for seeding their internal registry. In order to assist the user with this task, we introduce a new command `./kismatic seed-registry`. We also remove the `disable_registry_seeding` option from the plan file, as it no longer has an effect on the installation/upgrade.

In order to implement this command, we needed a centralized manifest that lists all the images required for a KET installation. In previous versions, this manifest was engrained in the ansible variables file. In this PR, a new file (and data structure) is proposed to house the list of images and their respective versions. 

Finally, this PR updates all the integration tests that have been waiting for this feature (the disconnected install tests, and the disconnected upgrade tests).


ansible/group_vars/all.yaml
```diff
@@ -1,35 +1,15 @@
...
-pause_version: "3.0"
-apprenda_tcp_healthz_version: "v1.0.0"
-helm_version: v2.5.1
-heapster_version: "v1.4.0"
-influxdb_version: "v1.1.1"
 glusterfs_server_version_rhel: "3.8.15-2.el7"
 glusterfs_server_version_ubuntu: "3.8.15-ubuntu1~xenial1"
+
+# TODO: The docker registry is going away...remove this var
+docker_registry_version: "2.6.2"
```
dkoshkin: Can this be removed, I dont see it get used anywhere.

docs/disconnected_install.md
```diff
@@ -1,20 +1,236 @@
...
+
+### Start Apache server
+We will use the Apache HTTP server for exposing the repository over HTTP.
+```
+systemctl enable httpd
+systemctl start httpd
+```
+
+### Configure nodes
+With this approach, we created five mirrors that must be configured on the nodes:
```
dkoshkin: It a little unclear that this must be repeated for all repos, also what should be in `$1` in `baseurl=$1/base`?

docs/kismatic-cli/kismatic_seed-registry.md
```diff
@@ -0,0 +1,42 @@
...
+kismatic seed-registry [flags]
+```
+
+### Options
+
+```
+  -h, --help               help for seed-registry
+      --list-only          when true, the images will only be listed but not pushed to the registry
+  -f, --plan-file string   path to the installation plan file (default "kismatic-cluster.yaml")
+      --server string      set to the location of the registry server (e.g. localhost:5000)
```
dkoshkin: Should we call out the `server` cannot contain `http://` or `https://`?

pkg/cli/seed.go
```diff
@@ -0,0 +1,184 @@
...
+	// Figure out the registry we are to seed
+	// The registry specified through the command-line flag takes precedence
+	// over the one defined in the plan file.
+	server := options.registryServer
+	if server == "" {
+		// we need to get the server from the plan file
+		planner := install.FilePlanner{File: options.planFile}
+		if !planner.PlanExists() {
+			util.PrettyPrintErr(stdout, "Reading installation plan file [ERROR]")
+			fmt.Fprintln(stdout, "Run \"kismatic install plan\" to generate it")
```
dkoshkin: How about
```
fmt.Fprintln(stdout, "Run \"kismatic install plan\" to generate it or use the \"--server\" option")
```

pkg/cli/seed.go
```diff
@@ -0,0 +1,184 @@
...
+			errs = append(errs, errors.New("The private registry's port must be set in the plan file."))
+		}
+		if plan.DockerRegistry.Port < 1 || plan.DockerRegistry.Port > 65535 {
+			errs = append(errs, fmt.Errorf("The docker registry port '%d' provided in the plan file is not valid.", plan.DockerRegistry.Port))
+		}
+		server = fmt.Sprintf("%s:%d", plan.DockerRegistry.Address, plan.DockerRegistry.Port)
+	}
+
+	// Read the list of images
+	varsBytes, err := ioutil.ReadFile(imageManifestFile)
```
dkoshkin: Can this be extracted from `doListImages` into a common func.

dkoshkin:
In the `rule_set.go` `kismatic-offline` package rules should be removed.

ansible/group_vars/all.yaml
alexbrand: Yes - thanks!

docs/kismatic-cli/kismatic_seed-registry.md
alexbrand: Good call

pkg/cli/seed.go
alexbrand: Ah yes! Thanks

pkg/cli/seed.go
alexbrand: Yep, sounds good.

dkoshkin:
Looks like the empty address is not being validated.
```
./kismatic install plan
./kismatic seed-registry

Seed Container Image Registry=======================================================
Error parsing reference: ":8443/gcr.io/google_containers/heapster-influxdb-amd64:v1.1.1" is not a valid repository/tag: invalid reference format
Error seeding image "gcr.io/google_containers/heapster-influxdb-amd64:v1.1.1": exit status 1
```

alexbrand:
Thanks! Fixed

dkoshkin:
LGTM!

PR #830, opened 2017-09-28
Allow package installation in upgrade integ test

### LINKED ISSUE #829: Integration test: The online upgrade (disconnected from internet) test is disabling package installation
In the usual case of a disconnected install/upgrade, the installation of packages is handled by KET. 

The current online/disconnected upgrade is broken as it's disabling package installation.

alexbrand:
```
The test is setup in a way that KET is responsible for installing the
packages. However, the plan file was not updated from the previous
implementation of the test that was performing the installation of
packages out of band from KET upgrade.
```

Fixes #829


dkoshkin:
LGTM!

PR #831, opened 2017-09-29
Remove upgrade-specific preflight role

### LINKED ISSUE #826: Upgrade preflight checks do not run the direct-lvm checks
When upgrading to 1.6.0, users will have to provide a block storage device on etcd nodes when using the direct-lvm docker storage option.

During upgrade, we must run the direct-lvm checks that we usually run during installation.

Furthermore, we should look into having a single preflight ansible play, and conditionally enabling/disabling specific preflights when performing an upgrade.

alexbrand:
```
Instead of having two separate preflight roles, we will have a single
one that will conditionally disable/enable checks depending on whether
performing an install or an upgrade.
```

Fixes #826


dkoshkin:
Great LGTM!

PR #833, opened 2017-09-29
Upgrade calico

### LINKED ISSUE #755: Upgrade Calico
Calico `v2.5` has been [released](https://docs.projectcalico.org/v2.5/releases/). Need to upgrade to the latest version at the time of KET v1.6.0 release.

The new version of Calico uses the new networking.k8s.io/NetworkPolicy API semantics.

dkoshkin:
Fixes #755 

- Using new API for Calico instead of ns annotations
- Update the deployment to use a dedicated serviceaccount, update the rest to to match 
- Use new `calico-kube-controllers`


alexbrand:
The only concern I had around this was that I wasn't sure if we had to do anything to upgrade Calico, but @dkoshkin mentioned that he reviewed the Calico release notes and there was nothing that would affect us.

LGTM

PR #835, opened 2017-09-29
Properly fail on kube-dns validation error

### LINKED ISSUE #827: If kubedns is not up and running, the installation continues
There seems to be a regression in our kubedns readiness validation, as I just ran into a case where kubedns did not come up properly, but the installation continued.

dkoshkin:
Fix #827 

It is possible that the `stdout` is empty when getting the `Running` pods or the list of broken pods. Catch these errors.

### DIFF
--- a/ansible/roles/kube-dns/tasks/main.yaml
+++ b/ansible/roles/kube-dns/tasks/main.yaml
@@ -23,8 +23,14 @@
     - name: fail if DNS pod validation command returned an error
       fail:
         msg: |
-          Attempted to validate the DNS pods, but got an error from kubectl: {{ readyReplicas.stderr }}
+          Attempted to validate the DNS pods, but got an error: {{ readyReplicas.stderr }}
       when: readyReplicas.stderr != ""
+
+    - name: fail if DNS pod validation command could not determine if the DNS pods are ready
+      fail:
+        msg: |
+          Waited for all DNS pods to be ready, but they took longer than 5 minutes to be in the ready state.
+      when: readyReplicas.stdout == ""
 
     - name: find the DNS pods that failed to start
       # Get the name and status/phase for all kubedns pods, and then filter out the ones that are not running.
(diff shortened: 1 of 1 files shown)


alexbrand:
I have a feeling that we are doing something wrong when getting the number of ready replicas from the deployment. Does kubectl really return an empty string when the number of ready replicas is zero?

dkoshkin:
I just tested and it looks like it does return an empty response when `availableReplicas` is 0.

alexbrand:
Cool. My guess is that the readyReplicas field is not serialized to json when it's value is the default value (`0`), which in turn results in the empty string because the field itself is not there and we are querying with jsonpath.

ansible/roles/kube-dns/tasks/main.yaml
```diff
@@ -23,9 +23,15 @@
...
       fail:
         msg: |
-          Attempted to validate the DNS pods, but got an error from kubectl: {{ readyReplicas.stderr }}
+          Attempted to validate the DNS pods, but got an error: {{ readyReplicas.stderr }}
       when: readyReplicas.stderr != ""
 
+    - name: fail if DNS pod validation command could not determine if the DNS pods are ready
+      fail:
+        msg: |
+          Attempted to validate the DNS pods, did not get a response: {{ readyReplicas.stderr }}
```
alexbrand: readyReplicas.stderr will always be empty here, right? Otherwise it would get caught in the task before this one.

This should probably say something like:
```
Waited for all DNS pods to be ready, but they took longer than 5 minutes to be in the ready state.
```

Which makes me wonder we the final task in this play is not catching this scenario.

alexbrand:
Looking at this again... By adding what we are adding, wouldn't we lose out on getting the logs of one of the pods that is failing in the subsequent tasks? I guess I wonder why those logs are not getting printed in this scenario.

dkoshkin:
`find the DNS pods that failed to start` would return an empty response so we dont have a pod name to get the logs from.

alexbrand:
Gotcha. Something still feels weird to me but I am not sure what it is. I say we merge this as it fixes the problem.

LGTM

Issue #767, opened 2017-08-30
Set kubernetes.io/cluster label when using aws cloud-provider


dkoshkin:
**Is this a BUG REPORT or FEATURE REQUEST?**:
Upstream Change

**What happened**:
In Kubernetes > v1.8, there is going to be a [new requirement](https://github.com/kubernetes/kubernetes/issues/48954) to set a label for `kubernetes.io/cluster` with a cluster ID, when using the aws cloud-provider

**What you expected to happen**:
The cluster to install successfully without needing a new label.

**How to reproduce it (as minimally and precisely as possible)**:
Install a new cluster with kubernetes v1.8 and aws cloud-provider.

**Anything else we need to know?**:
There will be a [new flag](https://github.com/kubernetes/kubernetes/issues/48954) on the kubernetes components to still function as it has before, but it will be marked deprecated immediatly.

PR #836, opened 2017-09-29
Documentation for the KET 1.6.0 release [skip ci]

alexbrand:
Documentation for the KET 1.6.0 release
- [x] Cloud provider documentation
- [x] Overriding component configuration
- [x] Installing with HTTP proxies
- [x] Seed registry documentation
- [x] Upgrade to 1.6.0 documentation
  - [x] Etcd now requires block device when using direct-lvm
  - [x] Disconnected upgrade documentation (you must seed registry before, etc...)
  - [x] Swap memory and Kubelet


docs/cloud_provider.md
```diff
@@ -0,0 +1,84 @@
...
+                "ecr:GetRepositoryPolicy",
+                "ecr:DescribeRepositories",
+                "ecr:ListImages",
+                "ecr:BatchGetImage"
+            ],
+            "Resource": "*"
+        }
+    ]
+}
+```
```
dkoshkin: Related to https://github.com/apprenda/kismatic/issues/767, it would be good to inform users of this new requirement.

--- 

Starting with Kubernetes v1.8.0, machines deployed on AWS with `cloud-provider` enabled must be labeled with the `ClusterID`( a cluster identifier unique to the AWS Availability Zone) by using the `kubernetes.io/cluster/$CLUSTERID` tag on the EC2 instances. KET will set the `--allow-untagged-cloud` flag on the kubc-controller-manager, however this field is marked deprecated by Kubernetes and will be removed in the future release.

If you choose not to tag your instances, do not deploy multiple clusters in a single AWS AZ as automatically created resources such as volumes and load balancers may be incorrectly assigned to the wrong cluster.

More info can be found in the [upstream issue](https://github.com/kubernetes/kubernetes/pull/49215).

alexbrand: Does this need to match the cluster name provided in the plan file?

dkoshkin: Checking now.

dkoshkin: No it does not, it just needs to be the same for all the nodes in the cluster.

docs/cloud_provider.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Prerequisites
+The following requirements must be completed before installing a Kubernetes cluster
+with the AWS cloud provider integration enabled:
+
+* The hostnames set in the plan file must match the Private DNS name that is 
+assigned to the machine on AWS.
+* All machines that are part of the Kubernetes cluster must be tagged with a 
+unique cluster name. More specifically, all machines in the cluster must have
+the `kubernetes.io/cluster` tag set to the same value.
```
dkoshkin: Should be `kubernetes.io/cluster/$UNIQUE_ID`

alexbrand: Oh really? Weird. What should be the value of the label then?

dkoshkin: Not sure if it matters, our tests just use an empty string.

alexbrand:
Updated the tag on the AWS cloud provider prereqs list.

dkoshkin:
LGTM!

alexbrand:
Going to rebase and merge.

PR #841, opened 2017-10-02
Improve plan file layout and update/remove old comments

alexbrand:
There were some comments that were outdated. I updated them and also removed some comments that seemed unnecessary. Also made some cosmetic changes to improve the file's layout and spacing.

### DIFF
--- a/pkg/install/test/plan-template-with-storage.golden.yaml
+++ b/pkg/install/test/plan-template-with-storage.golden.yaml
@@ -5,8 +5,7 @@ cluster:
   # used for administration without a security certificate.
   admin_password: password
 
-  # When true, installation will not occur if any node is missing the correct
-  # deb/rpm packages.
+  # Set to true if the nodes have the required packages installed.
   disable_package_installation: false
 
   # Set to true if you are performing a disconnected installation.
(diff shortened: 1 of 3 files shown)


dkoshkin:
What would this look like if we added a newline after each section.

I thin adding a new line [here](https://github.com/apprenda/kismatic/blob/plan-file-comments/pkg/install/plan.go#L174) and removing `\n` from [here](https://github.com/apprenda/kismatic/blob/plan-file-comments/pkg/install/plan.go#L205)

alexbrand:
Added a change to include a new line before a new "section".

pkg/install/test/plan-template-with-storage.golden.yaml
```diff
@@ -6,12 +6,13 @@ cluster:
   admin_password: password
 
   # When true, installation will not occur if any node is missing the correct
```
dkoshkin: To be consistent should we set this to something like
> Set to true if nodes have the required packages installed, installation will not occur if any node is missing the correct deb/rpm packages.

dkoshkin: Also any other `When true` to be consistent with `Set to...`

dkoshkin: Should we remove `you/your` and use `the` or `cluster` etc.

alexbrand: I was using you based on this style guide: https://kubernetes.io/docs/home/contribute/style-guide/#address-the-reader-as-you

Was there any specific line that you think would benefit from not using `you`?

dkoshkin: Actually I didnt realize that was in the style-guide, ignore that.

alexbrand:
Updated incosistent "When true, ..." comments.

dkoshkin:
LGTM

PR #843, opened 2017-10-02
Upgrade kuberntes to v1.8.0

### LINKED ISSUE #602: Enable node authorizer and admission plugin introduced (beta) in Kubernetes v1.7.0
The node authorizer limits node's access to the API server by only giving access to objects owned by the node (node object), scheduled to the node (pods), and owned by pods scheduled to the node (secrets, configmaps, etc).

### LINKED ISSUE #767: Set kubernetes.io/cluster label when using aws cloud-provider

### LINKED ISSUE #782: Upgrade to Kubernetes v1.8
FEATURE

**What happened**:
Kubernetes v1.8 is going to be released in a few weeks.
Will use this issue to track any work that needs to be done for the new version.

TODO:
- [ ] [Set kubernetes.io/cluster label when using aws cloud-provider](#767) 
- [ ] [Update storage objects before upgrading to 1.8](#621)

dkoshkin:
Fixes #782, fixes #767, fixes #602 

Set a tag `kubernetes.io/cluster/$UNIQUE_ID` on the EC2 machines created for integration tests.


alexbrand:
Exciting! LGTM

PR #844, opened 2017-10-02
Verify memory swap is disabled

### LINKED ISSUE #842: Add preflight check that verifies swap is disabled if the kubelet option is not overriden
Starting with Kubernetes 1.8.0, the kubelet will fail to start if swap is enabled, and the right kubelet flag is not overriden.

If possible, we should fail early if swap is enabled on the node and the user did not override the kubelet flag.

dkoshkin:
Fixes #842.

Mimic the validation [performed by Kubernetes](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/container_manager_linux.go#L200)


alexbrand:
LGTM

PR #845, opened 2017-10-03
Lower case CN for the kubelets to work with NodeRestriction policy

dkoshkin:
A regression as a side effect of https://github.com/apprenda/kismatic/issues/602 was introduced. The CN name must match exactly the nodeName, and the nodeName is always lowercased.

If users did have an uppercase in their certs during an upgrade the validation will catch and fail, the user can then delete (or regenerate) the certificates with the lowercase CN.

### DIFF
--- a/pkg/install/pki.go
+++ b/pkg/install/pki.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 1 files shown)


alexbrand:
Great, I am glad we have that integration test. LGTM.

PR #846, opened 2017-10-03
Bump provisioner to v1.6.0

### LINKED ISSUE #838: Update the version of the provision binary
Update the `provision` binary's version to the new version that is generating the new plan file format.

alexbrand:
Fix #838


dkoshkin:
LGTM

alexbrand:
Merging as tests will not catch anything related to the provisioner.

PR #847, opened 2017-10-03
Upgrade Calico to v2.6.1

dkoshkin:
Related to https://github.com/projectcalico/calico/issues/1135

Looks like `v2.6.0` had a bad version of Felix. https://docs.projectcalico.org/v2.6/releases/


alexbrand:
LGTM

PR #848, opened 2017-10-03
Improve the UX of the seed command

### LINKED ISSUE #832: Refine seed-registry command UX
The new `seed-registry` command UX could be refined in two ways:
- [ ] Add an indicator that shows the index of the current image being pushed, and the total number of images. For example, one possible impl:
```
./kismatic seed-registry --server localhost:5000

Seed Container Image Registry=======================================================
(1/23) Pushed apprenda/tcp-healthz-amd64:v1.0.0                                 [OK]
```
- [ ] The `Pushed ${imageName}` line gets printed after the image has been pushed. If the image is not available on the local machine, it takes a while, and it makes it seem like the command got "stuck". Solving this might be more challenge due to our way of printing status.

alexbrand:
Fix #832 

https://asciinema.org/a/movZbAnMGAlQFIAm1zED1IJA8

### DIFF
--- a/pkg/cli/seed.go
+++ b/pkg/cli/seed.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os/exec"
+	"strings"
 
 	yaml "gopkg.in/yaml.v2"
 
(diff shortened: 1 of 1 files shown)


dkoshkin:
Awesome LGTM!

PR #852, opened 2017-10-04
Upgrade weave to 2.0.4, mount xtables.lock in the container

dkoshkin:
Intermittent issues were discovered during integration tests. 
Use fix from https://github.com/weaveworks/weave/pull/3134


alexbrand:
Do you want to make the circle change here?

dkoshkin:
Can we do a different PR for that

alexbrand:
Sounds good. This LGTM

PR #861, opened 2017-10-16
Update Go, Glide, and container user permissions to work on Darwin 17…

emmetthitz:
Build changes to fix glide/container issues on high sierra.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,11 +13,11 @@ HOST_GOOS = $(shell go env GOOS)
 HOST_GOARCH = $(shell go env GOARCH)
 
 # Versions of external dependencies
-GLIDE_VERSION = v0.11.1
+GLIDE_VERSION = v0.13.0
 ANSIBLE_VERSION = 2.3.0.0
 PROVISIONER_VERSION = v1.6.1
 KUBERANG_VERSION = v1.2.2
-GO_VERSION = 1.8.0
+GO_VERSION = 1.8.4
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -13,11 +13,11 @@ HOST_GOOS = $(shell go env GOOS)
 HOST_GOARCH = $(shell go env GOARCH)
 
 # Versions of external dependencies
-GLIDE_VERSION = v0.11.1
+GLIDE_VERSION = v0.12.3
```
dkoshkin: Glide v0.13.0 is out

Makefile
```diff
@@ -13,11 +13,11 @@ HOST_GOOS = $(shell go env GOOS)
...
 ANSIBLE_VERSION = 2.3.0.0
 PROVISIONER_VERSION = v1.6.1
 KUBERANG_VERSION = v1.2.2
-GO_VERSION = 1.8.0
+GO_VERSION = 1.8.3
```
dkoshkin: Go has 1.8.4 and 1.9.1

Makefile
emmetthitz: Good catch.

Makefile
emmetthitz: I'll give go 1.9.1 a shot. It broke some stuff with ansible last time I gave it a shot, but let's see.

dkoshkin:
LGTM!

Issue #778, opened 2017-09-08
Remove ABAC authorization mode


alexbrand:
For the next minor release, we should remove the ABAC authorization mode, which has been deprecated since we introduced RBAC.

flyer103:
Hi, will this feature be planed to be released in the near future?

magowant:
We investigated disabling ABAC but immediately found issues, highlighting that the KET deployment of K8s is still heavily reliant on ABAC. 

For example, the kube-dns deployment is still using the default service account, which is restricted in RBAC only mode, effectively breaking kube-dns. Applying the principle of least privilege, rather than make the default service account more permissive, specific service accounts should be used for each component that needs access to the K8s API:

kube-proxy *
kube-scheduler *
kube-controller-manager *
kube-dns *
calico-node
calico-kube-controllers
tiller
rescheduler
ingress
others?

Some of the core kube-system components (*) already have specific service accounts made available by default. See https://kubernetes.io/docs/admin/authorization/rbac/#service-account-permissions. The others components will need to have service accounts created as part of the KET install.

swade1987:
Hey @magowant 

Thanks for this, you may be interested in https://github.com/apprenda/kismatic/pull/864 

This contains the work @dkoshkin has been doing regarding RBAC.

willvrny:
@alexbrand @magowant @swade1987 We will probably need service accounts for the components. One strategy is to create Roles/ClusterRoles with no permissions. Check for API requests failures and slowly enable permissions. This ensures that only the necessary capabilities are enabled.

It seems that some of the existing charts online are too permissive.

dkoshkin:
Closed in #864

PR #864, opened 2017-10-17
Use RBAC

### LINKED ISSUE #771: Is there a way to restrict `--authorization-mode=RBAC` for APIServer?

FEATURE REQUEST

**What happened**:

APIServer uses `RBAC` and `ABAC` as the authorization-mode and if modifying this parameter to `RBAC`, Calico will not access pod resources.

After setting `--authorization-mode=RBAC` and deleting `--authorization-policy-file` for APIServer manifest, APIServer will work as normally.

Set `--authorization-mode=RBAC` and delete `--authorization-policy-file` for APIServer manifest.

**Environment**:
- Kismatic version (use `kismatic version`): v1.5.3
- Kernel (e.g. `uname -a`):

### LINKED ISSUE #989: Upgrade dashboard to v1.8.0
The title says it all.

dkoshkin:
Fixes #771, fixes https://github.com/apprenda/kismatic/issues/778, fixes #989 

Removing ABAC needs to be a 2 step process.
1st - Modify any addons to use RBAC, this guarantees that during upgrades components will function during the process.
2nd - Remove ABAC in the next minor release

For any new installs only `ABAC` will not be enabled.

~A big TODO is how to deploy the kubernetes dashboard. The newest version allows to use token based auth.~
The installer will now generate a new kubeconfig file with a token having admin access.
```
➜  out git:(use-rbac) ✗ ./kismatic dashboard
Error generating a kubeconfig file, you may still use the dashboard with your own ServiceAccount token

Opening kubernetes dashboard in default browser...
Starting to serve on 127.0.0.1:8001

➜  out git:(use-rbac) ✗ ./kismatic dashboard
Opening kubernetes dashboard in default browser...
Use the kubeconfig in "generated/dashboard-admin-kubeconfig"
Starting to serve on 127.0.0.1:8001

```


alexbrand:
I don't think we can merge this without figuring out dashboard auth, right? Also, why do we have to remove ABAC in the next minor?

pkg/cli/apply.go
```diff
@@ -132,3 +144,46 @@ func (c *applyCmd) run() error {
 
 	return nil
 }
+
+func generateDashboardAdminKubeconfig(out io.Writer, generatedAssetsDir string, plan install.Plan) error {
+	// All of this is required because cannot set a label on the secret so no selectors
```
alexbrand: I am wondering if we should be doing this much hand-holding for the dashboard. It almost seems like we could just document how to access the dashboard, and let the user do it?

ansible/roles/kube-dns/templates/kubernetes-dns.yaml
```diff
@@ -1,3 +1,14 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kube-dns
```
alexbrand: Is this a special service account that has pre-existing clusterroles/clusterrolebindings?

swade1987:
Current customers are currently looking into disabling ABAC for their clusters, we have tried removing ABAC from the Kismatic configuration but have quickly found the Kismatic deployment of Kubernetes is still heavily reliant on ABAC. As noted within Kismatic's documentation we would either have to define the applicable policies for RBAC authorisation or we will have to wait for a future release for ABAC to be removed. It'd be great if we could fully remove ABAC and go all-in on RBAC.

ansible/roles/kube-dns/templates/kubernetes-dns.yaml
dkoshkin: Exactly
```
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:kube-dns
subjects:
- kind: ServiceAccount
  name: kube-dns
  namespace: kube-system
```

pkg/cli/apply.go
dkoshkin: I agree with you, but because we have the existing behavior it would be really cumbersome for the user to loose the ability being able to start a cluster and have a working dashboard account.

swade1987:
If the `AdminPassword` field is deprecated how are users going to obtain the password and further more store it?

dkoshkin:
A password will not be used for any new installations.

A `basicauth.csv` file(how these user/password get picked up by kubernetes) can no longer be used with RBAC, it is only supported with ABAC.

ansible/group_vars/all.yaml
```diff
@@ -153,9 +154,9 @@ kubernetes_api_server_option_defaults:
   "allow-privileged": "true"
   "apiserver-count": "{{ kubernetes_master_apiserver_count }}"
   "anonymous-auth": "false"
-  "authorization-mode": "Node,RBAC,ABAC"
-  "authorization-policy-file": "{{ kubernetes_authorization_policy_path }}"
-  "basic-auth-file": "{{ kubernetes_basic_auth_path }}"
+  "authorization-mode": "Node,RBAC{% if upgrading is defined and upgrading|bool == true %},ABAC{% endif %}" #TODO remove ABAC
```
alexbrand: This means we *have* to remove ABAC by the next release, right? As upgrading would result in ABAC getting enabled again.

ansible/roles/kubenode-cert/tasks/main.yaml
```diff
@@ -55,6 +55,10 @@
       mode: "{{ kubernetes_certificates_mode }}"
     when: "['master','worker','ingress','storage'] | intersect(group_names) | length > 0"
     with_items:
+      - src: "admin.pem"
```
alexbrand: This is something that we should probably improve later on. In my mind, putting the root creds on every single node is not great.

integration/install_test.go
```diff
@@ -256,9 +255,9 @@ var _ = Describe("kismatic", func() {
 						})
 
 						// Use master[0] public IP
-						sub.It("should have an accessible dashboard", func() error {
-							return canAccessDashboard(fmt.Sprintf("https://admin:abbazabba@%s:6443/ui", nodes.master[0].PublicIP))
-						})
+						// sub.It("should have an accessible dashboard", func() error {
```
alexbrand: Should we get rid of these tests?

pkg/cli/upgrade.go
```diff
@@ -235,6 +235,13 @@ without the "--partial-ok" flag to perform a full upgrade.
 		}
 	}
 
+	// Generate dashboard admin certificate
+	util.PrintHeader(out, "Generating Dashboard Admin Kubeconfig File", '=')
```
alexbrand: What's the need for generating this kubeconfig during upgrade?

pkg/install/plan.go
```diff
@@ -267,13 +267,6 @@ func (fp *FilePlanner) PlanExists() bool {
 
 // WritePlanTemplate writes an installation plan with pre-filled defaults.
 func WritePlanTemplate(planTemplateOpts PlanTemplateOptions, w PlanReadWriter) error {
-	if planTemplateOpts.AdminPassword == "" {
-		pw, err := generateAlphaNumericPassword()
```
alexbrand: Can we get rid of the generateAlphaNumericPassword func? I think this might let us remove a dependency from glide too.

pkg/cli/dashboard.go
```diff
@@ -30,76 +32,84 @@ func NewCmdDashboard(out io.Writer) *cobra.Command {
...
+		fmt.Fprintf(out, "Unexpected error opening the kubernetes dashboard: %v. You may access it at %q", err, url)
 	}
-	fmt.Fprintln(out, "Opening kubernetes dashboard in default browser...")
-	//Not obvious, but this is for escaping userinfo
-	urlFmted := fmt.Sprintf("https://%s@%s:6443/ui", url.UserPassword("admin", plan.Cluster.AdminPassword), plan.Master.LoadBalancedFQDN)
-	if err := browser.OpenURL(urlFmted); err != nil {
-		// Don't error. Just print a message if something goes wrong
-		fmt.Fprintf(out, "Unexpected error opening the kubernetes dashboard: %v. You may access it at %q", err, req.URL)
+
+	cmd := exec.Command("./kubectl", "proxy", "--kubeconfig", kubeconfig)
```
alexbrand: Will the user have to refresh the page after the proxy starts?

ansible/roles/kubenode-cert/tasks/main.yaml
swade1987: Surely this is making things less secure than it is already?

alexbrand: We currenlty deploy a kubeconfig on all nodes with the admin username/password.

swade1987: Damn, I didn't realise that

pkg/cli/upgrade.go
dkoshkin: None, forgot to remove after moving the generation to the `kismatic dashboard` command. Thanks

pkg/cli/dashboard.go
dkoshkin: Not on my Mac with Chrome, it automatically refreshes but not sure if it will be different on other platforms.

ansible/group_vars/all.yaml
dkoshkin: Made this better 
```
{% if upgrading is defined and upgrading|bool == true and kubernetes_admin_password is defined and kubernetes_admin_password != '' %}
```

integration/install_test.go
dkoshkin: I want to keep it as a reminder to look at these at some point.

ansible/roles/kubeconfig/templates/kubeconfig.j2
```diff
@@ -8,8 +8,12 @@ clusters:
 users:
 - name: admin
   user:
+{% if upgrading is defined and upgrading|bool == true %}
```
alexbrand: Do we need to check for admin_password here instead?

glide.lock
```diff
@@ -1,68 +1,68 @@
 hash: f1e1c7515c49bc69417f943bf7c8dc863edf1b0936bd768db5284244e97ca0d5
-updated: 2017-07-25T11:20:54.761474195-04:00
```
alexbrand: I think we need to update glide.yaml as well, right?

pkg/cli/dashboard.go
alexbrand: Hmmm okay. Might be worth starting the kubectl proxy before opening the browser, but not sure if it matters that much.

alexbrand:
Overall this LGTM

ansible/roles/kubeconfig/templates/kubeconfig.j2
dkoshkin: Good catch!

glide.lock
dkoshkin: Yeah that dep was not in `glide.yaml` not sure how thats possible though, am I missing something?

emmetthitz: Yeah. I think so. If the import gets removed, a `glide update` should axe it.

alexbrand: Huh, alrighty then!

Issue #809, opened 2017-09-25
Kube-proxy healthz and metrics endpoint was changed in 1.7


alexbrand:
This needs further investigation to figure out what needs to be changed on our side.

https://github.com/kubernetes/kubernetes/issues/48600

PR #865, opened 2017-10-17
Use port 10256 as the kube-proxy health port

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/809

Port `10249` is used for metrics and by default only binds to `127.0.0.1` no need for it to be *accessible* only *available*.

### DIFF
--- a/ansible/group_vars/all.yaml
+++ b/ansible/group_vars/all.yaml
@@ -18,7 +18,7 @@ etcd_k8s_client_port: 2379
 etcd_networking_client_port: 6666
 kubernetes_master_secure_port: 6443
 kubernetes_master_insecure_port: 8080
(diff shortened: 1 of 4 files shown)


alexbrand:
I believe @swade1987 or @willvrny wanted the metrics endpoint to be accessible from outside the node to scrape it from prometheus. Is that right @swade1987 / @willvrny?

dkoshkin:
There is also a kube-proxy flag that can be used to modify that behavior `--metrics-bind-addres`.

alexbrand:
Ah right. Users can set that flag if they want to expose the endpoint. That seems fine to me, but I wonder if our default should be to expose it.

swade1987:
@willvrny and myself tried this but couldn't seem to get it working properly with Prometheus.

PR #868, opened 2017-10-17
Docker upgrade

### LINKED ISSUE #739: Docker 1.12 is getting EOL'd in November
is it possible to use docker-ce instead of os distribution docker on ubuntu 16.04?

emmetthitz:
Exactly what it looks like

Fixes #739


ansible/group_vars/all.yaml
```diff
@@ -2,8 +2,8 @@
 # VERSIONS
 kubernetes_yum_version: 1.8.0-0
 kubernetes_deb_version: 1.8.0-00
-docker_engine_yum_version: 1.12.6-1.el7.centos
-docker_engine_apt_version: 1.12.6-0~ubuntu-xenial
+docker_engine_yum_version: 17.03.2.ce-1.el7.centos
+docker_engine_apt_version: 17.03.1~ce-0~ubuntu-xenial
```
dkoshkin: We cant use 2 different versions, especially since `17.03.1` is not the one thats tested by kubernetes.

emmetthitz: There is no 17.03.2 up on the ubuntu repositories though.

dkoshkin: Take a look at https://download.docker.com/linux/

emmetthitz: I lied.

swade1987:
@emmetthitz what is the latest status on this?

swade1987:
when is this likely to be merged?

emmetthitz:
@dkoshkin LGTM

ansible/roles/packages-docker/tasks/main.yaml
```diff
@@ -1,9 +1,8 @@
 ---
   # YUM
-  - name: install docker-engine yum package
-    yum:
-      name: docker-engine-{{ docker_engine_yum_version }}
-      state: present
+  # Can't use yum module as --setopt=obsoletes=0 is not supported
```
alexbrand: Might be useful to document why we need this flag.

emmetthitz:
Is there a reason we don't want to ship this with 1.7?

emmetthitz:
Or is it just the validation work that needs to be done?

swade1987:
We can't really be releasing new versions of KET with versions of Docker which are EOL

dkoshkin:
Ran a long test https://circleci.com/gh/apprenda/kismatic/1368 for this PR, this is ready to go.

Issue #837, opened 2017-09-30
Appending port to docker_registry_address causes problems


robb-foster-ck:
**Is this a BUG REPORT or FEATURE REQUEST?**:
Feature request.

**What happened**:
I am testing the disconnected installation with a private docker registry.  What I would like to do is specify a docker_registry_address of

`<fqdn of repository>/kismatic/1_5_0`

so I can version and organize the images that need to be installed locally.

**What you expected to happen**:
The above syntax to work.  Instead, the docker_registry_port is appended to the docker_registry_address, which breaks things.

`docker_registry_full_url: "{{ docker_registry_address }}:{{ docker_registry_port }}"`

**How to reproduce it (as minimally and precisely as possible)**:
Set docker_registry_address to something not rooted at the base of the docker repository.

**Anything else we need to know?**:

**Environment**:
- Kismatic version (use `kismatic version`): 1.5.0
- Plan file Gist (remove sensitive information if any): salient part
docker_registry:                         # Here you will provide the details of your Docker registry or setup an internal one to run in the cluster. This is optional and the cluster will always have access to the Docker Hub.
  setup_internal: false                  # When true, a Docker Registry will be installed on top of your cluster and used to host Docker images needed for its installation.
  address: "docker-registry.creditkarma.com/kismatic/1_5_0"                            # IP or hostname for your Docker registry. An internal registry will NOT be setup when this field is provided. Must be accessible from all the nodes in the cluster.
  port: 8443                             # Port for your Docker registry.

- Cloud provider or hardware configuration: HP Gen 9 blades
- OS (e.g. from /etc/os-release):
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
[...]

PR #869, opened 2017-10-17
Add new docker_registry.server to replace address and port

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/837

```
# plan file
docker_registry:                         
  server: registry.example.com:8443        
  CA: /certs/ca.rt    
```

This PR does not remove the fields completely just deprecates them, the user may still use the old fields.

TODO: modify provisioner.


docs/upgrade/v1.6.0/disconnected-upgrade.md
```diff
@@ -26,7 +26,7 @@ assist you in this task. For more information, see the
 
 ### Existing cluster with registry stood up by KET
 In this scenario, your local registry was stood up by KET and the docker registry
-related fields (`docker_registry.address` and `docker_registry.port`) are not 
+related field (`docker_registry.server`) is not
```
alexbrand: I wouldn't update this doc, given that it is specific to version 1.6.0, and this field won't be there in that version.

integration/plan_patterns.go
```diff
@@ -82,8 +81,7 @@ docker:
       block_device: "/dev/xvdb"
       enable_deferred_deletion: false{{end}}
 docker_registry:
-  address: {{.DockerRegistryIP}}
```
alexbrand: Will this removal affect upgrade tests where we need to set these values?

dkoshkin: I though that at first but the tests only use the registry when doing the upgrade.

alexbrand:
LGTM

PR #870, opened 2017-10-17
Fixes for #560 and #811

### LINKED ISSUE #560: dashboard fails when pass has slashes
eg. when pass is `zHWGW7SpuuafmUQuVdvjDVBa/xxhejAh`

```
$ ./kismatic dashboard -f ./kismatic-cluster.yaml
Error verifying connectivity to cluster dashboard: Get https://admin:zHWGW7SpuuafmUQuVdvjDVBa/xxhejAh@10.20.30.40:6443/ui: invalid URL port "zHWGW7SpuuafmUQuVdvjDVBa"
```
Possibly disturbing to some folks that it is printed to terminal.

This is on v1.3.3

### LINKED ISSUE #811: Kube-apiserver fails to properly restart when applying new kismatic-cluster info
Bug report

**What happened**:
`./kismatic install apply --restart-services --skip-preflight` does not properly restart the `kube-apiserver`, causing the apply checks to fail. I'm not sure if this fails when the `admin-password` is unchanged, but it does when the password is changed.

The `kube-apiserver` should restart, and properly accept the updated `admin-password` from the `kismatic-cluster.yaml`.

`./provision vagrant create`
`./kismatic install apply`
Change the admin password where specified previously
`./kismatic install apply --restart-services --skip-preflight`

**Environment**:
1.6.0-test
```cluster:
  name: kubernetes
  admin_password: xuzxu:orzujezg795                                 # This password is used to login to the Kubernetes Dashboard and can also be used for administration without a security certificate
  disable_package_installation: false                                     # When true, installation will not occur if any node is missing the correct deb/rpm packages. When false, the installer will attempt to install missing packages for you.
  package_repository_urls: ""                                             # Comma-separated list of URLs of the repositories that should be used during installation. These repositories must contain the kismatic packages and all their transitive dependencies.
  disconnected_installation: false                                        # Set to true if you have already installed the required packages on the nodes or provided a local URL in package_repository_urls containing those packages.
  disable_registry_seeding: false                                         # Set to true if you have seeded your registry with the required images for the installation.
  networking:
    pod_cidr_block: 172.16.0.0/16                                     # Kubernetes will assign pods IPs in this range. Do not use a range that is already in use on your local network!
    service_cidr_block: 172.20.0.0/16                             # Kubernetes will assign services IPs in this range. Do not use a range that is already in use by your local network or pod network!
    update_hosts_files: true                                              # When true, the installer will add entries for all nodes to other nodes' hosts files. Use when you don't have access to DNS.
    https_proxy:
    http_proxy:
    no_proxy:
  certificates:
    expiry: 17520h                                                        # Self-signed certificate expiration period in hours; default is 2 years.
    ca_expiry: 17520h                                                     # CA certificate expiration period in hours; default is 2 years.
  ssh:
    user: vagrant
    ssh_key: /Users/ehitz/.vagrant.d/insecure_private_key                        # Absolute path to the ssh public key we should use to manage nodes.
    ssh_port: 22
  kube_apiserver:
    option_overrides:
docker:
  storage:
    direct_lvm:                                                           # Configure devicemapper in direct-lvm mode (RHEL/CentOS only).
      enabled: false
      block_device: ""                                                    # Path to the block device that will be used for direct-lvm mode. This device will be wiped and used exclusively by docker.
      enable_deferred_deletion: false                                     # Set to true if you want to enable deferred deletion when using direct-lvm mode.
docker_registry:                                                          # Here you will provide the details of your Docker registry or setup an internal one to run in the cluster. This is optional and the cluster will always have access to the Docker Hub.
  setup_internal: false                  # When true, a Docker Registry will be installed on top of your cluster and used to host Docker images needed for its installation.
  address:                                    # IP or hostname for your Docker registry. An internal registry will NOT be setup when this field is provided. Must be accessible from all the nodes in the cluster.
  port: 8443                                      # Port for your Docker registry.
  CA:                                       # Absolute path to the CA that was used when starting your Docker registry. The docker daemons on all nodes in the cluster will be configured with this CA.
add_ons:
  cni:
    disable: false
    provider: calico                                                      # Options: 'calico','weave','contiv','custom'. Selecting 'custom' will result in a CNI ready cluster, however it is up to you to configure a plugin after the install.
    options:
      calico:
        mode: overlay                                                     # Options: 'overlay','routed'. Routed pods can be addressed from outside the Kubernetes cluster; Overlay pods can only address each other.
  dns:
    disable: false
  heapster:
    disable: false
    options:
      heapster_replicas: 2
      influxdb_pvc_name: ""                                               # Provide the name of the persistent volume claim that you will create after installation. If not specified, the data will be stored in ephemeral storage.
  dashboard:
    disable: false
  package_manager:
    disable: false
    provider: helm                                                        # Options: 'helm'
etcd:
  expected_count: 1
  nodes:
  - host: etcd001
    ip: 192.168.42.2
master:
  expected_count: 1
  nodes:
  - host: master001
    ip: 192.168.42.3
  load_balanced_fqdn: 192.168.42.3
  load_balanced_short_name: 192.168.42.3
worker:
  expected_count: 1
  nodes:
  - host: worker001
    ip: 192.168.42.4
ingress:
  expected_count: 1
  nodes:
  - host: worker001
    ip: 192.168.42.4
storage:
  expected_count: 0
  nodes:
```
3 node Vagrant cluster
macOS 10.12.6
- Kernel (e.g. `uname -a`):
Darwin Kernel Version 16.7.0

emmetthitz:
Fixes #560, fixes #811


dkoshkin:
LGTM

Issue #804, opened 2017-09-22
Add tolerations to components required on every node


willvrny:
FEATURE REQUEST

**Description**:

It would be great to add tolerations to the components, such as Calico, that Kismatic installs.

The scenario where this would be useful is if a taint was added to a node to only allow a certain workload on it. It would be great to be able to add another taint allowing core cluster components so that they are not evicted or can not be rescheduled onto the node in case of failure.

I am not sure of the granularity that of these taints and tolerations. I wouldn't expect the kubernetes-dashboard or the tiller pod to be included with components such as Calico. This is because certain components are a hard requirement on the node.

Until this is added it might make it either difficult or introduce risk when using taints in a cluster.

alexbrand:
We use the upstream deployment descriptors which already include tolerations. For example, the calico deployment includes the following tolerations:
```
      annotations:
        scheduler.alpha.kubernetes.io/critical-pod: ''
        scheduler.alpha.kubernetes.io/tolerations: |
          [{"key": "dedicated", "value": "master", "effect": "NoSchedule" },
           {"key":"CriticalAddonsOnly", "operator":"Exists"}]
```

I believe this is what you are referring to?

TimWoolford:
@alexbrand From my understanding those tolerations are only effective if the Rescheduler is running

https://github.com/apprenda/kismatic/issues/611

alexbrand:
@TimWoolford Would love to learn more about what you are doing, and see if we can bring it into KET.

Issue #611, opened 2017-06-26
Ensure critical add-ons remain scheduled on the cluster


alexbrand:
https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

willvrny:
Wouldn't that require installing this component:

https://github.com/kubernetes/kubernetes/blob/master/cluster/saltbase/salt/rescheduler/rescheduler.manifest

PR #873, opened 2017-10-20
Deploy pod resheduler by default

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/611, fixes https://github.com/apprenda/kismatic/issues/804

Deploy the static pod as described [here](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/)

### DIFF
--- a/ansible/roles/rescheduler/templates/rescheduler.yaml
+++ b/ansible/roles/rescheduler/templates/rescheduler.yaml
@@ -1,0 +1,26 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: rescheduler
+  namespace: kube-system
+  annotations:
+    scheduler.alpha.kubernetes.io/critical-pod: ''
+  labels:
+    k8s-app: rescheduler
+    version: {{ official_images.rescheduler.version }}
+    kubernetes.io/cluster-service: "true"
+    kubernetes.io/name: "Rescheduler"
+spec:
+  hostNetwork: true
+  containers:
+  - image: {{ images.rescheduler }}
+    name: rescheduler
+    volumeMounts:
+    # TODO: Make resource requirements depend on the size of the cluster
+    resources:
+      requests:
+        cpu: 10m
+        memory: 100Mi
+    command:
+    - /rescheduler 
+    - --running-in-cluster=false
--- a/docs/plan-file-reference.md
+++ b/docs/plan-file-reference.md
@@ -74,6 +74,8 @@
   * [package_manager](#add_onspackage_manager)
     * [disable](#add_onspackage_managerdisable)
     * [provider](#add_onspackage_managerprovider)
+  * [rescheduler](#add_onsrescheduler)
+    * [disable](#add_onsreschedulerdisable)
 * [features _(deprecated)_](#features-deprecated)
   * [package_manager _(deprecated)_](#featurespackage_manager-deprecated)
     * [enabled _(deprecated)_](#featurespackage_managerenabled-deprecated)
@@ -718,6 +720,20 @@
 | **Required** |  Yes |
 | **Default** | ` ` | 
 | **Options** |  `helm`
+
+###  add_ons.rescheduler
+
+ The Rescheduler add-on configuration. Because the Rescheduler does not have leader election and therefore can only run as a single instance in a cluster, it will be deployed as a static pod on the first master. More information about the Rescheduler can be found here: https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/ 
+
+###  add_ons.rescheduler.disable
+
+ Whether the pod rescheduler add-on should be disabled. When set to true, the rescheduler will not be installed on the cluster. 
+
+| | |
+|----------|-----------------|
+| **Kind** |  bool |
+| **Required** |  No |
+| **Default** | `false` | 
 
 ##  features _(deprecated)_
 
--- a/pkg/install/test/plan-template-with-storage.golden.yaml
+++ b/pkg/install/test/plan-template-with-storage.golden.yaml
@@ -170,6 +170,10 @@ add_ons:
     # Options: 'helm'
     provider: helm
 
+  # The rescheduler ensures that critical add-ons remain running on the cluster.
+  rescheduler:
+    disable: false
+
 # Etcd nodes are the ones that run the etcd distributed key-value database.
 etcd:
   expected_count: 3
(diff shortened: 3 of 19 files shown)


ansible/roles/rescheduler/templates/rescheduler.yaml
```diff
@@ -0,0 +1,36 @@
...
+      requests:
+        cpu: 10m
+        memory: 100Mi
+    command:
+    # TODO: split this out into args when we no longer need to pipe stdout to a file #6428
+    - sh
+    - -c
+    - '/rescheduler --running-in-cluster=false 1>>/var/log/rescheduler.log 2>&1'
+  volumes:
+  - hostPath:
```
alexbrand: Is there any docs around why this is a requirement? Seems weird to log directly to the host.

docs/plan-file-reference.md
```diff
@@ -719,6 +721,20 @@
 | **Default** | ` ` | 
 | **Options** |  `helm`
 
+###  add_ons.rescheduler
+
+ The Rescheduler add-on configuration. More invo https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/
```
alexbrand: Typo "invo". A full sentence here is probably more appropriate. "More information about the Rescheduler can be found here: ..."

docs/plan-file-reference.md
```diff
@@ -719,6 +721,20 @@
 | **Default** | ` ` | 
 | **Options** |  `helm`
 
+###  add_ons.rescheduler
+
+ The Rescheduler add-on configuration. More invo https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/ 
+
+###  add_ons.rescheduler.disable
+
+ Whether the pod rescheduler add-on should be disabled. When set to true, the package manager will not be installed on the cluster.
```
alexbrand: "package manager" mentioned

pkg/install/test/plan-template-with-storage.golden.yaml
```diff
@@ -170,6 +170,10 @@ add_ons:
     # Options: 'helm'
     provider: helm
 
+  # Rescheduler add-on used for scheduling of other critical add-ons.
```
alexbrand: Does the rescheduler actually take over scheduling the critical add-ons? My understanding is that it would only kick in under certain conditions? I think this could say: "The rescheduler ensures that critical add-ons remain running on the cluster"

ansible/roles/rescheduler/templates/rescheduler.yaml
dkoshkin: `kubectl logs` doesn't work for static pods, this is a way to log it without having to use `docker logs`

willvrny: Would it be better to run the rescheduler as a DaemonSet? Then we can utilise existing logging pipelines? 

In the docs for static pods (https://goo.gl/ZCfbcZ) it says:

```If you are running clustered Kubernetes and are using static pods to run a pod on every node, you should probably be using a DaemonSet!```

My concern is that this is another log file that might need to be rotated and managed?

Comment (dkoshkin) [reply ansible/roles/rescheduler/templates/rescheduler.yaml]: From what I read, you can only have one instance of the Rescheduler. The reasoning as to why run this as a static pod is something needs to make sure that this pod is up, running it as a Deployment would not guarantee this.

Comment (alexbrand) [reply ansible/roles/rescheduler/templates/rescheduler.yaml]: Ah, I hadn't seen that we were putting it only in the first master. I think this is a problem... what if the node dies?

Comment (dkoshkin) [reply ansible/roles/rescheduler/templates/rescheduler.yaml]: This seems like a problem that doesn't yet have a good solution until it gets fixed upstream.

Comment (willvrny) [reply ansible/roles/rescheduler/templates/rescheduler.yaml]: I see. It is a shame that we can't use docker logs to keep consistency with other containers.

File: ansible/roles/rescheduler/templates/rescheduler.yaml [hunk 5]
```diff
@@ -0,0 +1,36 @@
...
+    # TODO: Make resource requirements depend on the size of the cluster
+    resources:
+      requests:
+        cpu: 10m
+        memory: 100Mi
+    command:
+    # TODO: split this out into args when we no longer need to pipe stdout to a file #6428
+    - sh
+    - -c
+    - '/rescheduler --running-in-cluster=false 1>>/var/log/rescheduler.log 2>&1'
```
willvrny: Is there any way we can support the flags you can pass to the rescheduler e.g. initialDelay, listenAddress (metrics) or gracePeriod?

See here: https://github.com/kubernetes/contrib/blob/master/rescheduler/rescheduler.go

ansible/roles/rescheduler/templates/rescheduler.yaml
alexbrand: Damn, this doesn't feel good. Also, what is the reason for this pod being different than the other static pods that go through docker logs? This adds another log file that needs to be rotated, instead of being rotated automatically by docker.

dkoshkin: Yeah I agree, I don't like the logging into a separate file, I think logging it out to stdout/stderr and let docker handle that would be better.

willvrny: Can we not run this in cluster with and set the instance to 1? Then we can treat it like a normal pod with regards to logging etc?

dkoshkin: But its not guaranteed to be scheduled then. From https://github.com/kubernetes/kubernetes/issues/44265
>Rescheduler is a component of the same class as scheduler. As you expect to have scheduler running in your cluster out of the box also rescheduler should run there out of the box. If not this is a bug/feature request to your setup tool. We do not expect users to run rescheduler on their own. Rescheduler must run on master node and master is usually marked as unschedulable, so that it's hard for user to run it there.

willvrny: Okay I understand now. I agree with you both on logging to stdout/stderr is better as it avoids filling the disk on the node. Can we make a change to do that?

alexbrand:
I think this looks good to me. We should probably call out somewhere that the rescheduler is running on the first master node, and that this is due to the limitation of the rescheduler not having leader election built in.

dkoshkin:
Added the note in the plan-file-reference [doc](https://github.com/apprenda/kismatic/pull/873/files#diff-f5f00b312fbf6dd1efbb751aa6e1f8b1R724). I dont think it needs any more as the actual docs linked in there explains how it works.

alexbrand:
Sounds good!

emmetthitz:
I think this LGTM as well.

Issue #780, opened 2017-09-11
Duplicate error checks are printed when a node has multiple roles 


dkoshkin:
**Is this a BUG REPORT or FEATURE REQUEST?**:
BUG

**What happened**:
Inspector prints duplicate error checks are printed
```
=> The following checks failed on "worker001":
   - Package "docker-engine 1.12.6-0~ubuntu-xenial": package is not installed, and is not available in known package repositories
   - Package "docker-engine 1.12.6-0~ubuntu-xenial": package is not installed, and is not available in known package repositories
   - Package "kubelet 1.7.4-00": package is not installed, and is not available in known package repositories
   - Package "kubelet 1.7.4-00": package is not installed, and is not available in known package repositories
   - Package "kubectl 1.7.4-00": package is not installed, and is not available in known package repositories
   - Package "kubectl 1.7.4-00": package is not installed, and is not available in known package repositories
```

**What you expected to happen**:
The same check is printed once
```
=> The following checks failed on "worker001":
   - Package "docker-engine 1.12.6-0~ubuntu-xenial": package is not installed, and is not available in known package repositories
   - Package "kubelet 1.7.4-00": package is not installed, and is not available in known package repositories
   - Package "kubectl 1.7.4-00": package is not installed, and is not available in known package repositories
```

**How to reproduce it (as minimally and precisely as possible)**:
In the plan file have a single node for both the `worker` and `ingress`.
Do not install packages on the node and set `disable_package_installation: true`

**Anything else we need to know?**:

**Environment**:
- Kismatic version (use `kismatic version`):
```
Kismatic:
  Version: 1.5.3-15-ge8d76fb1-dirty
  Built: Mon Sep 11 18:48:31 UTC 2017
  Go Version: go1.8
```
- Plan file Gist (remove sensitive information if any): 
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):

dkoshkin:
Closed in #875

Issue #672, opened 2017-07-22
Suggestion : Validate should check for Docker on Node


robotica72:
Obviously this isn't a huge issue, but since there is a validate command to check the nodes for a smooth upgrade, you might want to add a check for the installation of Docker on the nodes.  I forgot I had docker on my one image and I cloned it out to 13 machines for a new lab at home, validate was clean - so I kicked off an apply and it bombed at the Docker check since it was installed on the nodes.  Easy enough to remove it with an ssh command, but still - an add on into validate might help newer people?

Just a thought... and great work as always....

alexbrand:
@robotica72 Can you share a bit more about how it bombed with Docker? Thanks!

robotica72:
Hey - sorry meant to add to my statement that day but got sidetracked - It bombs out installing docker due to docker conflicting with container-selinux-2.12-2.gite7096ce.el7.noarch - But the validate passed this section without any issues.  (I should have known better, I just forgot about my base template and what was in it).   So overall, just a suggestion for anyone who may be newer/etc ...  

But excellent installer overall - best I have seen for K8s.

dkoshkin:
Closed in #875

dkoshkin:
Commented out the checks until there is a mechanism to have docker preinstalled on the node by the user.

dkoshkin:
Closed in #1051

PR #875, opened 2017-10-25
Validate docker package

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/672, fixes https://github.com/apprenda/kismatic/issues/780

Added new inspector rules to check for packages that should not be installed, ie `docker`
```
Run Cluster Pre-Flight Checks                                                   [ERROR]
- Task: run pre-flight checks using Kismatic Inspector from the master
=> The following checks failed on "worker001":
   - Port Available: 4194
   - Port Available: 10248
   - Port Available: 10249
   - Port Available: 10256
   - Port Available: 10250
   - Port Available: 10255
   - Port Available: 80
   - Port Available: 443
   - Port Available: 10254
   - Package "docker": package should not be installed
   - Package "docker-engine": package should not be installed
   - Port Accessible: 4194
   - Port Accessible: 10256
   - Port Accessible: 10250
   - Port Accessible: 80
   - Port Accessible: 443
   - Port Accessible: 10254
=> The following checks failed on "master001":
   - Port Available: 6443
   - Port Available: 8080
   - Port Available: 10251
   - Port Available: 10252
   - Port Available: 4194
   - Port Available: 10248
   - Port Available: 10249
   - Port Available: 10256
   - Port Available: 10250
   - Port Available: 10255
   - Package "docker": package should not be installed
   - Package "docker-engine": package should not be installed
   - Port Accessible: 6443
   - Port Accessible: 10251
   - Port Accessible: 10252
   - Port Accessible: 4194
   - Port Accessible: 10256
   - Port Accessible: 10250
=> The following checks failed on "etcd001":
   - Port Available: 2379
   - Port Available: 6666
   - Port Available: 2380
   - Port Available: 6660
   - Package "docker": package should not be installed
   - Package "docker-engine": package should not be installed
   - Port Accessible: 2379
   - Port Accessible: 6666
   - Port Accessible: 2380
   - Port Accessible: 6660
```

This also fixed some rules that didn't actually run, ie:
```
 # cAdvisor
 - kind: TCPPortAvailable
-  when: ["master","worker","ingress","storage"]
+  when: ["master|worker|ingress|storage"]
   port: 4194
```
The rule above would require the node to have all those roles prior to the change.

### DIFF
--- a/pkg/inspector/check/package.go
+++ b/pkg/inspector/check/package.go
@@ -18,7 +18,6 @@ func (p PackageQuery) String() string {
 // package is installed.
 type PackageCheck struct {
 	PackageQuery         PackageQuery
-	ShouldNotBeInstalled bool
 	PackageManager       PackageManager
 	InstallationDisabled bool
 }
@@ -28,19 +27,12 @@ type PackageCheck struct {
 // there is no guarantee that the node will have the kismatic package repo configured.
 // For this reason, this check is a no-op when package installation is disabled.
 func (c PackageCheck) Check() (bool, error) {
+	if !c.InstallationDisabled {
+		return true, nil
+	}
 	installed, err := c.PackageManager.IsInstalled(c.PackageQuery)
 	if err != nil {
 		return false, fmt.Errorf("failed to determine if package is installed: %v", err)
-	}
-	if !c.InstallationDisabled {
-		if installed && c.ShouldNotBeInstalled {
-			return false, fmt.Errorf("package should not be installed")
-		}
-		return true, nil
-	}
-	// When installation is disabled, dont check what packages are installed
-	if c.ShouldNotBeInstalled {
-		return true, nil
 	}
 	if installed {
 		return true, nil
--- a/pkg/inspector/check/package.go
+++ b/pkg/inspector/check/package.go
@@ -1,7 +1,6 @@
 package check
 
 import (
-	"errors"
 	"fmt"
 )
 
@@ -20,6 +19,7 @@ func (p PackageQuery) String() string {
 // package is installed.
 type PackageCheck struct {
 	PackageQuery         PackageQuery
+	ShouldNotBeInstalled bool
 	PackageManager       PackageManager
 	InstallationDisabled bool
 }
@@ -29,12 +29,19 @@ type PackageCheck struct {
 // there is no guarantee that the node will have the kismatic package repo configured.
 // For this reason, this check is a no-op when package installation is disabled.
 func (c PackageCheck) Check() (bool, error) {
-	if !c.InstallationDisabled {
-		return true, nil
-	}
 	installed, err := c.PackageManager.IsInstalled(c.PackageQuery)
 	if err != nil {
 		return false, fmt.Errorf("failed to determine if package is installed: %v", err)
+	}
+	if !c.InstallationDisabled {
+		if installed && c.ShouldNotBeInstalled {
+			return false, fmt.Errorf("package should not be installed")
+		}
+		return true, nil
+	}
+	// When installation is disabled, dont check what packages are installed
+	if c.ShouldNotBeInstalled {
+		return true, nil
 	}
 	if installed {
 		return true, nil
@@ -45,7 +52,7 @@ func (c PackageCheck) Check() (bool, error) {
 		return false, fmt.Errorf("failed to determine if package is available for install: %v", err)
 	}
 	if !available {
-		return false, errors.New("package is not installed, and is not available in known package repositories")
+		return false, fmt.Errorf("package is not installed, and is not available in known package repositories")
 	}
-	return false, errors.New("package is not installed, but is available in a package repository")
+	return false, fmt.Errorf("package is not installed, but is available in a package repository")
 }
--- a/pkg/inspector/check/package.go
+++ b/pkg/inspector/check/package.go
@@ -6,9 +6,8 @@ import (
 
 // PackageQuery is a query for finding a package
 type PackageQuery struct {
-	Name       string
-	Version    string
-	AnyVersion bool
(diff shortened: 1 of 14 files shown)


emmetthitz:
LGTM

alexbrand:
If I am following this correctly, the ruleset will always fail. We are creating a rule that says docker-engine version X should be installed, and also have a rule saying that docker-engine any version should _not_ be installed.

The other thing I'd like us to consider is an alternate syntax for representing the when conditions:
```
when: # This list is ANDed
  - ["master", "worker", "ingress", "storage"] #these are ORd 
  - ["ubuntu"]
```

dkoshkin:
@alexbrand the `InstallationDisabled` option makes these rules mutually exclusive, although this is not clear from when writing the rules or reading the check,  I would love some more set of eyes on [this function](https://github.com/apprenda/kismatic/pull/875/files#diff-9295e7919a731815a3d9843d9a903d81L69) to see if it have be simplified, its too complicated.

On your second point I was hesitant at first to do that because I didnt want to mess up on any of the rules but seeing how much simpler it got after combining them Im going to change it to what you suggested.

pkg/inspector/check/package_not_installed.go
```diff
@@ -0,0 +1,41 @@
+package check
+
+import "fmt"
+
+type PackageNotInstalledCheck struct {
```
alexbrand: I wonder if there's a better name for this. Maybe `PackageVersionOrMissing`?

pkg/inspector/rule/package_not_installed.go
```diff
@@ -0,0 +1,38 @@
+package rule
+
+import (
+	"errors"
+	"fmt"
+)
+
+// The PackageNotInstalled validates that a specified package in not installed.
+type PackageNotInstalled struct {
```
alexbrand: As with the check, I wonder if there's a better name for this.... as it actually allows the package to be installed.

pkg/inspector/rule/rule_set.go
```diff
@@ -22,645 +22,527 @@ const defaultRuleSet = `---
 
 # Executables required by kubelet
 - kind: ExecutableInPath
-  when: ["master","worker"]
+  when:
+  - ["master", "worker", "ingress", "storage"]
```
alexbrand: Might be good to document _somewhere_ how these conditions work.. Maybe a comment at the top of this file?

pkg/inspector/check/package.go
```diff
@@ -1,19 +1,13 @@
...
 )
 
 // PackageQuery is a query for finding a package
 type PackageQuery struct {
-	Name       string
-	Version    string
-	AnyVersion bool
-}
-
-func (p PackageQuery) String() string {
```
alexbrand: Any reason for removing this?

pkg/inspector/check/package_not_installed.go
```diff
@@ -0,0 +1,41 @@
...
+	AcceptablePackageVersion string
+	PackageManager           PackageManager
+	InstallationDisabled     bool
+}
+
+// Check returns true if the specified package is not installed.
+// This will also return true if the version installed matches AcceptablePackageVersion.
+// When InstallationDisabled is true this check will always return true.
+func (c PackageNotInstalledCheck) Check() (bool, error) {
+	// don't check when installation is disabled
```
alexbrand: I think it makes sense to check even installation is disabled, no?

pkg/inspector/rule/engine.go
```diff
@@ -67,17 +67,19 @@ func (e *Engine) CloseChecks() error {
 }
 
 func shouldExecuteRule(rule Rule, facts []string) bool {
```
alexbrand: I think having unit tests for this func will be useful, or maybe expanding the existing `TestEngine` ones to exercise the new OR/AND capabilities.

pkg/inspector/rule/rule_set.go
dkoshkin: Done.

pkg/inspector/rule/engine.go
dkoshkin: Done

pkg/inspector/check/package_not_installed.go
dkoshkin: The motivation for the name was that because we want to check that a certain package is not installed. 

Having "missing" in the name makes me think of missing from a repo, not really that its "not installed".

pkg/inspector/check/package_not_installed.go
dkoshkin: This way its consistent with the package check. A user might also want to install some other version of a package, maybe a different Docker package that might be on the list of "NotInstalled"

pkg/inspector/check/package.go
dkoshkin: The `String()` method is not used.

The `AnyVersion` seemed unnecessary, as it is simpler to just pass in an empty `Version` string.

It also consistent with the newly introduced check. As it checks for a package with any version to not be installed.

pkg/inspector/check/package_not_installed.go
alexbrand: Yeah, I don't have a better suggestion unfortunately.

pkg/inspector/rule/rule_set.go
alexbrand: Sweet!

alexbrand:
Overall LGTM

pkg/inspector/check/package.go
dkoshkin: Ahh yep I will put it back.

dkoshkin: Fixed

Issue #878, opened 2017-10-26
New Command to start KET in HTTP server mode


dkoshkin:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST

**What you expected to happen**:
KET should expose a new method to interact with the clusters it manages. In addition to the existing CLI interface, a new HTTP interface will be added. 

The HTTP(s) server will be responsible for exposing a RESTful API to create, mutate, delete and view clusters.

Criteria:
- [x] new command - `kismatic server` starts an HTTP(s) server
- [x] handle JSON payloads
- [x] support TLS

alexbrand:
Completed in #883.

PR #883, opened 2017-10-30
Http server

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/878

Added a new `server` package that will contain all code for the KET API.
* Using [httprouter](https://github.com/julienschmidt/httprouter) - a lightweight router
* Using [negroni](https://github.com/urfave/negroni) - some important middleware(Logger, Recovery)
* Added a `/heatlhz` route
* Added a `main()` for simpler/faster testing cycle

Added the new `server` command

TODO:
- [ ] Regenerate CLI docs

### DIFF
--- a/pkg/server/http/server.go
+++ b/pkg/server/http/server.go
@@ -1,0 +1,121 @@
+package http
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"time"
+
+	"crypto/tls"
+
+	"github.com/apprenda/kismatic/pkg/server/http/handler"
+	"github.com/julienschmidt/httprouter"
+	"github.com/urfave/negroni"
+)
+
+// Server to run an HTTPs server
+type Server interface {
+	RunTLS() error
+	Shutdown(timeout time.Duration) error
+}
+
+type loggerHttpServer struct {
+	httpServer *http.Server
+	logger     *log.Logger
+}
+
+// New creates a configured http server
+// If certificates are not provided, a self signed CA will be used
+// Use 0 for no read and write timeouts
+func New(logger *log.Logger, port string, certFile string, keyFile string, readTimeout time.Duration, writeTimeout time.Duration) (Server, error) {
+	if logger == nil {
+		return nil, fmt.Errorf("logger cannot be nil")
+	}
+	if port == "" {
+		return nil, fmt.Errorf("port cannot be empty")
+	}
+	addr := fmt.Sprintf(":%s", port)
+	if readTimeout < 0 {
+		return nil, fmt.Errorf("readTimeout cannot be negative")
+	}
+	if readTimeout == 0 {
+		logger.Printf("ReadTimeout is set to 0 and will never timeout, you may want to provide a timeout value\n")
+	}
+	if writeTimeout < 0 {
+		return nil, fmt.Errorf("writeTimeout cannot be negative")
+	}
+	if writeTimeout == 0 {
+		logger.Printf("WriteTimeout is set to 0 and will never timeout, you may want to provide a timeout value\n")
+	}
+	// use self signed CA
+	var keyPair tls.Certificate
+	if certFile == "" || keyFile == "" {
+		logger.Printf("Using self-signed certificate\n")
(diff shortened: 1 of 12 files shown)


pkg/server/http/server_test.go
```diff
@@ -0,0 +1,102 @@
...
+func TestHandler(t *testing.T) {
+	if testing.Short() {
+		return
+	}
+	logger := DefaultLogger(os.Stdout, "[tester] ")
+	s, err := New(logger, "", "", "", -1, -1)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	ts := httptest.NewServer(s.httpServer.Handler)
```
alexbrand: This seems like a redundant test to me, as we are not testing the actual server, and we already have a test for the handler.

pkg/server/http/server.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/urfave/negroni"
+)
+
+const (
+	writeTimeout = 10 * time.Second
+	readTimeout  = 10 * time.Second
+)
+
+// Server holds the details for an http(s) server
+// Leave cert and key empty to use a self signed CA
```
alexbrand: I don't see a `cert` or `key` field in the Server struct. We can probably remove from this comment?

If the Server struct needs to be initialized via `New`, I think I'd make it unexported.

pkg/server/http/server.go
```diff
@@ -0,0 +1,136 @@
...
+	logger     *log.Logger
+}
+
+const (
+	loggerPrefix   = "[kismatic] "
+	defaultPort    = "8443"
+	defaultTimeout = 15 * time.Second
+)
+
+// New creates a configured http server with a gorilla mux router
```
alexbrand: I think we can remove implementation details (gorilla, negroni) out of this comment, as they are already outdated.

pkg/server/http/server.go
```diff
@@ -0,0 +1,136 @@
...
+// And negroni Recovery and Logger middleware
+// If certificates are not provided, a self signed CA will be used
+// Use negative numbers for default read and write timeouts
```
alexbrand: Not a fan of this. I am wondering if we are better off not having defaults.

pkg/server/http/server.go
```diff
@@ -0,0 +1,136 @@
...
+		},
+		logger: logger,
+	}
+
+	return s, nil
+}
+
+// RunTLS support
+func (s *Server) RunTLS() error {
+	s.logger.Printf("Listening on 0.0.0.0%s\n", s.httpServer.Addr)
```
alexbrand: Typo `0.0.0.0 + s.httpServer.Addr`?

pkg/tls/ca.go
```diff
@@ -35,12 +36,15 @@ func NewCACert(csrFile string, commonName string, expiry string) (key, cert []by
 	if err != nil {
 		return nil, nil, fmt.Errorf("error opening %q", csrFile)
 	}
+	return NewCACertFromIOReader(f, commonName, expiry)
+}
+
+func NewCACertFromIOReader(r io.Reader, commonName string, expiry string) (key, cert []byte, err error) {
```
alexbrand: Thoughts on dropping IO from the func's name here and renaming `r` to `csrReader`?

pkg/server/http/server.go
dkoshkin: `s.httpServer.Addr` will be `:443`, did you mean the missing `:`?

pkg/server/http/server.go
alexbrand: Actually, not sure about making Server unexported.

pkg/server/http/server.go
alexbrand: Oh I see. So there's no way to start the server on a specific interface?

alexbrand:
This looks good to me overall. The only concern I have at this point is the license of `github.com/julienschmidt/httprouter`. Seems to be a non-standard license? I can't tell.

dkoshkin:
Merging for now, we can discuss the license later on.

PR #889, opened 2017-10-31
[WIP] Cluster controller for daemon mode

alexbrand:
Initial work on the cluster controller, mainly to get eyes early on in the process. 

It currently handles a single cluster, and it ignores a bunch of errors. I still need to think more about those scenarios.


emmetthitz:
LGTM so far.

alexbrand:
I think this is ready to go. There's some follow up work to be done once the provisioning piece is ready.

dkoshkin:
LGTM! Please consider merging in https://github.com/apprenda/kismatic/pull/905, this should also allow you to fix the leaky gor test.

Issue #629, opened 2017-07-06
Version labels for all cluster components


jonnymatts:
Version labels on cluster components would be useful for cluster monitoring.

For example, the kubedns and calico pods.

willvrny:
Hey @jonnymatts, 

It would good to expand this a little bit more. 

- What components do you want specifically? 
- Are you just expecting a `version: 1.0.0` label?

It looks like this could be done with a change to the yaml files in the Ansible part of Kismatic.

@dkoshkin & @alexbrand can we label this as 'help wanted'. This would be a good starting point for anyone looking to contribute.

swade1987:
Would also be useful to have the kubernetes version as a label on all component pods.

jonnymatts:
It would be good to get this information for all of the components that Kismatic installs for you. A label like you have suggested would be great.

swade1987:
@jonnymatts sounds good I will work with the team to get working on this.

alexbrand:
I am curious about the use case here, as the version is readily available in the podspec.

jonnymatts:
We add the versions of our other components as a label so it will be scraped by prometheus. We use these values to show the versions of our dependencies in a dashboard.

willvrny:
With the right logging tools we can also get the label as metadata on the logs for debugging purposes.

dkoshkin:
@jonnymatts and @willvrny how do you see a deployment with multiple containers working? 
Also for your usecase does it have to be a label or can it be an annotation?

jonnymatts:
I think it would have to be a label for Prometheus to scrape it. And I would imagine a label like '<component>_version: 1.0.0' so it should work with multiple containers. What do you think?

willvrny:
@jonnymatts So as an example for kube-dns on the deployment we'd have:

```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: kube-dns
  namespace: kube-system
  labels:
    k8s-app: kube-dns
    kubernetes.io/cluster-service: "true"
    addonmanager.kubernetes.io/mode: Reconcile
    kubedns_version: "1.14.5"
    kubedns_dnsmasq_version: "1.14.5"
    kubedns_sidecar_version: "1.14.5"
  annotations:
    kismatic/version: "{{ kismatic_short_version }}"
spec:
...
```

[...]

PR #892, opened 2017-11-02
Cluster components version labels

willvrny:
This adds version labels to cluster components as requested in issue #629.

I've added these labels to pods as this seemed like a better resource to label. Also if using a log forwarder for getting metadata it will fetch the labels from the pod.

I've added labels for all containers to cover when a resource has multiple containers with different versions.

The only component missing this label in Tiller as it is installed via `helm init`. To add this we'd need to install Tiller via normal Kubernetes resource files.

I also set explicit selectors on components that were missing them as they would default to using all labels on a pod which includes the version label. This version is likely to change between upgrades. In Kubernetes deployment `apps/v1beta2` selectors will become mandatory:

>In API version apps/v1beta2, .spec.selector and .metadata.labels no longer default to .spec.template.metadata.labels if not set. So they must be set explicitly. Also note that .spec.selector is immutable after creation of the Deployment in apps/v1beta2 - [Kubernetes Docs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#selector)

### DIFF
--- a/ansible/roles/calico-network-policy/templates/policy-controller.yaml
+++ b/ansible/roles/calico-network-policy/templates/policy-controller.yaml
@@ -25,6 +25,7 @@ spec:
       labels:
         tier: control-plane
         k8s-app: calico-kube-controllers
+        calico_kube_controller_component_version: "{{official_images.calico_kube_controller.version}}"
     spec:
       # The controllers must run in the host network namespace so that
       # it isn't governed by policy that would prevent it from working.
--- a/ansible/roles/calico-network-policy/templates/policy-controller.yaml
+++ b/ansible/roles/calico-network-policy/templates/policy-controller.yaml
@@ -16,6 +16,9 @@ metadata:
 spec:
   # The controllers can only have a single active instance.
   replicas: 1
+  selector:
+    matchLabels:
+      k8s-app: calico-kube-controllers
   strategy:
     type: Recreate
   template:
--- a/ansible/roles/calico/templates/calico.yaml
+++ b/ansible/roles/calico/templates/calico.yaml
@@ -72,6 +72,8 @@ spec:
     metadata:
       labels:
         k8s-app: calico-node
+        calico_node_component_version: "{{official_images.calico_node.version}}"
+        calico_cni_component_version: "{{official_images.calico_cni.version}}"
       annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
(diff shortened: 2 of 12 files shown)


alexbrand:
Interesting - so even when the component has a single container, we will still use `blah_component_version`?

willvrny:
@alexbrand Yes for consistency. At some point it might be good to review all the templates to ensure consistency.

jaycoon:
I don't like adding all these labels to the spec templates, we need a better solution. The metrics are exposed if using kube-state-metrics, though there are other options to surface these for Prometheus. It definitely shouldn't be a default label though.

willvrny:
@jaycoon @alexbrand There's also the scenario of getting this metadata into the logs. It might be useful to see component version if you start seeing errors in the logs. You could track it down to when the version changed.

We should ideally be enforcing a consistent set of metrics across all our components.

swade1987:
@jaycoon it's about having all necessary information when performing root-cause analysis and log reviewal.

Note: Kube-state-metrics is only useful when monitoring and alerting, when you need to start looking at the logs this information isn't available out the box.

willvrny:
@alexbrand Is this okay to merge?

alexbrand:
I think it looks good to me, but it seems like there was some disagreement on it? 

cc @jaycoon @dkoshkin

dkoshkin:
LGTM!

jaycoon:
Sorry guys, I still do not like putting image component versions in a label... this kind of data should be in an annotation. e.g.

```
kubedns/dnsmasq/version: "1.14.5"
kubedns/sidercar/version: "1.14.5"
```

Annotations are used for debugging, labels are for selectors and making decisions... it doesn't make sense to put component versions in a label (unless you need it to make a scheduling decision, which is not the use case being described here).

alexbrand:
I actually agree with @jaycoon and brought that point up when the issue was originally filed.

dkoshkin:
+1 with @jaycoon however as I understand this, prometheus uses labels and not annotations.

swade1987:
Labels are useful for logging, you can't easily obtain the annotations from tools like fluentbit.

swade1987:
What is the agreement on how to take this forward?

emmetthitz:
This should also be moved to a branch and run against CI. I don't think this nor #903 will have any issues, but just to be safe.

swade1987:
What is the status on this one?

PR #898, opened 2017-11-02
[WIP] Added store pkg

dkoshkin:
Added a watched store backed by [boltdb](https://github.com/boltdb/bolt).

### DIFF
--- a/pkg/store/boltdb.go
+++ b/pkg/store/boltdb.go
@@ -1,0 +1,178 @@
+package bolt
+
+import (
+	"fmt"
+	"log"
+	"os"
+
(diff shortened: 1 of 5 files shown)


pkg/store/boltdb.go
```diff
@@ -0,0 +1,178 @@
...
+	List(bucket string) ([]Response, error)
+	Delete(bucket, key string) error
+}
+
+type boltDB struct {
+	db       *bolt.DB
+	notifier chan<- interface{}
+}
+
+type Response struct {
```
alexbrand: I think this could be called `Entry` or something like that

Issue #879, opened 2017-10-27
HTTP Server Endpoints


dkoshkin:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST

**What you expected to happen**:
KET needs to expose a set of API endpoints to support cluster management over HTTP. 
The endpoints to Create,Update and Delete the cluster resource will all be asynchronous.  
The cluster name will be used as a key and will need to be unique for all clusters managed by that instance of KET.

**Cluster Resource**
- name: string
- state: string
- desiredState: string
- plan: KETPlan

**Endpoints**
- [x] GET /clusters
  Req body: empty
  Response: List of clusters

- [x] POST /clusters
  Req body: cluster resource
  Resp code: 201 (resource created)

- [x] PATCH /clusters/${name} (Kicking off a retry, adding/deleting workers)
  Resp: 202 Accepted
  Req body: cluster resource

- [x] GET /clusters/${name}
  Resp: Cluster details, including state (created, provisioning, installing, running, modifying, deleting)

- [x] DELETE /clusters/${name}
  Resp: 202 Accepted
  The client is responsible for checking if deletion was successful

- [x] GET /clusters/${name}/statuslogs
  Resp: If an operation is in progress, the logs of the current operation are returned. In any other 
  case, the logs of the previous operation are returned.

- [x] GET /clusters/${name}/kubeconfig
  Resp: Kubeconfig file for the cluster

- [x] GET /clusters/${name}/assets
  Resp: Returns a tarball or zip file containing all the generated assets of the cluster (kubeconfig, certificates, etc).

alexbrand:
Closing, as this is done.

PR #900, opened 2017-11-03
Added handlers to POST and GET /clusters

dkoshkin:
Starting on https://github.com/apprenda/kismatic/issues/879

### DIFF
--- a/pkg/server/http/handler/clusterers.go
+++ b/pkg/server/http/handler/clusterers.go
@@ -1,0 +1,52 @@
+package handler
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/apprenda/kismatic/pkg/server/http/model"
+	"github.com/apprenda/kismatic/pkg/server/http/service"
+	"github.com/julienschmidt/httprouter"
+)
+
+type Clusters struct {
+	Service service.Clusters
+}
+
+func (api Clusters) Create(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	c := &model.ClusterRequest{}
+	if err := json.NewDecoder(r.Body).Decode(c); err != nil {
+		http.Error(w, fmt.Sprintf("could not decode body: %s\n", err.Error()), http.StatusBadRequest)
+		return
+	}
+	err := api.Service.Create(c)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "%s\n", err.Error())
+	}
+	w.WriteHeader(http.StatusAccepted)
+	w.Write([]byte("ok\n"))
+}
+
+func (api Clusters) Get(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	cluster, err := api.Service.Get(p.ByName("name"))
+	if err != nil {
+		if err == service.ErrClusterNotFound {
+			w.WriteHeader(http.StatusNotFound)
+		} else {
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+		fmt.Fprintf(w, "%s\n", err.Error())
+		return
+	}
+	// Write content-type, statuscode, payload
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	resp, err := json.MarshalIndent(cluster, "", "    ")
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprint(w, "could not marshall response\n")
+	}
+	fmt.Fprintln(w, string(resp))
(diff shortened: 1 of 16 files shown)


pkg/server/http/handler/clusterers.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/apprenda/kismatic/pkg/server/http/model"
+	"github.com/apprenda/kismatic/pkg/server/http/service"
+	"github.com/julienschmidt/httprouter"
+)
+
+type Clusters struct {
+	service service.Clusters
+}
+
+func NewClusterAPI(s service.Clusters) Clusters {
```
alexbrand: Seems like this isn't really doing anything? Not sure it's necessary.

pkg/server/http/handler/clusterers.go
```diff
@@ -0,0 +1,53 @@
...
+	return Clusters{service: s}
+}
+
+func (api Clusters) Create(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+	c := &model.Cluster{}
+	if err := json.NewDecoder(r.Body).Decode(c); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
```
alexbrand: I think I would consider this a bad request.

This can also be simplified to 
```
http.Error(w, "error.....", http.StatusBadRequest)
```

pkg/server/http/service/clusters.go
```diff
@@ -0,0 +1,81 @@
...
+type clustersServiceWithStore struct {
+	store  store.WatchedStore
+	bucket string
+}
+
+func NewClustersServiceWithStore(store store.WatchedStore, bucket string) Clusters {
+	return clustersServiceWithStore{store: store, bucket: bucket}
+}
+
+func (cs clustersServiceWithStore) Get(name string) ([]byte, error) {
```
alexbrand: I think I would expect this to return an actual type. This is basically dumping what is in the store into the response, right?

We'll probably not want to dump credentials into the response.

pkg/server/http/service/clusters.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+var ErrClusterNotFound = errors.New("cluster details not found in the store")
+
+type Clusters interface {
+	Create(c *model.Cluster) error
+	Get(name string) ([]byte, error)
+}
+
+type clustersServiceWithStore struct {
```
alexbrand: Seems like this could be called clustersService

pkg/server/http/service/clusters.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		return err
+	}
+	if err := cs.store.Put(cs.bucket, c.Name, b); err != nil {
+		return fmt.Errorf("could not persist the cluster in the store: %v", err)
+	}
+	return nil
+}
+
+func ClusterToPlaWrapperBytes(c *model.Cluster) ([]byte, error) {
```
alexbrand: type ClusterToPla

pkg/server/http/service/clusters.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	var p *install.Plan
+	p, err := planner.Read()
+	if err != nil {
+		return nil, fmt.Errorf("could not read plan: %v", err)
+	}
+	// set some defaults in the plan
+	p.Cluster.Name = c.Name
+	pw := store.PlanWrapper{
+		DesiredState: c.State,
```
alexbrand: eventually we'll need to validate the desired state requested by the user

pkg/store/models.go
```diff
@@ -0,0 +1,12 @@
+package store
+
+import "github.com/apprenda/kismatic/pkg/install"
+
+type PlanWrapper struct {
```
alexbrand: we'll figure out a better name for this. Cluster is probably fine.

pkg/server/http/handler/clusterers.go
emmetthitz: Yeah. I think it's idiomatic to use 400 for a "you did something wrong with this", and 500 for a "you did ok, but we messed up".

pkg/store/models.go
emmetthitz: I think plan already wraps Cluster. If anything, I think this should just embed Plan instead of potentially messing with having multiple wrappers tied to a single plan.

alexbrand: I see this as the "Cluster" resource that is defined and stored by the server. A Cluster has an installation plan, and a couple of other things.

Embedding might work, but I think it will be painful on the consumer side that strictly needs a plan struct.

alexbrand:
This looks great! There are some small things we can iterate on later.

Issue #674, opened 2017-07-24
Implement log rotation for all components that KET deploys


alexbrand:
To prevent nodes from filling up their disks, KET should configure log rotation on all the components that are deployed.

We can most likely take advantage of systemd's built-in log rotation capabilities for systemd services. In the case of containerized components, we can use docker's built-in log rotation. This, however, would need more consideration as it would affect workload logs as well.

willvrny:
@alexbrand Would be great to see this added to KET soon. I agree with leveraging the built in log rotation in Docker. We might need this for any Kubernetes components that are logging to the node.

@swade1987 @dkoshkin Do you have any further thoughts on this?

Issue #685, opened 2017-07-26
Have a way to define log-opts for docker daemon


prat0318:
I want log rotation to be enabled for docker daemon so that i don't face disk full issues later. Hence, i want:
```
  "log-opts": {
    "max-size": "20m",
    "max-file": "10"
  }
```
in the `/etc/docker/daemon.json` in addition to the `direct-lvm` settings added by kismatic.

Either these settings should be taken in the `kismatic-cluster.yaml` or have a way to insert `log-opts` to the daemon settings.

prat0318:
Might be related to #674

alexbrand:
Thanks @prat0318. We will tackle this soon.

mrgleeco:
+1 for customzied `daemon.json`.  In addition to altering log type, would like to implement overlay as storage driver.

alexbrand:
As part of this issue, I think we should consider opening up the ability to configure any docker daemon option.

willvrny:
I think the `daemon.json` option is better than passing a flag. The Docker docs state a really good reason for this:

> The daemon includes many configuration options, which you can pass as flags when starting Docker manually, or set in the daemon.json configuration file. The second method is recommended because those configuration changes persist when you restart Docker.

It would be good to provide the `daemon.json` option as default. We currently only do it when direct-lvm is used.

alexbrand:
We won't be able to tackle this as part of the 1.6.0 release. We will look into it for the next one.

PR #903, opened 2017-11-06
Configure docker log rotation

willvrny:
Allows the user to configure the settings in the plan file to fix issue: #674 

```
logs:
    max_size: 10m
    max_file: 1
```

To keep the implementation simpler only two options are exposed with the assumption the user will always use the "json-file" logging driver. This is the driver Kismatic is currently using.

A docker daemon.json file is now always created to persist settings with the minimum content being:

```
"log-driver": "json-file",
```
I've also added a test assertion library to make the unit tests easier to write.


pkg/install/docker_logs_options.go
```diff
@@ -0,0 +1,26 @@
+package install
+
+import (
+	"errors"
+	"regexp"
+	"strings"
+)
+
+func (dl DockerLogs) validate() (bool, []error) {
```
dkoshkin: Any reason not to put this with the other validation code?

pkg/install/execute.go
```diff
@@ -303,10 +303,10 @@ func (ae *ansibleExecutor) RunSmokeTest(p *Plan) error {
 	t := task{
 		name:           "smoketest",
 		playbook:       "smoketest.yaml",
-		explainer:      ae.defaultExplainer(),
 		plan:           *p,
 		inventory:      buildInventoryFromPlan(p),
 		clusterCatalog: *cc,
+		explainer:      ae.defaultExplainer(),
```
dkoshkin: Was this moved by your IDE or did you do this intentionally?

pkg/install/plan.go
```diff
@@ -289,6 +298,10 @@ func buildPlanFromTemplateOptions(templateOpts PlanTemplateOptions) Plan {
 	p.Cluster.Certificates.Expiry = "17520h"
 	p.Cluster.Certificates.CAExpiry = defaultCAExpiry
 
+	// Set docker log option defaults
+	p.Docker.Logs.MaxSize = "10m"
```
dkoshkin: Is there any supporting material why `10m` is the default?

pkg/install/execute.go
```diff
@@ -691,6 +691,10 @@ func (ae *ansibleExecutor) buildClusterCatalog(p *Plan) (*ansible.ClusterCatalog
 	}
 
 	// Setup docker options
+	cc.DockerLogOptsEnabled = p.Docker.Logs.MaxSize != "-1"
```
dkoshkin: I just tested passing `"-1"` to the `"max-size"` option and it worked fine, I think you can get rid of of this and the check in ansible `daemon.json` file and just pass these options directly since the defaults will always be set.
```
root@master001:/etc/docker# cat daemon.json
{
      "log-driver": "json-file",
    "log-opts": {
      "max-size": "-1",
      "max-file": "1"
    }
  }
```

pkg/install/plan_types.go
```diff
@@ -212,10 +212,20 @@ type CloudProvider struct {
...
 // Docker includes the configuration for the docker installation owned by KET.
 type Docker struct {
+	// Log configuration for the docker engine
+	Logs DockerLogs
 	// Storage configuration for the docker engine
 	Storage DockerStorage
 }
 
+// DockerLogs includes the log-specific configuration for docker.
+type DockerLogs struct {
```
dkoshkin: I wonder if the plan file should look like:
```
docker:
  logs:
    log_driver:
    log_opts
      map[string]string
```
This has the advantage of the user specifying their own settings and log-driver.
We would still set the default values for them.

In code this would actually be pretty nice, `Logs` can be a pointer and in `setDefaults()` you can just check for nil and set the defaults.

glide.lock
```diff
@@ -144,6 +144,10 @@ imports:
   - doc
 - name: github.com/spf13/pflag
   version: e57e3eeb33f795204c1ca35f56c44f83227c6e66
+- name: github.com/stretchr/testify
```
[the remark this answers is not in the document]
willvrny: Yes sorry I can do this.

pkg/install/docker_logs_options.go
willvrny: I noticed that some of the validation code was in a separate file. Having it all in the same file would make the test file quite large. Pulling it out allowed me to create a test file just for this validation. It has a few scenarios.

pkg/install/execute.go
willvrny: I did this intentionally for consistency with tasks in the file.

pkg/install/execute.go
willvrny: Okay I'll make the change. Thanks for testing this.

pkg/install/plan.go
willvrny: No 10m felt like a good default to have. I'm open to suggestions.

pkg/install/plan_types.go
willvrny: How would we handle validation in this scenario? I like the flexibility but I'm not sure if we could offer validation on the log options. We'd have to know all the options Docker provides for each driver and then check the user has not given any invalid options.

dkoshkin: We can start validating the `json-file` since thats what we set by default.

willvrny: This could couple us tightly to docker changes as when we change a version we'd need to check for new or deprecated log opts. Should we leave out the validation or perform minimal validation?

glide.lock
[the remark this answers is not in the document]
willvrny: I've made this change.

pkg/install/execute.go
willvrny: I've made this change as well.

willvrny:
@dkoshkin I've made some changes as suggested. Is it possible to get this merged as a first step and being more flexible regarding logs options could be a future pull request?

pkg/install/plan_types.go
dkoshkin: I see this as more of an advance feature, I think its fine to to validate.

swade1987:
Fixes #685

dkoshkin:
I think this PR should have the change to support different options.

swade1987:
@prat0318 would the changes @willvrny meet your requirements?

willvrny:
@alexbrand @dkoshkin Could we add this to the 1.7 milestone? It would be great to get this in the next release.

I may not have the bandwidth to be able to work on it but I'm more than happy for someone else to continue with it?

swade1987:
Please note that we are having to make these changes manually to every node on customer site which is both time-consuming and laborious.

pkg/install/docker_logs_options.go
emmetthitz: I think this is my only gripe as well. Not having all the validation work together makes it less maintainable.

emmetthitz: Plus, moving it should trigger CI now, and I'd be curious to see if this passes.

swade1987: I feel having the validation/testing code for each feature in a separate code file makes sense from an extensibility point of you. Having everything in a single file makes it unmanageable

emmetthitz: When we get there, sure. I think that probably makes sense as well. In the mean time, segregating validation for only new features is probably going to cause more confusion than it solves.

swade1987: What harm is having it as a separate file doing? To me, it's starting a good practice and removing re-work down the line. Also, moving this back to a single file is not really adding much.

emmetthitz: I guess. We're arguing mostly semantics. It's preference for consistency over more work in my case, and vice versa in your case. This isn't a huge deal, and I'd be cool merging this as is - I'm mostly looking for an excuse to trigger CI.

swade1987:
@alexbrand @dkoshkin @emmetthitz guys what is left to do to get this merged in for the 1.7.0 release?

emmetthitz:
Ideally I'd like to get this checked by CI. @willvrny can you move these changes to a branch and re-open the PR?

dkoshkin:
Opened a new PR in #977. Thanks @willvrny for the initial work.

PR #904, opened 2017-11-06
Upgrade tests to use RHEL 7.4

dkoshkin:


swade1987:
Should we not be testing both RHEL 7.3 and RHEL 7.4?

emmetthitz:
@swade1987 we actually talked about this today during standup. It's a non-trivial amount of work to test both (as well as adding a significant amount of build/test time for a relatively little gain). What version do most clients seem to use? We'd like to test against what's most popular for obvious reasons.

swade1987:
@emmetthitz clients use a mixture of RHEL 7.3 and 7.4.

We can't force clients to upgrade to 7.4 just because we can only test one distribution version, this seems like a limitation which needs to be resolved in our CI/CD toolchain. We should really be testing an N and N-1 versions of our supported OS's

alexbrand:
It indeed is a limitation that we have been aware of for quite some time. Running the entire test suite against multiple distros is not trivial right now, both from a $ perspective and tooling perspective.

swade1987:
In this case, we need to be extremely explicit about the version of OS we support moving forward.
We need to state RHEL 7.4 and that 7.3 is out of scope, and the same with Ubuntu.

PR #905, opened 2017-11-07
Store fixes

dkoshkin:
Cleaned up some store code, mostly simplified and extracted some common code that both the controller and the server shared.


pkg/store/boltdb.go
```diff
@@ -45,7 +47,7 @@ type WatchResponse struct {
 	Value []byte
 }
 
-func NewBoltDB(path string, mode os.FileMode, logger *log.Logger) (WatchedStore, error) {
+func DefaultStore(path string, mode os.FileMode, logger *log.Logger) (WatchedStore, error) {
```
alexbrand: Any reason for not calling this `New`?

pkg/store/watch_manager.go
```diff
@@ -23,57 +23,61 @@ type watchMgr struct {
...
+					}
+				}
+			default:
+				mgr.logger.Printf("unknown message sent to watch mgr: %T\n", msg)
 			}
-
-		default:
-			mgr.logger.Printf("unknown message sent to watch mgr: %T\n", msg)
+		case <-ctx.Done():
+			return
```
alexbrand: Is there anything that needs to be done before returning here? Closing channels, cleaning up anything?

pkg/store/cluster.go
```diff
@@ -0,0 +1,77 @@
...
+// so that the clients don't need to worry about the bucket
+// or marshaling/unmarshaling
+type ClusterStore interface {
+	Get(key string) (*Cluster, error)
+	Put(key string, cluster Cluster) error
+	GetAll() (map[string]Cluster, error)
+	Watch(ctx context.Context, buffer uint) <-chan WatchResponse
+}
+
+type CS struct {
```
alexbrand: Given that this is exported, it should have a more meaningful name. We can punt on this for now, but I think we'll need to review the `store` package's API.

pkg/store/watch_manager.go
dkoshkin: True, the `mailbox` chan needs to be closed.

As for the watch channels, it needs to be up to the client to cancel their context and end reading from that channel, if the manager closes it the client could have a nasty side effect of getting nils from the channel and interpreting it as an event.

alexbrand:
If clients cancel watches after the store is closed, the program will panic. Also, what happens if new watches are issued on a closed store?

It seems like we might need a way for clients to know when the store gets closed. We can work on this in a follow up issue, as it needs more thought.

PR #909, opened 2017-11-07
Use a separate executor for each cluster

### LINKED ISSUE #907: Use separate executors for each cluster controller
The multi cluster controller should take in a func that creates executors that are cluster specific. This way, each cluster controller has it's own executor that logs to a specific place, etc.

alexbrand:
Fixes #907 

Each cluster needs it's own directory structure separate from other clusters. This directory will include the runs dir, the generated assets dir, and a log file where all logs will be stored. It might make sense to have a log dir where multiple log files are created, but that will require more work.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -22,10 +22,10 @@ type ClusterController interface {
 type ExecutorCreator func(clusterName string) (install.Executor, error)
 
 // New returns a cluster controller
-func New(l *log.Logger, execCreater ExecutorCreator, cs store.ClusterStore, reconFreq time.Duration) ClusterController {
+func New(l *log.Logger, execCreator ExecutorCreator, cs store.ClusterStore, reconFreq time.Duration) ClusterController {
 	return &multiClusterController{
 		log:                l,
-		newExecutor:        execCreater,
+		newExecutor:        execCreator,
 		clusterStore:       cs,
 		reconcileFreq:      reconFreq,
 		clusterControllers: make(map[string]chan<- struct{}),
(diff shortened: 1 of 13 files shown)


pkg/controller/controller.go
```diff
@@ -14,14 +17,52 @@ type ClusterController interface {
...
+			return nil, fmt.Errorf("error creating directories for executor: %v", err)
+		}
+		logFile, err := os.Create(filepath.Join(rootDir, clusterName, "kismatic.log"))
+		if err != nil {
+			return nil, fmt.Errorf("error creating log file for executor: %v", err)
+		}
+		executorOpts := install.ExecutorOptions{
+			GeneratedAssetsDirectory: filepath.Join(rootDir, clusterName, "generated"),
+			RunsDirectory:            filepath.Join(rootDir, clusterName, "runs"),
+			RestartServices:          true,
```
dkoshkin: `RestartServices` feels like it should not be part of the executor struct but be passed in into the methods.
Also does this need to be `true`?

pkg/controller/controller.go
```diff
@@ -14,14 +17,52 @@ type ClusterController interface {
 	Run(ctx context.Context)
 }
 
+// ExecutorCreater creates executors that can be used for executing actions
+// against a specific cluster.
+type ExecutorCreater func(clusterName string) (install.Executor, error)
```
dkoshkin: Should be "Creator*"

alexbrand: Thanks!

pkg/controller/controller.go
alexbrand: Yeah, I can definitely see removing RestartServices from the options struct. Will look into that.

I was thinking of failure scenarios when I set it to true, but I am not sure I have a strong enough case for it. What are your thoughts?

alexbrand:
I made the restart services change in a separate commit. I am not sure I like it, but I think it's as far as it can go without a major refactor.

dkoshkin:
LGTM!

emmetthitz:
LGTM as well

PR #913, opened 2017-11-09
Added more API endpoints

dkoshkin:
Added 4 new endpoints:
```
GET "/clusters"  #returns all clusters
DELETE "/clusters/:name"
GET "/clusters/:name/kubeconfig"
GET "/clusters/:name/logs"
GET "/clusters/:name/assets"
```

Sample POST request
```
{
  "name": "foo",
  "desiredState": "installed",
  "etcdCount": 1,
  "masterCount": 1,
  "workerCount": 1,
  "ingressCount": 0,
  "provisioner": {
    "provider": "aws",
    "options": {
      "accessKeyID": "__AWS_ACCESS_KEY_ID__",
      "secretAccessKey": "__AWS_SECRET_ACCESS_KEY__"
    }
  }
}
```
Sample response
```
{
  "name": "foo",
  "desiredState": "installed",
  "currentState": "provisionFailed",
  "clusterIP": "",
  "etcdCount": 1,
  "masterCount": 1,
  "workerCount": 1,
  "ingressCount": 0,
  "provisioner": {
    "provider": "aws"
  }
}
```

### DIFF
--- a/pkg/cli/server.go
+++ b/pkg/cli/server.go
@@ -80,7 +80,7 @@ func doServer(stdout io.Writer, options serverOptions) error {
 	// Needs to be an absoulute path for the HTTP server
 	pwd, err := os.Getwd()
 	if err != nil {
-		logger.Fatalf("Could not get current directory for assets")
+		logger.Fatalf("Could not get current directory for assets: %v", err)
 	}
 	assetsDir := path.Join(pwd, assetsFolder)
 	err = os.MkdirAll(assetsDir, 0700)
@@ -89,7 +89,7 @@ func doServer(stdout io.Writer, options serverOptions) error {
 	}
 
 	// create handlers
-	clusterAPI := handler.Clusters{Store: clusterStore, AssetsDir: assetsDir}
+	clusterAPI := handler.Clusters{Store: clusterStore, AssetsDir: assetsDir, Logger: logger}
 
 	// Setup the HTTP server
 	server := http.HttpServer{
--- a/pkg/cli/server.go
+++ b/pkg/cli/server.go
@@ -7,6 +7,7 @@ import (
 	nethttp "net/http"
 	"os"
 	"os/signal"
+	"path"
 	"syscall"
 	"time"
 
@@ -22,6 +23,7 @@ const (
 	loggerPrefix   = "[kismatic] "
 	defaultTimeout = 10 * time.Second
 	clustersBucket = "kismatic"
+	assetsFolder   = "assets"
 )
 
 type serverOptions struct {
@@ -63,10 +65,10 @@ func doServer(stdout io.Writer, options serverOptions) error {
 
 	// Create the store
 	s, err := store.New(options.dbFile, 0600, logger)
-	defer s.Close()
 	if err != nil {
 		logger.Fatalf("Error creating store: %v", err)
 	}
+	defer s.Close()
 	err = s.CreateBucket(clustersBucket)
 	if err != nil {
 		logger.Fatalf("Error creating bucket in store: %v", err)
@@ -74,8 +76,20 @@ func doServer(stdout io.Writer, options serverOptions) error {
 
 	clusterStore := store.NewClusterStore(s, clustersBucket)
 
+	// Create a dir where all the controller-related files will be stored
+	// Needs to be an absoulute path for the HTTP server
+	pwd, err := os.Getwd()
+	if err != nil {
+		logger.Fatalf("Could not get current directory for assets")
+	}
+	assetsDir := path.Join(pwd, assetsFolder)
+	err = os.MkdirAll(assetsDir, 0700)
+	if err != nil {
+		logger.Fatalf("Error creating assets directory %q: %v", assetsDir, err)
+	}
+
 	// create handlers
-	clusterAPI := handler.Clusters{Store: clusterStore}
+	clusterAPI := handler.Clusters{Store: clusterStore, AssetsDir: assetsDir}
 
 	// Setup the HTTP server
 	server := http.HttpServer{
@@ -103,15 +117,9 @@ func doServer(stdout io.Writer, options serverOptions) error {
 		BinaryPath: "./../../bin/terraform",
 	}
 
-	// Create a dir where all the controller-related files will be stored
-	rootDir := "server"
-	err = os.MkdirAll(rootDir, 0700)
-	if err != nil {
-		logger.Fatalf("error creating directory %q: %v", rootDir, err)
-	}
 	ctrl := controller.New(
 		logger,
-		controller.DefaultExecutorCreator(rootDir),
+		controller.DefaultExecutorCreator(assetsDir),
 		controller.DefaultProvisionerCreator(terraform),
 		clusterStore,
 		10*time.Minute,
--- a/pkg/server/http/handler/clusters.go
+++ b/pkg/server/http/handler/clusters.go
@@ -71,6 +71,34 @@ func (api Clusters) Get(w http.ResponseWriter, r *http.Request, p httprouter.Par
 	fmt.Fprintln(w, string(resp))
 }
 
+func (api Clusters) GetAll(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	clustersResp, err := getAllFromStore(api.Store)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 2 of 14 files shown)


pkg/server/http/handler/clusters.go
```diff
@@ -71,6 +74,54 @@ func (api Clusters) Get(w http.ResponseWriter, r *http.Request, p httprouter.Par
...
+	if err := api.Store.Delete(id); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "could not delete from the store: %v", err)
+	}
+	w.WriteHeader(http.StatusAccepted)
+	w.Write([]byte("ok\n"))
+}
+
+// GetKubeconfig will return the kubeconfig file for a cluster :name
+// A 404 is returned if a file is not found
```
alexbrand: Hmmmm. I am torn on this.

I think a 404 should be returned if there is no cluster defined in the API with the given name. With that said, that does open the question of what to return if the cluster is defined but the file hasn't been generated yet.

I think a 500 makes sense if the cluster is in a state where the files _should_ exist, and we find that the file is not there.

pkg/server/http/handler/clusters.go
```diff
@@ -71,6 +74,54 @@ func (api Clusters) Get(w http.ResponseWriter, r *http.Request, p httprouter.Par
...
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "%s\n", err.Error())
+		return
+	}
+	// Write content-type, statuscode, payload
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	resp, err := json.MarshalIndent(clustersResp, "", "    ")
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
```
alexbrand: What are your thoughts on using https://golang.org/pkg/net/http/#Error throughout the handler funcs to handle errors?

pkg/server/http/handler/clusters.go
```diff
@@ -71,6 +74,54 @@ func (api Clusters) Get(w http.ResponseWriter, r *http.Request, p httprouter.Par
...
+		fmt.Fprint(w, "could not marshall response\n")
+	}
+	fmt.Fprintln(w, string(resp))
+}
+
+func (api Clusters) Delete(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	id := p.ByName("name")
+	if err := api.Store.Delete(id); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "could not delete from the store: %v", err)
```
alexbrand: We should define our stance on server-side errors. What I have usually seen is that the error is not leaked to the client, but instead it gets logged on the server side.

pkg/server/http/handler/clusters.go
```diff
@@ -71,6 +74,54 @@ func (api Clusters) Get(w http.ResponseWriter, r *http.Request, p httprouter.Par
...
 
+func (api Clusters) GetAll(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	clustersResp, err := getAllFromStore(api.Store)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "%s\n", err.Error())
+		return
+	}
+	// Write content-type, statuscode, payload
+	w.Header().Set("Content-Type", "application/json")
```
alexbrand: Might make sense to set this in middleware, but not sure if that complicates things.

pkg/cli/server.go
```diff
@@ -62,19 +64,31 @@ func doServer(stdout io.Writer, options serverOptions) error {
...
 		logger.Fatalf("Error creating bucket in store: %v", err)
 	}
 
 	clusterStore := store.NewClusterStore(s, clustersBucket)
 
+	// Create a dir where all the controller-related files will be stored
+	// Needs to be an absoulute path for the HTTP server
+	pwd, err := os.Getwd()
+	if err != nil {
+		logger.Fatalf("Could not get current directory for assets")
```
alexbrand: This probably won't happen, but can we log the error here as well?

pkg/server/http/handler/clusters.go
```diff
@@ -4,35 +4,157 @@ import (
...
 type ClusterRequest struct {
-	Name         string
-	DesiredState string
-	AwsID        string
-	AwsKey       string
-	Etcd         int
-	Master       int
-	Worker       int
+	Name         string      `json:"name"`
+	DesiredState string      `json:"desired-state"`
```
alexbrand: I think camelCase makes more sense here... at least that's what I have seen used, including Kuberenetes API.

Was there a reason for going kebab-case?

pkg/server/http/handler/clusters.go
```diff
@@ -57,26 +204,118 @@ func (api Clusters) Get(w http.ResponseWriter, r *http.Request, p httprouter.Par
...
+		return
+	}
+	f := path.Join(api.AssetsDir, id, "kismatic.log")
+	if stat, err := os.Stat(f); os.IsNotExist(err) || stat.IsDir() {
+		w.WriteHeader(http.StatusInternalServerError)
+		api.Logger.Println(errorf("logs for cluster %s could not be retrieved: %v", id, err))
+		return
+	}
+	// set so the browser downloads it instead of displaying it
+	w.Header().Set("Content-Disposition", "attachment; filename=kismatic.log")
```
alexbrand: Logs should be fine to display instead of forcing download

pkg/server/http/handler/clusters.go
dkoshkin: There wasn't a strong consensus when I looked, but if the k8s API uses 🐍 I'll be consistent.

pkg/server/http/handler/clusters.go
```diff
@@ -4,35 +4,157 @@ import (
...
+}
+
+type Provisioner struct {
+	// Options: aws
+	Provider   string
+	AWSOptions *AWSProvisionerOptions `json:"options,omitempty"`
+}
+
+type Cluster struct {
+	ClusterIP    string `json:"cluster_ip"`
```
alexbrand: Looks like this isn't used?

pkg/server/http/handler/clusters.go
```diff
@@ -112,11 +366,46 @@ func getFromStore(name string, cs store.ClusterStore) (*ClusterResponse, error)
...
+	if msc == nil {
+		return resp, nil
+	}
+	for key, sc := range msc {
+		resp = append(resp, *buildResponse(key, &sc))
+	}
+	return resp, nil
+}
+
+func buildResponse(name string, sc *store.Cluster) *ClusterResponse {
```
alexbrand: Are pointers needed here?

pkg/server/http/handler/clusters.go
dkoshkin: Using it now.

pkg/server/http/handler/clusters.go
```diff
@@ -4,35 +4,160 @@ import (
...
 type ClusterResponse struct {
-	Name         string
-	DesiredState string
-	CurrentState string
-	install.Plan
+	Name         string      `json:"name"`
+	DesiredState string      `json:"desiredState"`
+	CurrentState string      `json:"currentState"`
+	Provisioner  Provisioner `json:"provisioner"`
+	Cluster      Cluster     `json:"cluster"`
```
alexbrand: I wonder if we really need this. Makes sense from a consistency perspective (provisioner vs cluster), but at the same time, the REST resources we are creating are "Clusters", so it's weird to have a Cluster field inside a Cluster resource.

pkg/server/http/handler/clusters.go
```diff
@@ -57,26 +207,158 @@ func (api Clusters) Get(w http.ResponseWriter, r *http.Request, p httprouter.Par
...
+		return
+	}
+	// archive the directory
+	err = archiver.TarGz.Make(tmpf.Name(), []string{dir})
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		api.Logger.Println(errorf("could not archive tge assets file for cluster %s: %v", id, err))
+		return
+	}
+	attachmentName := fmt.Sprintf("attachment; filename=%s-assets.tar.gz", id)
```
alexbrand: Very nice!

PR #915, opened 2017-11-10
Basic terraform vendoring and scripts. No automated variable generati…

emmetthitz:
…on yet.

So @alexbrand can start on the controller integration.


alexbrand:
Thanks @emmetthitz 

Merging as we are both blocked by this, even though CI is failing.

PR #916, opened 2017-11-13
Controller provision

alexbrand:
Opening this PR to unblock other work that is dependent on this. 

The provision pkg needs some love, and there are some ideas I have there, but I think it can be done in a follow up PR.


pkg/controller/cluster.go
```diff
@@ -60,88 +62,119 @@ func (c *clusterController) reconcile(clusterName string, cluster store.Cluster)
...
+// transition performs an action to take the cluster to the next state. The
+// action to be performed depends on the current state and the desired state.
+// Once the action is done, an updated cluster spec is returned that reflects
+// the outcome of the action.
+func (c *clusterController) transition(cluster store.Cluster) store.Cluster {
+	if cluster.CurrentState == cluster.DesiredState {
+		panic("cannot transition a cluster that is already at it's desired state")
+	}
 	switch cluster.CurrentState {
 	case planned:
```
dkoshkin: Can we add an `""` case.

pkg/controller/cluster.go
```diff
@@ -60,88 +62,119 @@ func (c *clusterController) reconcile(clusterName string, cluster store.Cluster)
...
 	err = c.executor.GenerateKubeconfig(plan)
 	if err != nil {
 		c.log.Printf("error generating kubeconfig file: %v", err)
-		return installFailed, false
+		cluster.CurrentState = installFailed
+		cluster.CanContinue = false
+		return cluster
 	}
 
 	err = c.executor.Install(&plan, true)
```
dkoshkin: Need to validate before install.

pkg/provision/terraform.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	defer os.Chdir("../../../")
+
+	// Setup the environment for all Terraform commands.
+	cmdEnv := append(os.Environ(), aws.getCommandEnvironment()...)
+
+	providerDir := fmt.Sprintf("../../providers/%s", plan.Provisioner.Provider)
+
+	// Terraform init
+	initCmd := exec.Command(terraformBinaryPath, "init", providerDir)
```
dkoshkin: You can set the directory to run the command from in the `cmd`, this would simplify all of the directory changes.

pkg/controller/cluster.go
alexbrand: Talked about this offline. We arrived at the conclusion that it doesn't really make sense to have clusters in the store that have no current state. Instead, the server will set the currentState to `planned` when the cluster spec is first created.

dkoshkin:
LGTM!

PR #921, opened 2017-11-16
Fix typos in docs and user-facing msgs

alexbrand:


dkoshkin:
LGTM!
Did you search the repo with a spellchecker or did you just find these, wondering if there are others?

PR #924, opened 2017-11-20
Update /clusters DELETE to modify the state

dkoshkin:
Split up the functions that interact with the store for reusability. 
 
Also made `install.AWSProvisionerOptions` not a pointer.


pkg/server/http/handler/clusters.go
```diff
@@ -220,13 +228,17 @@ func (api Clusters) Get(w http.ResponseWriter, r *http.Request, p httprouter.Par
...
 func (api Clusters) GetAll(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
-	clustersResp, err := getAllFromStore(api.Store)
+	fromStore, err := getAllFromStore(api.Store)
 	if err != nil {
 		w.WriteHeader(http.StatusInternalServerError)
 		api.Logger.Println(errorf(err.Error()))
 		return
 	}
 
+	clustersResp := make([]ClusterResponse, 0)
```
alexbrand: Does it make sense to set this to len(fromStore)?

PR #925, opened 2017-11-21
Terraform AWS multicluster implementation complete. Lots of test  fix…

emmetthitz:
…es. CLI improvements to include provisioner information.


docs/plan-file-reference.md
```diff
@@ -863,7 +874,7 @@
 
 ###  master.expected_count
 
- Number of master nodes that are part of the cluster. 
+ Number of nodes.
```
emmetthitz: Oh hell. Need to clean this up some more.

pkg/install/pki_test.go
```diff
@@ -246,7 +246,7 @@ func TestGenerateClusterCAPlanFileExpirationIsRespected(t *testing.T) {
 	}
 
 	expirationDate := time.Now().Add(validity)
-	if caCert.NotAfter.Year() != expirationDate.Year() || caCert.NotAfter.YearDay() != expirationDate.YearDay() {
+	if caCert.NotAfter.YearDay() < expirationDate.YearDay() {
```
emmetthitz: These were failing despite the fact that the `caCert.NotAfter` was not after the `expirationDate`, just not equal to it.

pkg/install/plan.go
```diff
@@ -297,14 +299,21 @@ func WritePlanTemplate(planTemplateOpts PlanTemplateOptions, w PlanReadWriter) e
...
+
+	p.Cluster.Name = templateOpts.ClusterName
 	p.Cluster.AdminPassword = templateOpts.AdminPassword
 	p.Cluster.DisablePackageInstallation = false
 	p.Cluster.DisconnectedInstallation = false
 
 	// Set SSH defaults
-	p.Cluster.SSH.User = "kismaticuser"
-	p.Cluster.SSH.Key = sshKey()
+	p.Cluster.SSH.User = ""
```
emmetthitz: @dkoshkin had a concern about changing this to be an empty default, and I think it makes sense (even in the bare metal, already-provisioned cases). The user should still consciously have to provide us with an ssh key and user for their already-existing machines.

pkg/install/plan_types.go
```diff
@@ -438,7 +441,7 @@ type DeprecatedPackageManager struct {
 
 // MasterNodeGroup is the collection of master nodes
 type MasterNodeGroup struct {
-	// Number of master nodes that are part of the cluster.
+	// Number of nodes.
```
emmetthitz: Will fix this and regenerate the docs. Oops.

pkg/install/pki_test.go
dkoshkin: Yes! This was the test that would fail at 10 or 11 AM.

pkg/install/plan_types.go
```diff
@@ -465,7 +468,14 @@ type NodeGroup struct {
 }
 
 // An OptionalNodeGroup is a collection of nodes that can be empty
-type OptionalNodeGroup NodeGroup
```
dkoshkin: Does this need to change?

pkg/install/validate.go
```diff
@@ -156,6 +156,33 @@ func (p *Plan) validate() (bool, []error) {
 	return v.valid()
 }
 
+func (p *Provisioner) validate() (bool, []error) {
+	awsRegionsRegexp := `(us-(east|west)-(1|2))|
```
dkoshkin: I believe we agreed on not validating the region as this may change somewhat frequently.

pkg/cli/plan_test.go
```diff
@@ -1,103 +1,103 @@
 package cli
 
-import (
-	"bytes"
-	"io"
-	"strings"
-	"testing"
-)
+// import (
```
alexbrand: Why are these tests commented out?

pkg/install/pki_test.go
alexbrand: This is changing the assertion to one that I am not sure is correct. We want the requested expiration date to be the expiration date of the certificate.

pkg/install/plan.go
alexbrand: I _think_ the reasoning behind this change makes sense. Thoughts @dkoshkin?

If we are to do this, I would remove the `sshKey()` func if it's not being used anywhere else.

pkg/install/plan_types.go
alexbrand: I am also wondering why this was needed.

pkg/install/upgrade_test.go
```diff
@@ -138,8 +138,7 @@ func TestDetectNodeUpgradeSafetyMasterCountUnsafe(t *testing.T) {
 func TestDetectNodeUpgradeSafetyMasterLoadBalancingUnsafe(t *testing.T) {
 	plan := Plan{
 		Master: MasterNodeGroup{
-			ExpectedCount:    2,
-			LoadBalancedFQDN: "foo",
+			ExpectedCount: 2,
```
alexbrand: What was the reason for changing this file? Was there a change made that affected upgrades?

pkg/install/validate.go
alexbrand: Yeah... Let's not tie ourselves to specific AWS regions. Otherwise we'll have to re-release whenever a new region is added.

pkg/provision/aws.go
```diff
@@ -0,0 +1,187 @@
...
+	if err := ssh.NewKeyPair(pubKeyPath, privKeyPath); err != nil {
+		return nil, fmt.Errorf("error generating SSH key pair: %v", err)
+	}
+	plan.Cluster.SSH.Key = privKeyPath
+
+	// Write out the terraform variables
+	data := AWSTerraformData{
+		Region:            plan.Provisioner.AWSOptions.Region,
+		ClusterName:       plan.Cluster.Name,
+		MasterCount:       len(plan.Master.Nodes),
```
alexbrand: This and subsequent counts should be set to the `ExpectedCount` field, instead of the len of the slice of nodes.

pkg/install/plan.go
dkoshkin: Yep, I like this change. +1 to remove `sshKey()`

pkg/provision/aws.go
```diff
@@ -0,0 +1,187 @@
...
+	tfNodes, err := aws.getTerraformNodes(plan.Cluster.Name, "master")
+	if err != nil {
+		return nil, err
+	}
+	masterNodes := nodeGroupFromSlices(tfNodes.IPs, tfNodes.InternalIPs, tfNodes.Hosts)
+	mng := install.MasterNodeGroup{
+		ExpectedCount: masterNodes.ExpectedCount,
+		Nodes:         masterNodes.Nodes,
+	}
+	mng.LoadBalancedFQDN = tfNodes.InternalIPs[0]
```
alexbrand: This should be the public IP. Otherwise the installation will fail when trying to install Helm.

pkg/cli/plan.go
```diff
@@ -19,16 +20,30 @@ func NewCmdPlan(in io.Reader, out io.Writer, options *installOpts) *cobra.Comman
...
-			return doPlan(in, out, planner, options.planFilename)
+			return doPlan(in, out, planner)
 		},
 	}
 
 	return cmd
 }
 
-func doPlan(in io.Reader, out io.Writer, planner install.Planner, planFile string) error {
+func doPlan(in io.Reader, out io.Writer, planner *install.FilePlanner) error {
```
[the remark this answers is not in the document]
emmetthitz: I guess not, no.

pkg/install/validate.go
emmetthitz: Sure. I'll ice it.

pkg/cli/plan.go
```diff
@@ -88,17 +105,28 @@ func doPlan(in io.Reader, out io.Writer, planner install.Planner, planFile strin
...
+		WorkerNodes:               workerNodes,
+		IngressNodes:              ingressNodes,
+		StorageNodes:              storageNodes,
+		NFSVolumes:                nfsVolumes,
+	}
+	if provisioner != "" {
+		//If a provider is given,
+		//write to the terraform state location
+		dir := fmt.Sprintf("terraform/clusters/%s", planTemplate.ClusterName)
+		if err := os.MkdirAll(dir, 0700); err != nil {
```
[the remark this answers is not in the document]
emmetthitz: Yeah. I can see the concern here. I did this mostly with regards to having more seamless UX in the cases where they are using plan. Having to copy/move your plan file to the state location just felt bad. 

The directory does also get generated if it doesn't exist during provisioning, so it should cover the case you're describing.

pkg/install/plan_types.go
emmetthitz: Oops 👍

pkg/cli/plan_test.go
emmetthitz: The changes to `install plan` horribly broke some of the tests that were being run here. I commented this out in the interest of time, but they should be re-implemented.

pkg/install/upgrade_test.go
emmetthitz: I had to do a revert by hand after changing branches, and this probably just got past. Thanks.

pkg/install/pki_test.go
emmetthitz: Yeah, and I think it's some janky rounding that happens internal to the library where the requested expiration wouldn't be equal to the cert expiration. It would still be "NotAfter" which probably why that field is named the way it is. This test still fails the way you expect it to in the cases where `NotAfter` is greater than the requested expiration.

pkg/provision/aws.go
```diff
@@ -162,7 +165,12 @@ func (aws AWS) Destroy(clusterName string) error {
 	cmd.Stdout = aws.Terraform.Output
 	cmd.Stderr = aws.Terraform.Output
 	cmd.Env = aws.getCommandEnvironment()
-	cmd.Dir = aws.getClusterStateDir(clusterName)
+	fmt.Println(cmd.Env)
```
dkoshkin: Remove

pkg/provision/aws.go
```diff
@@ -22,13 +22,16 @@ func (aws AWS) getCommandEnvironment() []string {
...
+	clusterStateDir, err := aws.getClusterStateDir(plan.Cluster.Name)
+	if err != nil {
+		return nil, err
+	}
 	if err := os.MkdirAll(clusterStateDir, 0700); err != nil {
 		return nil, fmt.Errorf("error creating directory to keep cluster state: %v", err)
 	}
 
 	// Setup the environment for all Terraform commands.
-	cmdEnv := append(os.Environ(), aws.getCommandEnvironment()...)
```
dkoshkin: Didnt we do this because we need to get the ENVs from the users PATH and then append the AWS ones?

pkg/cli/provision.go
```diff
@@ -70,10 +76,15 @@ func NewCmdDestroy(in io.Reader, out io.Writer, opts *installOpts) *cobra.Comman
 				Output:     out,
 				BinaryPath: filepath.Join(path, "terraform/bin/terraform"),
 			}
-			fmt.Println(plan.Provisioner.Provider)
 			switch plan.Provisioner.Provider {
 			case "aws":
-				aws := provision.AWS{Terraform: tf}
+				access := os.Getenv("AWS_ACCESS_KEY_ID")
+				secret := os.Getenv("AWS_SECRET_ACCESS_KEY")
+				aws := provision.AWS{
```
dkoshkin: Why are we getting the vars here and in `pkg/provision/aws.go` ?

pkg/provision/aws.go
emmetthitz: Oops.

pkg/cli/provision.go
dkoshkin: Oh wait this is correct for the CLI.

emmetthitz: `aws.getCommandEnvironment()` doesn't actually get them from the environment, it gets them from the `AWS` struct.

pkg/provision/aws.go
dkoshkin: Hmm I wonder if its messing up when those vars are already set in your path, maybe we need to check for that.

pkg/cli/plan_test.go
alexbrand: Thanks! Turns out that these tests have been broken for some time.

dkoshkin:
LGTM!

PR #927, opened 2017-11-21
Added /clusters PUT endpoint

dkoshkin:


pkg/server/http/handler/clusters.go
```diff
@@ -102,6 +102,42 @@ func (p *Provisioner) validate() (bool, []error) {
 	return v.valid()
 }
 
+type clusterPatch struct {
```
alexbrand: I think calling this `patch` might be confusing, as this is really validating a PUT request?

pkg/server/http/handler/clusters.go
```diff
@@ -400,13 +494,14 @@ func getAllFromStore(cs store.ClusterStore) (map[string]store.Cluster, error) {
...
-		WorkerNodes:  req.WorkerCount,
-		IngressNodes: req.IngressCount,
+		AdminPassword: password,
+		EtcdNodes:     req.EtcdCount,
+		MasterNodes:   req.MasterCount,
+		WorkerNodes:   req.WorkerCount,
+		IngressNodes:  req.IngressCount,
 	}
 	planner := &install.BytesPlanner{}
 	if err := install.WritePlanTemplate(planTemplate, planner); err != nil {
```
alexbrand: I am a little worried about creating a new plan on every PUT request, especially in the case that we update this func with new defaults in a new KET release. It might have unintended consequences for existing clusters, but maybe I am being too paranoid?

dkoshkin: I think its too early to tell at this point, lets keep this for now and then if we see it a problem in the future when updating the plan struct we can think of a better approach.

alexbrand: Sounds good. Just heads up we might have to rework some of this as a result of #962

alexbrand:
LGTM!

PR #930, opened 2017-11-22
Don't set ingress/storage nodes unless requested

alexbrand:
Looks like this slipped through after the refactor.


emmetthitz:
LGTM

PR #931, opened 2017-11-22
Cluster controller: Add support for destroy

alexbrand:
Hooked up the destroy functionality on the controller. We can now destroy clusters using the DELETE request.

### DIFF
--- a/pkg/controller/cluster.go
+++ b/pkg/controller/cluster.go
@@ -19,38 +19,55 @@ const (
 	modifying       = "modifying"
 	modifyFailed    = "modifyFailed"
 	destroying      = "destroying"
+	destroyFailed   = "destroyFailed"
 	destroyed       = "destroyed"
 )
 
 // The clusterController manages the lifecycle of a single cluster.
 type clusterController struct {
+	clusterName    string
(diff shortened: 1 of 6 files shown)


emmetthitz:
LGTM

pkg/controller/cluster.go
```diff
@@ -101,23 +118,31 @@ func (c *clusterController) transition(cluster store.Cluster) store.Cluster {
 		}
 		cluster.CurrentState = installing
 		return cluster
+	case installed:
+		if cluster.DesiredState == destroyed {
```
dkoshkin: Does this not yet support, mutating clusters?

emmetthitz: I think that's out of scope for this PR for now.

dkoshkin: Gotcha wanted to confirm that.

dkoshkin:
LGTM

pkg/controller/cluster.go
alexbrand: We'll have to figure out the mechanics for _modify_. I think the server might have to set current state to `planned`.

PR #936, opened 2017-11-27
Provisioner tagging improvements

### LINKED ISSUE #935: Provisioner: Improve tagging of the resources provisioned by KET
In order to quickly and easily find infrastructure that was created by KET, all resources should be tagged with a consistent scheme.

We must devise our own tagging style and be consistent through out. I have seen different styles in AWS for the tag keys: `aws:cloudformation:logical-id`, `kubernetes.io/cluster/testcluster7405`, `OpsFunction`, etc.

At a minimum, all resources created for a specific cluster should be tagged with `kismatic/clusterName` = `${clusterName}` (or something of the sort).

Nodes should be tagged with their roles: `kismatic/nodeRole` = `${join(nodeRoles, ",")}`.

emmetthitz:
Fixes #935


terraform/providers/aws/variables.tf
```diff
@@ -19,12 +19,22 @@ variable "public_ssh_key_path" {
   default = ""
 }
 
+variable "ssh_user" {
+  default = "roo1"
```
emmetthitz: oops

terraform/providers/aws/variables.tf
```diff
@@ -19,12 +19,22 @@ variable "public_ssh_key_path" {
   default = ""
 }
 
+variable "ssh_user" {
```
alexbrand: The expectations is for this variable to always be provided by KET to terraform. I don't think there's a case where having a default here is useful.

terraform/providers/aws/variables.tf
```diff
@@ -19,12 +19,22 @@ variable "public_ssh_key_path" {
...
+  default = "root"
+  //These will have to change when we want to also support RHEL/CentOS
+}
+
 variable "cluster_name" {
   default = "kismatic-cluster"
```
alexbrand: The expectations is for this variable to always be provided by KET to terraform. I don't think there's a case where having a default here is useful.

alexbrand:
It also looks like the tags required by #767 are missing.

emmetthitz:
Is that not the ClusterID?

terraform/providers/aws/variables.tf
emmetthitz: 👍

terraform/providers/aws/variables.tf
emmetthitz: Makes sense. Easy fix.

pkg/provision/aws.go
```diff
@@ -35,6 +36,12 @@ func (aws AWS) Provision(plan install.Plan) (*install.Plan, error) {
 	cmdDir := clusterStateDir
 	providerDir := fmt.Sprintf("../../providers/%s", plan.Provisioner.Provider)
 
+	//Get the user's name for cluster tagging
+	user, err := user.Current()
```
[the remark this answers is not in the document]
emmetthitz: Yeah. You're right. My line of thought here was that it should be reasonably easy to track who was generating clusters, but this definitely only makes sense in the CLI mode. Would it make more sense to add this information to the struct and have the CLI logic set it? Similar to how we already handle the binary path for the terraform executable?

emmetthitz:
Should all be good now. Let me know if there's anything else that needs to be done.

terraform/providers/aws/main.tf
```diff
@@ -35,14 +35,18 @@ resource "aws_vpc" "kismatic" {
   enable_dns_support    = true
   enable_dns_hostnames  = true
   tags {
-    Name = "kismatic - cluster"
+    kismatic/ClusterName = "${var.cluster_name}"
```
dkoshkin: `kismatic/ClusterName` looks strange to me, is there a standard that everyone else follows. Maybe `kismatic/clusterName`?

terraform/providers/aws/variables.tf
dkoshkin: Not sure if the commend was addressed.

emmetthitz: Yeah, it was. The comment just sat on the line above the change, so when I fixed it, the comment wasn't marked as old.

terraform/providers/aws/main.tf
emmetthitz: Sure. I thought the latter looked a little strange, since kismatic is already lower case. If you have any strong preference one way or the other, I'd be cool with doing either.

terraform/providers/aws/variables.tf
alexbrand: Seems like the default is still there.

terraform/providers/aws/variables.tf
```diff
@@ -19,12 +23,21 @@ variable "public_ssh_key_path" {
   default = ""
 }
 
+variable "ssh_user" {
+  default = "ubuntu"
+}
+
 variable "cluster_name" {
   default = "kismatic-cluster"
```
alexbrand: This probably shouldn't have a default either.

pkg/provision/provision.go
```diff
@@ -18,6 +20,8 @@ const terraformBinaryPath = "../../bin/terraform"
 type Terraform struct {
 	Output     io.Writer
 	BinaryPath string
+	Owner      string
```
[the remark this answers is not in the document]
emmetthitz: 👍

terraform/providers/aws/variables.tf
emmetthitz: Yeah. This is really the result of a to-do. It was either make the default here Ubuntu, or do that during planning. I think it's easier to change here, but you're right.

pkg/provision/provision.go
```diff
@@ -18,6 +20,8 @@ const terraformBinaryPath = "../../bin/terraform"
...
+	Version    semver.Version
```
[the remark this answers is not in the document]
emmetthitz: 👍

terraform/providers/aws/variables.tf
emmetthitz: I ended up having to add it back 🙃

emmetthitz: The issue here really stems from only supporting ubuntu for now. When we decide what to do with how to gather the user's desired OS, this will obviously have to change. I figured that was probably out of scope for now, so I think setting the default here is fine for now.

alexbrand:
LGTM. Needs squash.

PR #937, opened 2017-11-27
Make port preflight checks idempotent

### LINKED ISSUE #928: Improve preflight checks so that they can be run after the cluster is up
Currently, if an installation fails midway, the user is forced to skip preflight checks because some of the checks fail after specific installation steps have completed. For example, the API server port checks fail once the API server is running.

Ideally, the inspector should be able to run against a running cluster so that a) the user doesn't have to add an extra flag when retrying a failed installation, b) we get closer to an idempotent `apply` command.

This will also play an important role in our cluster controller.

alexbrand:
Fixes #928

### DIFF
--- a/integration/install_test.go
+++ b/integration/install_test.go
@@ -270,6 +270,10 @@ var _ = Describe("kismatic", func() {
 						sub.It("nodes should contain expected component overrides", func() error {
 							return ContainsOverrides(nodes, sshKey)
 						})
+
+						sub.It("should allow for running preflight checks idempotently", func() error {
+							return runValidate("kismatic-testing.yaml")
+						})
 					})
 				})
 			})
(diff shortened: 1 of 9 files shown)


dkoshkin:
Awesome LGTM!

alexbrand:
When working on this I was using Ubuntu. Before merging, I wanted to make sure that `lsof` is installed by default in RHEL / CentOS. Unfortunately, it isn't. So I'll have to rework this.

dkoshkin:
Even more LGTM! This is some cool stuff.

emmetthitz:
Right? No more need to --skip-preflight hype

dkoshkin:
We are doing an upgrade and it made me think that we should add port preflights to it. What do you guys think?

emmetthitz:
@dkoshkin elaborate?

swade1987:
@alexbrand @dkoshkin @emmetthitz is this going to be available in the 1.7.0 release?

Issue #938, opened 2017-11-27
Secure Helm and Tiller


jaycoon:
**Is this a BUG REPORT or FEATURE REQUEST?**
Feature Request

Kismatic installs Helm and Tiller, and uses RBAC for security, though there are 3 more security recommendations (summarized in a recent [blog post](http://technosophos.com/2017/11/20/securing-helm.html) from @technosophos) that KET can easily add during setup:

**Enable TLS between Helm and Tiller**
KET will need to generate 2 certs (one for Tiller and one for Helm), and add relevant flags to `helm init`. 

**Utilize Secrets for Tiller release information**
KET will change the default storage setting in the `tiller-deploy` deployment from ConfigMaps to Secrets.

**Run Tiller in a separate namespace**
KET will install Tiller in to an isolated namespace (created by KET), and grant the serviceaccount permissions needed to deploy to other namespaces. This ensures that Tiller is not running in the same namespace it is deploying to, and also sets up the framework to add additional Tiller instances (e.g. other groups/teams) segregated by namespace and RBAC, if the user chooses to.

technosophos:
This would be really cool!

magowant:
We are particularly interested in having tiller run in it's own namespace as this would allow us to properly secure user access to helm. With tiller running in kube-system we have to give users permissions to port-forward to that namespace, which is obviously not something we desire.

Would you reconsider adding this feature back into scope for KET 1.8.0? Thanks

PR #939, opened 2017-11-28
Secure Helm: separate ansible tasks

jaycoon:
First step to #938 is creating ansible tasks and templates for Helm


emmetthitz:
Wish we could get CI to trigger on forked merges. 
Can you potentially change the base branch to master-merge-from-forks, and then we can merge it in from there?

PR #940, opened 2017-11-28
Separating Helm Ansible tasks and adding Helm RBAC resources

emmetthitz:
@jaycoon began work on separating ansible tasks and adding Helm RBAC resources, but cannot trigger CI since he isn't one of the devs. This is a workaround for that.


alexbrand:
Closing as we won't pursue this through this route.

PR #942, opened 2017-11-29
Add first integration test for daemon mode

alexbrand:
Initial integration test for testing the creation of a cluster using the REST API exposed by the KET daemon.


integration/daemon_test.go
```diff
@@ -0,0 +1,267 @@
...
+var secretAccessKey = os.Getenv("AWS_SECRET_ACCESS_KEY")
+var tlsIgnoringClient = &http.Client{Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}}
+
+var daemonProcess *os.Process
+var daemonPort int
+
+// Had to redefine these structs instead of reusing the ones in the handler pkg
+// due to some weird import issue with cfssl.
+// See https://circleci.com/gh/apprenda/kismatic/1470
+type clusterRequest struct {
```
emmetthitz: There has to be some better workaround for this. I don't like the added maintainability this imposes simply because two packages decided to use flags with the same name.

emmetthitz: Did you try flattening dependencies when you ran into this problem, @alexbrand? 
`glide install --strip-vendor` should resolve the error you were seeing in https://circleci.com/gh/apprenda/kismatic/1470 without having to redefine everything. Hopefully everything depends on compatible versions...

alexbrand: Yeah. I don't love this either. I spent some time looking into but sinked enough that I decided to go down this path. I'll take another look and see where I end up, but we might have to live with it for now.

alexbrand: With that said, it could also be seen as a good thing, as if we ever change or break the API we would realize it immediately.

emmetthitz: Non-breaking changes or upgrades could also just completely fly under the radar as passing tests.

alexbrand: I think I understand this problem now. The problem is that the integration dir has it's own vendoring, so we would have to vendor kismatic in there, which basically results in this recursive vendoring issue.

emmetthitz: oooooh

integration/daemon_test.go
```diff
@@ -0,0 +1,231 @@
...
+	"strconv"
+	"time"
+
+	"github.com/apprenda/kismatic/pkg/server/http/handler"
+	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/config"
+	. "github.com/onsi/gomega"
+)
+
+var accessKeyID = os.Getenv("AWS_ACCESS_KEY_ID")
```
emmetthitz: How are we going to want to do this in the not-AWS cases?

emmetthitz:
I think this otherwise looks good. It's still a bit up in the air while we figure out how we want to handle creds across providers, though.

integration/daemon_test.go
alexbrand: Excellent question. We are currently only setup to run integration tests on AWS, so it's hard to tell at this point.

dkoshkin:
LGTM!

PR #943, opened 2017-11-29
Bump calico version to v2.6.3

### LINKED ISSUE #941: Update Calico to 2.6.3
There is a new point release of Calico, [see here](https://github.com/projectcalico/calico/releases/tag/v2.6.3).

It appears there are a number of security and bug fixes, see below for the changeset:

> - Addresses several outstanding common vulnerabilities and exposures (CVE).
> - A new node controller for Kubernetes deployments clears data associated with deleted nodes from the Calico datastore, preventing conflicts that can lead to crash loops.
> - Reduces default confd log level from debug to info.
> - #418: If the CNI-plugin fails to re-network an existing endpoint, it no longer releases the IP allocations to the pool that are still attached to the endpoint. (@fasaxc)
> - #408: The CNI-plugin no longer throws file exists while programming route (@gunjan5)

alexbrand:
Fixes #941

### DIFF
--- a/ansible/roles/calico/templates/calico.yaml
+++ b/ansible/roles/calico/templates/calico.yaml
@@ -1,9 +1,9 @@
-# Calico Version v2.6.0
-# https://docs.projectcalico.org/v2.6/releases#v2.6.0
+# Calico Version v2.6.3
+# https://docs.projectcalico.org/v2.6/releases#v2.6.3
 # This manifest includes the following component versions:
-#   calico/node:v2.6.0
-#   calico/cni:v1.11.0
-#   calico/kube-controllers:v1.0.0
+#   calico/node:v2.6.3
+#   calico/cni:v1.11.1
+#   calico/kube-controllers:v1.0.1
 
 # This ConfigMap is used to configure a self-hosted Calico installation.
 kind: ConfigMap
@@ -77,6 +77,9 @@ spec:
     spec:
       hostNetwork: true
       serviceAccountName: calico-node
+      # Minimize downtime during a rolling upgrade or deletion; tell Kubernetes to do a "force
+      # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
+      terminationGracePeriodSeconds: 0
       tolerations:
         - effect: NoSchedule
           operator: Exists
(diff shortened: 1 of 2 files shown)


alexbrand:
Pending review of calico deployment spec files.

swade1987:
Based on the official Calico configuration.

We are missing the following from our configuration template under the `calico-node` container.

```
 # Cluster type to identify the deployment type
            - name: CLUSTER_TYPE
              value: "k8s"
```

dkoshkin:
@swade I believe that option sets the backend to k8s instead of etcd, which is not what we want.

swade1987:
@dkoshkin it identifies the deployment type options are either "k8s" or "bgp"

swade1987:
I am basing this off https://docs.projectcalico.org/v2.6/getting-started/kubernetes/installation/hosted/calico.yaml

dkoshkin:
You are correct @swade1987 looks like we need to update those fields.

swade1987:
@dkoshkin also if we are moving to RBAC might be worth leveraging https://docs.projectcalico.org/v2.6/getting-started/kubernetes/installation/rbac.yaml

dkoshkin:
Might be good to take a look at https://github.com/apprenda/kismatic/issues/968, especially if we are modifying the `CLUSTER_TYPE` in this PR.

alexbrand:
To me it seems like `CLUSTER_TYPE` should be `"k8s,bgp"`, as set in the standard manifest: https://docs.projectcalico.org/v2.6/getting-started/kubernetes/installation/hosted/calico.yaml

ansible/roles/calico/templates/calico.yaml
```diff
@@ -77,6 +77,9 @@ spec:
     spec:
       hostNetwork: true
       serviceAccountName: calico-node
+      # Minimize downtime during a rolling upgrade or deletion; tell Kubernetes to do a "force
+      # deletion": https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods.
+      terminationGracePeriodSeconds: 0
```
dkoshkin: Nice! This should also fix https://github.com/apprenda/kismatic/issues/964.

dkoshkin:
LGTM

PR #960, opened 2017-12-01
Added ELBs on AWS for ingress and master nodes.

emmetthitz:
Only minor thing I'm going to leave out of this is the fact that they don't log (yet).

### DIFF
--- a/pkg/provision/aws.go
+++ b/pkg/provision/aws.go
@@ -49,9 +49,10 @@ func (aws AWS) Provision(plan install.Plan) (*install.Plan, error) {
 
 	// Write out the terraform variables
 	data := AWSTerraformData{
+		Version:           aws.Terraform.KismaticVersion.String(),
 		Region:            plan.Provisioner.AWSOptions.Region,
 		ClusterName:       plan.Cluster.Name,
-		ClusterOwner:      aws.Terraform.Owner,
+		ClusterOwner:      aws.Terraform.ClusterOwner,
 		MasterCount:       plan.Master.ExpectedCount,
 		EtcdCount:         plan.Etcd.ExpectedCount,
 		WorkerCount:       plan.Worker.ExpectedCount,
--- a/pkg/provision/aws.go
+++ b/pkg/provision/aws.go
@@ -120,8 +120,17 @@ func (aws *AWS) buildPopulatedPlan(plan install.Plan) (*install.Plan, error) {
 		ExpectedCount: masterNodes.ExpectedCount,
 		Nodes:         masterNodes.Nodes,
 	}
-	mng.LoadBalancedFQDN = tfNodes.IPs[0]
-	mng.LoadBalancedShortName = tfNodes.IPs[0]
+	if masterNodes.ExpectedCount > 1 {
+		mlb, err := aws.getLoadBalancer(plan.Cluster.Name, "master")
+		if err != nil {
+			return nil, err
+		}
+		mng.LoadBalancedFQDN = mlb
+		mng.LoadBalancedShortName = mlb
+	} else {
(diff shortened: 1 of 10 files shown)


pkg/provision/aws.go
```diff
@@ -120,8 +120,17 @@ func (aws *AWS) buildPopulatedPlan(plan install.Plan) (*install.Plan, error) {
...
-	mng.LoadBalancedShortName = tfNodes.IPs[0]
+	if masterNodes.ExpectedCount > 1 {
+		mlb, err := aws.getLoadBalancer(plan.Cluster.Name, "master")
+		if err != nil {
+			return nil, err
+		}
+		mng.LoadBalancedFQDN = mlb
+		mng.LoadBalancedShortName = mlb
+	} else {
+		mng.LoadBalancedFQDN = mng.Nodes[0].InternalIP
```
emmetthitz: oh oops.

terraform/providers/aws/main.tf
```diff
@@ -4,9 +4,9 @@ provider "aws" {
   $ export AWS_SECRET_ACCESS_KEY=YOUR_AWS_SECRET_ACCESS_KEY
   $ export AWS_DEFAULT_REGION=us-east-1
   */
-  region = "${var.region}"
-  access_key = "${var.access_key}"
-  secret_key = "${var.secret_key}"
+  region      = "${var.region}"
```
emmetthitz: It's probably worth breaking this whole thing up into its respective components so we don't have 500 lines of a monolith terraform script. TODO.

terraform/providers/aws/main.tf
```diff
@@ -121,72 +120,313 @@ resource "aws_subnet" "kismatic_private" {
...
+    "kismatic/version"      = "${var.version}"
+    "kismatic/subnet"       = "ingress"
+    "kubernetes.io/cluster" = "${var.cluster_name}"
+  }
+  lifecycle {
+    ignore_changes = ["tags.kismatic/dateCreated", "tags.Owner", "tags.PrincipalID"]
+  }
+}
+
+resource "aws_security_group" "kismatic_public" {
```
alexbrand: Can we call this `kismatic_ssh`, or something? I think it's going to make it more clear when referenced in other resources.

pkg/provision/provision.go
```diff
@@ -49,6 +49,29 @@ type Provisioner interface {
...
+
+	//load balancer
+	tfCmdOutputLB := exec.Command(tf.BinaryPath, "output", "-json", tfOutLB)
+	tfCmdOutputLB.Dir = cmdDir
+	stdoutStderrLB, err := tfCmdOutputLB.CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("Error collecting terraform output: %s", stdoutStderrLB)
+	}
+	lbData := tfOutputVar{}
+	json.Unmarshal(stdoutStderrLB, &lbData)
```
alexbrand: We need to handle the error here, as this can fail.

terraform/providers/aws/main.tf
```diff
@@ -121,72 +120,313 @@ resource "aws_subnet" "kismatic_private" {
...
+
+  //access_logs {
+  //  bucket = "${aws_s3_bucket.lb_logs.bucket}"
+  //  bucket_prefix = "${var.cluster_name}/ingress"
+  //}
+
+  listener {
+    instance_port     = 6443
+    instance_protocol = "tcp"
+    lb_port           = 4443
```
alexbrand: Typo

terraform/providers/aws/main.tf
emmetthitz: Yeah, sure.

dkoshkin:
LGTM

PR #966, opened 2017-12-04
Upgrade NGINX Ingress Controller to version 0.9.0 #854

### LINKED ISSUE #854: Upgrade Nginx Ingress Controller

**FEATURE REQUEST**

**What happened**:

Current version of Kismatic (1.6.0) still uses an oldder version Nginx Ingress controller: 0.8.3 

From group_vars/container_images.yaml:

```yaml
  nginx_ingress_controller:
    name: gcr.io/google_containers/nginx-ingress-controller
    version: 0.8.3
```

Can this version be updated to current or at least in the 0.9.x to take advantage of all the new features?

**Environment**:
- Kismatic version (use `kismatic version`): 1.6.0

swade1987:
Fixes https://github.com/apprenda/kismatic/issues/854

Have made the following changes:

- Upgraded NGINX Ingress Controller to version 0.9.0
- Upgraded the defaultbackend image to version 1.4
- Added the Prometheus scrape annotation to the NGINX Ingress controller
- Added a config map to the NGINX Ingress Controller to enable VTS metrics
- Turned off profiling on the NGINX Ingress Controller

Fixes #854 

cc @aledbf @jonmosco @jarkomes @mrgleeco @willvrny

### DIFF
--- a/ansible/roles/kube-ingress/templates/nginx-ingress-controller.yaml
+++ b/ansible/roles/kube-ingress/templates/nginx-ingress-controller.yaml
@@ -10,6 +10,8 @@ spec:
         name: ingress
       annotations:
         kismatic/version: "{{ kismatic_short_version }}"
+        prometheus.io/port: "10254"
+        prometheus.io/scrape: "true"
     spec:
       terminationGracePeriodSeconds: 60
       hostNetwork: true # required in a CNI networkd
@@ -49,3 +51,21 @@ spec:
         args:
         - /nginx-ingress-controller
         - --default-backend-service=kube-system/default-http-backend
+        - --configmap=$(POD_NAMESPACE)/nginx-conf
+        - --profiling=false
+        - --annotations-prefix=ingress.kubernetes.io
+---
+apiVersion: v1
+data:
+  enable-vts-status: "true"
+  access-log-path: "/dev/stdout"
+  error-log-path: "/dev/stdout"
+  error-log-level: "warn"
+  proxy-connect-timeout: "60"
+  proxy-read-timeout: "60"
+  proxy-send-timeout: "60"
+kind: ConfigMap
+metadata:
+  name: nginx-conf
+  namespace: kube-system
+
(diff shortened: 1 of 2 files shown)


aledbf:
@swade1987  this LGTM

ansible/roles/kube-ingress/templates/nginx-ingress-controller.yaml
```diff
@@ -49,3 +51,21 @@ spec:
         args:
         - /nginx-ingress-controller
         - --default-backend-service=kube-system/default-http-backend
+        - --configmap=$(POD_NAMESPACE)/nginx-conf
+        - --profiling=false
+        - --annotations-prefix=nginx.ingress.kubernetes.io
```
alexbrand: Is this different than the annotation prefix that was used in previous versions? If so, this will have upgrade implications right?

swade1987: The configmap was never there in previous releases.

Also the NGINX ingress annotations contains a new prefix nginx.ingress.kubernetes.io. This change is behind a flag to avoid breaking running deployments. This is why I added the `--annotations-prefix=ingress.kubernetes.io`

alexbrand: Is this a typo then? The prefix being set is `nginx.ingress.kubernetes.io`

swade1987:
@alexbrand @aledbf see the change to `annotation-prefix` now.

emmetthitz:
I think this LGTM as well, just remember to squash when you merge.

swade1987:
@emmetthitz I will let one of you guys squash and merge.

ansible/roles/kube-ingress/templates/nginx-ingress-controller.yaml
magowant: @swade1987 this is the wrong annotation please use --annotations-prefix=ingress.kubernetes.io otherwise current deployements will break. Please see https://github.com/kubernetes/ingress-nginx/blob/master/Changelog.md

dkoshkin:
LGTM! Will merge after the tests pass again.

PR #967, opened 2017-12-04
Fixes non-HA provisioning cases

emmetthitz:
image
Bugs vanish


emmetthitz:
Not sure why this is diffing weirdly. Can redo the PR if we care that much.

Issue #934, opened 2017-11-27
Provisioner: AWS provider should create a load balancer when standing up multiple API servers


alexbrand:
Currently, the AWS provisioner is able to create multiple master nodes. However, this is still not an HA setup given that all cluster components are being setup to talk to the first master node. 

In order to achieve a truly HA setup, we must create a load balancer when multiple master nodes are in play to load balance access to the API server endpoint. The load balancer configuration should include health checking to detect when a downstream instance is down.

emmetthitz:
My only concern here is extending this then to the provider-agnostic cases. Is there no way to do this with kube-proxy?

emmetthitz:
Should there be separate buckets for load balancer logs for each cluster?

PR #969, opened 2017-12-09
Load balancer fixes, Terraform readability changes.

### LINKED ISSUE #961: Provisioner: Break up the monolith aws/main.tf into more easily ingestible bites

emmetthitz:
Fixes #961, #934 
- Breaks up the related terraform config components into their related categories
- Fixes load balancer health checks to bring them online more quickly


pkg/provision/aws_types.go
```diff
@@ -11,9 +11,10 @@ type AWS struct {
 type AWSTerraformData struct {
 	Version           string `json:"version"`
 	Region            string `json:"region,omitempty"`
+	AvailabilityZone  string `json:"AZ,omitempty"`
 	PrivateSSHKeyPath string `json:"private_ssh_key_path"`
 	PublicSSHKeyPath  string `json:"public_ssh_key_path"`
-	SSHUser           string `json:"ssh_user,omitempty"`
+	SSHUser           string `json:"ssh_user"`
```
dkoshkin: Can this ever be empty?

terraform/providers/aws/elb.tf
```diff
@@ -0,0 +1,95 @@
+resource "aws_s3_bucket" "lb_logs" {
+  count  = 0
+  //"${var.master_count > 1 || var.ingress_count > 1 ? 1 : 0}"
```
dkoshkin: This conditional can be removed now.

terraform/providers/aws/instances.tf
```diff
@@ -0,0 +1,203 @@
+resource "aws_instance" "bastion" {
```
dkoshkin: Is the bastion being used? I think we wanted to hold off on this until we figure some more things out?

emmetthitz: Yeah. The count is set to 0. I left todos where stuff will need to be changed, otherwise this isn't actually doing anything.

terraform/providers/aws/elb.tf
emmetthitz: 👍

alexbrand:
The tests are failing because the load balancers take a while to come up (I have seen them take ~3 minutes). We will have to bump up the daemon integration test timeout... probably something like 5 minutes.

emmetthitz:
Oh bless. I was slowly going crazy because I couldn't reproduce what was happening with the tests manually.

pkg/provision/aws_types.go
emmetthitz: I.e. no user is specified in the plan, so the tfvar doesn't get marshalled. No tfvar for that field defaults to root. 

Instead, always marshalling the tfvar causes it to be empty string in the case where it's not set in the plan, which causes failure down the pipe (here).

pkg/provision/aws.go
```diff
@@ -165,8 +161,6 @@ func (aws *AWS) buildPopulatedPlan(plan install.Plan) (*install.Plan, error) {
 		plan.Storage = install.OptionalNodeGroup(nodeGroupFromSlices(tfNodes.IPs, tfNodes.InternalIPs, tfNodes.Hosts))
 	}
 
-	// SSH
-	plan.Cluster.SSH.User = "ubuntu"
```
emmetthitz: This needs to be set before the TFVars get marshalled. This line of code wasn't actually doing anything useful.

terraform/providers/aws/variables.tf
```diff
@@ -20,7 +24,7 @@ variable "public_ssh_key_path" {
 }
 
 variable "ssh_user" {
-  default = "ubuntu"
+  default = ""
```
emmetthitz: pkg/provision/aws.go:169 was being caught by this.

alexbrand: This is exactly why I wanted us to have no defaults in this file.

emmetthitz: Alex "Confucius" Brand at it again.

PR #971, opened 2017-12-11
Enable kube dns metrics

### LINKED ISSUE #945: Add labels to kube-dns to enable prometheus scraping
The following labels have to be added to the kubedns service so that prometheus can get metrics:
```
prometheus.io/port: “10054”
prometheus.io/scrape: “true”
```

cc: @swade1987

swade1987:
Closes #945


emmetthitz:
LGTM, though I'm not sure how valuable this will end up being, since I think the plan is to eventually switch to CoreDNS. That's what the community seems to be leaning towards, at least.

swade1987:
Agreed but for our current implementation, I feel it's useful to have metrics on the internal cluster DNS.

dkoshkin:
LGTM.

PR #972, opened 2017-12-11
Make ssh key generation idempotent

emmetthitz:


pkg/provision/aws.go
```diff
@@ -42,11 +42,22 @@ func (aws AWS) Provision(plan install.Plan) (*install.Plan, error) {
 	}
 	pubKeyPath := filepath.Join(absPath, fmt.Sprintf("/terraform/clusters/%s/%s-ssh.pub", plan.Cluster.Name, plan.Cluster.Name))
 	privKeyPath := filepath.Join(absPath, fmt.Sprintf("/terraform/clusters/%s/%s-ssh.pem", plan.Cluster.Name, plan.Cluster.Name))
-	if err := ssh.NewKeyPair(pubKeyPath, privKeyPath); err != nil {
-		return nil, fmt.Errorf("error generating SSH key pair: %v", err)
+
+	privFlag := true
```
alexbrand: What does `privFlag` mean? Maybe `privExists` and `pubExists` is more meaningful? or `privMissing` and `pubMissing`?

pkg/provision/aws.go
```diff
@@ -42,11 +42,22 @@ func (aws AWS) Provision(plan install.Plan) (*install.Plan, error) {
...
+		pubFlag = false
+	}
+	if _, err := os.Stat(privKeyPath); err == nil {
+		privFlag = false
+	}
+	if privFlag && pubFlag {
+		if err := ssh.NewKeyPair(pubKeyPath, privKeyPath); err != nil {
+			return nil, fmt.Errorf("error generating SSH key pair: %v", err)
+		}
+		plan.Cluster.SSH.Key = privKeyPath
```
alexbrand: We should always be setting these two, right? regardless of the existence of SSH keys?

pkg/provision/aws.go
emmetthitz: Sure thing.
>Coding is really simple, but we insist on making it complicated.

pkg/provision/aws.go
```diff
@@ -42,8 +42,19 @@ func (aws AWS) Provision(plan install.Plan) (*install.Plan, error) {
...
+
+	privKeyMissing := true
+	pubKeyMissing := true
+	if _, err := os.Stat(pubKeyPath); err == nil {
+		pubKeyMissing = false
+	}
+	if _, err := os.Stat(privKeyPath); err == nil {
+		privKeyMissing = false
+	}
+	if privKeyMissing && pubKeyMissing {
```
dkoshkin: Wouldn't we want to regenerate if either one is missing? So `or` not `and`

pkg/provision/aws.go
```diff
@@ -42,8 +42,19 @@ func (aws AWS) Provision(plan install.Plan) (*install.Plan, error) {
 	}
 	pubKeyPath := filepath.Join(absPath, fmt.Sprintf("/terraform/clusters/%s/%s-ssh.pub", plan.Cluster.Name, plan.Cluster.Name))
 	privKeyPath := filepath.Join(absPath, fmt.Sprintf("/terraform/clusters/%s/%s-ssh.pem", plan.Cluster.Name, plan.Cluster.Name))
-	if err := ssh.NewKeyPair(pubKeyPath, privKeyPath); err != nil {
-		return nil, fmt.Errorf("error generating SSH key pair: %v", err)
+
+	privKeyMissing := true
```
dkoshkin: Key we reverse these and use default values of false, `privKeyExists`, `pubKeyExists` and then set to false in the if.

emmetthitz: Sure. Is there a reason why?

pkg/provision/aws.go
emmetthitz: My original thought here was that if only one is missing, the user could either delete the one they have, or try to recover the one that was lost. Regenerating both keys does *not* populate the machines that have already been created with the new key, so you effectively just lose access to any machines when that happens.

dkoshkin: But dont we need both the keys? The public when doing TF stuff and then the private when running Ansible?

emmetthitz: Yes, but AWS registers a key pair with a particular instance for SSH access. If we're forced to regenerate a key pair, the key pair that was registered with all the machines during creation is still the one being used by the machines (even though the key pair the user now has is the one that's registered). Regenerating keys potentially forces the user to destroy any instances using the prior key pair, unless they want to update it manually.

emmetthitz: And even then, that's not possible in the case where the private key goes missing.

pkg/provision/aws.go
```diff
@@ -42,8 +42,26 @@ func (aws AWS) Provision(plan install.Plan) (*install.Plan, error) {
...
+	var privKeyExists, pubKeyExists bool
+	if _, err := os.Stat(pubKeyPath); err == nil {
+		pubKeyExists = true
+	}
+	if _, err := os.Stat(privKeyPath); err == nil {
+		privKeyExists = true
+	}
+
+	if pubKeyExists != privKeyExists {
+		if !pubKeyExists {
```
dkoshkin: This should be `!privKeyExists `

dkoshkin:
LGTM

PR #973, opened 2017-12-12
Adding specific OS versions supported

swade1987:


README.md
```diff
@@ -38,9 +38,9 @@ KET operational tools include:
 [Download KET here](https://github.com/apprenda/kismatic/releases)
 
 ## Supported Operating Systems
-- RHEL 7
-- CentOS 7
-- Ubuntu 16.04
+- RHEL 7.4
+- CentOS 7.3
+- Ubuntu 16.04.3
```
emmetthitz: I'm not sure it's worth adding the patch release version here. Patch releases generally don't include breaking, so 16.04 should be enough.

swade1987: Agreed but I would like to be specific about what we test against.

emmetthitz: Then I'm not sure this is the place to do it. I even think that the patch release versions get upgraded automatically when new releases hit the AMI marketplace, so this information won't be accurate for long.

swade1987: But our tests are pinned to a specific AMI, right?

If so, they will have a specific patch version unless we do an "apt dist-upgrade" or equivalent.

emmetthitz: Some of them are. I know for certain the daemon tests are pinned to the most recent patch release. I _think_ the others use a specific version.

swade1987: My reasoning behind this is because if we are specific for Ubuntu we should be specific for RHEL and CentOS as well to be consistent.

emmetthitz: Are RHEL releases not semver? IE not major.minor.patch?

swade1987: Based on https://access.redhat.com/articles/3078 it doesn't look like the case.

dkoshkin:
LGTM!

Issue #962, opened 2017-12-04
Server mode: Handlers and controllers can clobber each other during operation


alexbrand:
After posting an initial cluster definition, the controller goes to work and starts bringing the cluster up. While transitioning the cluster through the states, the controller updates the cluster spec in the store so that the user can tell the cluster's current state. This is working as designed, but the current implementation has a couple of flaws that become apparent when wanting to modify or destroy a cluster.

Currently, any changes that are made through the API to a cluster are wiped if the change is performed while the controller is transitioning a cluster. This means, for example, that the destruction of a cluster cannot be requested while the cluster is being provisioned or installed. 

The root of the problem is that the controller reads the original cluster spec before performing the transition, and will write it back (with updates to a subset of fields) after the transition is complete. It seems like we might have to arrive at stronger read/write contracts between the handlers and the controllers.

This is an example of a scenario that I would expect us to handle properly:
1. The user posts a cluster specification
2. The controller starts provisioning the cluster
3. The user issues a DELETE on the cluster resource while the controller is still provisioning the infrastructure
4. The controller proceeds to destroy the cluster after the provisioning phase is done, instead of installing it.

PR #975, opened 2017-12-12
Fixes issues where controller and server can clobber each other

alexbrand:
As described in #962, the controller and server overwrite their changes during operation. This is a problem because any change requested by the user can be wiped by the controller when it finishes an in-flight operation.

To avoid this issue, the controller should only be able to touch the current state field, while the server should only touch the desired state field. Given that these fields live in the same type and that controller operations are long-running, the controller needs to re-read the cluster object before updating it to pick up any changes that might have occurred while performing the operation. Ideally, this would be done in a transaction, but the current implementation of the store does not expose them. We will have to address this in a follow up issue.

Another outstanding issue is that of changing the plan file, as both the controller and server are currently manipulating the plan. We can get away with it for now because they touch different fields, but it might become a problem in the future. I am starting to think that the plan file should be considered an implementation detail that is handled by the controller, but this needs more thought.

### DIFF
--- a/pkg/controller/cluster.go
+++ b/pkg/controller/cluster.go
@@ -6,6 +6,7 @@ import (
 	"github.com/apprenda/kismatic/pkg/install"
 	"github.com/apprenda/kismatic/pkg/provision"
 	"github.com/apprenda/kismatic/pkg/store"
+	"github.com/google/go-cmp/cmp"
 )
 
 const (
@@ -26,6 +27,7 @@ const (
 // The clusterController manages the lifecycle of a single cluster.
 type clusterController struct {
 	clusterName    string
+	clusterSpec    store.Cluster
 	log            *log.Logger
 	executor       install.Executor
 	newProvisioner func(store.Cluster) provision.Provisioner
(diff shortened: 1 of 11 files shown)


alexbrand:
After an initial iteration on this, I realized that I could decouple server and cluster even more by actually having them update different structs. That is what I am proposing here.

The server will write/update the cluster specification, whereas the controller will write/update the status.

emmetthitz:
Makes sense to me. Is this still WIP then?

alexbrand:
Yes, I think this is ready. 

There is some outstanding work that I want to complete in follow up PRs to not make this one even larger.

pkg/controller/cluster.go
```diff
@@ -62,161 +108,248 @@ func (c *clusterController) run(watch <-chan struct{}) {
...
+		EtcdNodes:     clusterSpec.EtcdCount,
+		MasterNodes:   clusterSpec.MasterCount,
+		WorkerNodes:   clusterSpec.WorkerCount,
+		IngressNodes:  clusterSpec.IngressCount,
+	}
+	planner := &install.BytesPlanner{}
+	if err := install.WritePlanTemplate(planTemplate, planner); err != nil {
+		return fmt.Errorf("could not write plan template: %v", err)
+	}
+	var p *install.Plan
```
emmetthitz: I think the next line makes this unnecessary?

pkg/controller/cluster.go
```diff
@@ -62,161 +108,248 @@ func (c *clusterController) run(watch <-chan struct{}) {
...
+	// Set values in the plan
+	p.Cluster.Name = clusterName
+	p.Provisioner = install.Provisioner{Provider: clusterSpec.Provisioner.Provider}
+	// TODO: Handle provisioner specific options (e.g. AWS Region)
+	switch clusterSpec.Provisioner.Provider {
+	case "aws":
+		p.Provisioner.AWSOptions = &install.AWSProvisionerOptions{}
+		// nothing
+	case "azure":
+		p.AddOns.CNI.Provider = "weave"
```
emmetthitz: Do we have to use weave on azure?

alexbrand: We do, yes. Calico is not supported on Azure.

pkg/controller/cluster.go
alexbrand: Fixed. Thanks!

dkoshkin:
LGTM!

PR #977, opened 2017-12-13
Expose docker log driver options

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/685, fixes https://github.com/apprenda/kismatic/issues/674

Sample plan file:
```
docker:
  logs:
    driver: json-file
    opts:
      max-file: "1"
      max-size: 50m
```

The values above will be used in the docker config json unless the user overwrites these options.

The `opts` is a `map[string]string` so any option can be set.

### DIFF
--- a/pkg/install/plan.go
+++ b/pkg/install/plan.go
@@ -99,6 +99,14 @@ func readDeprecatedFields(p *Plan) {
 }
 
 func setDefaults(p *Plan) {
+	if p.Docker.Logs.Driver == "" {
+		p.Docker.Logs.Driver = "json-file"
+		p.Docker.Logs.Opts = map[string]string{
+			"max-size": "50m",
+			"max-file": "1",
+		}
+	}
+
 	if p.AddOns.CNI == nil {
 		p.AddOns.CNI = &CNI{}
 		p.AddOns.CNI.Provider = cniProviderCalico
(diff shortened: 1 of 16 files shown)


dkoshkin:
This PR is a simplified version of https://github.com/apprenda/kismatic/pulls

pkg/install/plan.go
```diff
@@ -99,6 +99,14 @@ func readDeprecatedFields(p *Plan) {
 }
 
 func setDefaults(p *Plan) {
+	if p.Docker.Logs.Driver == "" {
+		p.Docker.Logs.Driver = "json-file"
+		p.Docker.Logs.Opts = map[string]string{
+			"max-size": "50m",
```
emmetthitz: Can we change this 50MB or something a little less ambiguous?

dkoshkin: Thats a docker setting https://docs.docker.com/engine/admin/logging/json-file/

emmetthitz: That is astonishingly stupid. Alright, never mind then.

swade1987:
So what will be the defaults shipped in KET for `max-file` and `max-size` ?

dkoshkin:
@swade1987 the defaults will be:
```
      max-file: "1"
      max-size: 50m
```

pkg/install/plan.go
alexbrand: @emmetthitz Can you elaborate?

alexbrand:
I think this LGTM. I think the last piece is adding docs about this to our docker.md.

swade1987:
@dkoshkin just an FYI, we normally set `max-size: 3m`

dkoshkin:
@swade1987 lets discuss this today.

PR #982, opened 2017-12-14
Deploy kube-proxy as a daemon set

### LINKED ISSUE #806: kube-proxy not running as a DaemonSet
**BUG REPORT**:

**What happened**:

kube-proxy is deployed as a pod to each node. 

As kube-proxy needs to run on each node to manage the iptables/ipvs rules on it, it should be deployed as a DeamonSet because if the proxy dies on one of the nodes it'll never be rescheduled.

### LINKED ISSUE #951: kube-proxy not marked as a critical pod
We are seeing, what I think is, premature pod eviction where disk pressure is reported, however, there is ample space on each partition on the node reported.

I have attached a log from the journal for the kubelet process, where it has reported disk pressure and decided to evict the kube-proxy (see gist below)

https://gist.github.com/swade1987/13696860b151bdc2a1b9e579d60712f8

In the official Kubernetes repo, the kube-proxy manifest is a DaemonSet and marked as a critical pod.

https://github.com/kubernetes/kubernetes/blob/47d0db0e876a120dc1e04f8e7e812540916e9d5d/cluster/addons/kube-proxy/kube-proxy-ds.yaml#L24-L25

If the kube-proxy was a critical pod it would remain on the node.

### LINKED ISSUE #958: Storage and Ingress node labels aren't applied so services are unavailable
**BUG REPORT**:

**What happened**:
Applying clean plan with storage and ingress enabled, nodes aren't labeled with kismatic/storage=true and kismatic/ingress=true.  Without these labels ingress and storage daemonsets have no endpoints as no nodes match the selector.

When nodes are assigned to the plan document in storage and ingress sections... they should get the necessary labels so the specified services work.

Create a default plan and add nodes to ingress and/or storage.  Look at daemonsets and notice that there are no running pods (DESIRED, CURRENT, READY, UP-TO-DATE,AVAILABLE columns are all zero) for gluster-healthz, ingress, default-http-backend services.

If node labeling is meant to be manual, please udpdate documentation to reflect so.  Manually labeling the nodes immediately fixed the problem as pods were created and the services worked as intended.

**Environment**:
- Kismatic version (use `kismatic version`): 1.6.1 & 1.6.3
- Plan file Gist (remove sensitive information if any): n/a
- Cloud provider or hardware configuration: bare metal
- OS (e.g. from /etc/os-release): CentOS7
- Kernel (e.g. `uname -a`): 3.10

dkoshkin:
Fixes #806, fixes #951, fixes #958 

Deploy from [add-ons](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/kube-proxy)

The new deployment no longer relies on a `kubeconfig` and instead uses a new service-account. Because the proxy is started before `kube-dns`, we need to set a few `envs`.
```
        env:
        - name: KUBERNETES_SERVICE_HOST
          value: "{{ kubernetes_load_balanced_fqdn }}"
        - name: KUBERNETES_SERVICE_PORT
          value: "{{ kubernetes_master_secure_port }}"
```
More [context](https://github.com/kubernetes/kubernetes/pull/52183)

To handle upgrades:
1) remove the manifest and wait for it to shutdown
2) label the nodes with `kismatic/kube-proxy=true` 
3) deploy kube-proxy DS with a node selector
4) wait for `desiredNumberScheduled` == `numberReady`

This PR also modifies how nodes get labeled. Prior to this node labels were performed by setting the `node-labels` kubelet flag, however this flag only gets used on initial node registration. This is a problem for cluster upgrades.
A related change is that installer set `kismatic/host` label on the nodes and then was used for node selections, due to the behavior described above this did not work correctly on upgrades. Instead the installer will use the well defined label `kubernetes.io/hostname` that is always set.


swade1987:
Why the need for a node selector?

dkoshkin:
The selector is required to do a a rolling upgrade from the old method of deploying with manifests. We have a similar process with calico.

dkoshkin:
CircleCI is having a hard time running slow tests for some reason, the builds keep getting cancelled. The basic tests passed and I ran a few upgrade tests locally
```
• [SLOW TEST:1112.775 seconds]
Upgrade
/Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:12
  Upgrading a cluster using online mode
  /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:14
    From KET version v1.5.3
    /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:687
      Using a minikube layout
      /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:693
        Using RedHat 7
        /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:702
          when using AWS infrastructure [aws]
          /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/framework.go:13
            should be upgraded [slow] [upgrade]
            /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/framework.go:14
------------------------------
SSSSSSSSSSSSSSSSSSSSSSSSStar: Removing leading '/' from member names
Uploading logs to S3. Get them at https://console.aws.amazon.com/s3/buckets/kismatic-integration-tests/logs/kismatic-integration-2017-12-15T18:06:49+0000.tar.gz

Ran 1 of 54 Specs in 1113.291 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 53 Skipped PASS

• [SLOW TEST:1106.931 seconds]
Upgrade
/Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:12
  Upgrading a cluster using online mode
  /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:14
    From KET version v1.6.2
    /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:15
      Using a minikube layout
      /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:21
        Using RedHat 7
        /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:30
          when using AWS infrastructure [aws]
          /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/framework.go:13
            should be upgraded [slow] [upgrade]
            /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/framework.go:14
------------------------------
SSSSSSSSSSSSSSSSSSStar: Removing leading '/' from member names
Uploading logs to S3. Get them at https://console.aws.amazon.com/s3/buckets/kismatic-integration-tests/logs/kismatic-integration-2017-12-15T18:26:04+0000.tar.gz

Ran 1 of 54 Specs in 1107.954 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 53 Skipped PASS

• [SLOW TEST:7499.742 seconds]
Upgrade
/Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:12
  Upgrading a cluster using online mode
  /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:14
    From KET version v1.6.2
    /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:15
      Using a skunkworks cluster
      /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/upgrade_test.go:42
        when using AWS infrastructure [aws]
        /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/framework.go:13
          should result in an upgraded cluster [slow] [upgrade]
          /Users/dkoshkin/go/src/github.com/apprenda/kismatic/integration/framework.go:14
------------------------------
SSSSSSSSSSSSSSSSSStar: Removing leading '/' from member names
Uploading logs to S3. Get them at https://console.aws.amazon.com/s3/buckets/kismatic-integration-tests/logs/kismatic-integration-2017-12-15T23:35:16+0000.tar.gz

Ran 1 of 54 Specs in 7500.803 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 53 Skipped PASS
```

If the PR looks good, lets merge it and we can test all the changes together when releasing.

ansible/roles/kube-proxy/tasks/main.yaml
```diff
@@ -1,18 +1,59 @@
...
       src: kube-proxy.yaml
-      dest: "{{ kubelet_pod_manifests_dir }}/kube-proxy.yaml"
+      dest: "{{ kubernetes_spec_dir }}/kube-proxy.yaml"
       owner: "{{ kubernetes_owner }}"
       group: "{{ kubernetes_group }}"
       mode: "{{ kubernetes_service_mode }}"
 
-  - include: ../validate-pod/tasks/validate-pod.yaml  name="kube-proxy" selector="component=kube-proxy,kismatic/host={{ inventory_hostname }}"
+  - name: get the name of the kube-proxy pod running on this node
+    command: kubectl get pods -l=k8s-app=kube-proxy --template {%raw%}'{{range .items}}{{if eq .spec.nodeName{%endraw%} "{{ inventory_hostname|lower }}"{%raw%}}}{{.metadata.name}}{{"\n"}}{{end}}{{end}}'{%endraw%} -n kube-system
```
alexbrand: I am having a hard time understanding what is going on in the `--template` flag. Maybe a comment would help?

ansible/roles/kube-proxy/templates/kube-proxy.yaml
```diff
@@ -1,61 +1,109 @@
...
+  apiGroup: rbac.authorization.k8s.io
+---
+
+apiVersion: extensions/v1beta1
+kind: DaemonSet
 metadata:
   labels:
-    tier: control-plane
-    component: kube-proxy
-    kismatic/host: {{ inventory_hostname }}
```
alexbrand: Do you recall why we added these? I unfortunately don't remember.

dkoshkin: I think it was meant to be used as a selector, although I dont its actually correct as we run `kubectl apply` once and all the pods will have the same label.

I will check if we use these anywhere and remove.

dkoshkin: Scratch that, this does get used as selectors when validating the control plane pods and because they are all deployed individually on the nodes.

ansible/roles/kube-proxy/tasks/main.yaml
alexbrand: Would it make sense to use the hostname label to get the kube-proxy pod, instead of using the template? What's the `\n` in there for?

dkoshkin: The `hostname` label will not be accurate since `kubectly apply` will set the labels on all of the nodes with the same `hostname`.

When I originally worked on this I found a discussion on how to do this, I don't remember why its there :(

dkoshkin:
Merging as the failures are unrelated. Will fix all these errors when all PRs are merged in for the release.

PR #983, opened 2017-12-15
Add provisioner support for Azure

emmetthitz:
There's still some obvious things that need to be changed within the HCL/provision package, but I figured I'd get this up just to get some eyes on it.


emmetthitz:
image
👿

pkg/install/plan_types.go
```diff
@@ -79,18 +79,29 @@ type Plan struct {
 type Provisioner struct {
 	// The provider where the infrastructue will be provisioned to.
 	// The provisioner will expect provider specific ENV variables to be set.
-	// Options: aws
+	// Options: aws, azure, ""
 	Provider string
 	// AWS specific options.
 	// Only set if using "aws" provider
-	AWSOptions *AWSProvisionerOptions `yaml:"options,omitempty"`
+	AWSOptions *AWSProvisionerOptions `yaml:"aws_options,omitempty"`
```
emmetthitz: @dkoshkin this is what I was referring to

dkoshkin: That's too bad, I was really hoping that it would work. At runtime is one of these always `nil` or maybe they are getting some default empty values?

emmetthitz: Yeah, one of these should always be nil. The marshaller would just panic and the error would report a duplicate key.

alexbrand: Would something like this work:
```
{
  "provider": "aws"
  "options": {
    "aws": {
      "region": "us-east-1"
    }
  }
}
```

terraform/providers/azure/main.tf
```diff
@@ -0,0 +1,15 @@
+provider "azurerm" {
+  /*
+  $ export ARM_SUBSCRIPTION_ID = your_azure_sub_id
+  $ export ARM_CLIENT_ID = your_client_id
+  $ export ARM_CLIENT_SECRET = your_client_secret
+  $ export ARM_TENANT_ID= your_tenant_id
+  */
+  subscription_id = "${var.sub_id}"
```
alexbrand: We should grab these secrets from the environment, instead of vars.

terraform/providers/azure/securitygroups.tf
```diff
@@ -0,0 +1,181 @@
...
+    protocol                   = "*"
+    source_port_range          = "*"
+    destination_port_range     = "*"
+    source_address_prefix      = "10.0.0.0/16"
+    // The address space of the resource group
+    destination_address_prefix = "*"
+  }
+  security_rule {
+    name                       = "${var.cluster_name}-private-out"
+    description                = "Allow all communication between nodes."
```
alexbrand: This description is misleading to me. Should this be `Allow all outbound communications from nodes` instead?

terraform/providers/azure/securitygroups.tf
```diff
@@ -0,0 +1,181 @@
...
+    access                     = "Allow"
+    protocol                   = "TCP"
+    source_port_range          = "6443"
+    destination_port_range     = "6443"
+    source_address_prefix      = "*"
+    destination_address_prefix = "*"
+  }
+  security_rule {
+    name                       = "${var.cluster_name}-private-in"
+    description                = "Allow all communication between nodes."
```
alexbrand: Is this required in the load balancer's security group?

terraform/providers/azure/securitygroups.tf
```diff
@@ -0,0 +1,181 @@
...
+    protocol                   = "*"
+    source_port_range          = "*"
+    destination_port_range     = "*"
+    source_address_prefix      = "10.0.0.0/16"
+    // The address space of the resource group
+    destination_address_prefix = "*"
+  }
+  security_rule {
+    name                       = "${var.cluster_name}-private-out"
+    description                = "Allow all communication between nodes."
```
alexbrand: Is this required in the load balancer's security group?

terraform/providers/azure/securitygroups.tf
```diff
@@ -0,0 +1,181 @@
...
+    access                     = "Allow"
+    protocol                   = "TCP"
+    source_port_range          = "443"
+    destination_port_range     = "443"
+    source_address_prefix      = "*"
+    destination_address_prefix = "*"
+  }
+    security_rule {
+    name                       = "${var.cluster_name}-private-in"
+    description                = "Allow all communication between nodes."
```
alexbrand: Is this required in the ingress load balancer's security group?

terraform/providers/azure/securitygroups.tf
```diff
@@ -0,0 +1,181 @@
...
+    protocol                   = "*"
+    source_port_range          = "*"
+    destination_port_range     = "*"
+    source_address_prefix      = "10.0.0.0/16"
+    // The address space of the resource group
+    destination_address_prefix = "*"
+  }
+  security_rule {
+    name                       = "${var.cluster_name}-private-out"
+    description                = "Allow all communication between nodes."
```
alexbrand: Is this required in the ingress load balancer's security group?

terraform/providers/azure/variables.tf
```diff
@@ -0,0 +1,72 @@
...
+variable "ssh_user" {
+  default = ""
+}
+
+variable "version" {
+  default = ""
+}
+
+variable "cluster_name" {
+  default = "kismatic-cluster"
```
alexbrand: We don't want defaults on variables that we expect from the user.

terraform/providers/azure/variables.tf
```diff
@@ -0,0 +1,72 @@
...
+  default = ""
+}
+
+variable "instance_size" {
+  default = "Standard_D1"
+}
+
+variable master_count {
+  description = "Number of k8s master nodes"
+  default     = 1
```
alexbrand: We don't want defaults on variables that we expect from the user.

terraform/providers/azure/variables.tf
```diff
@@ -0,0 +1,72 @@
...
+}
+
+variable etcd_count {
+  description = "Number of etcd nodes"
+  default     = 1
```
alexbrand: We don't want defaults on variables that we expect from the user.

terraform/providers/azure/variables.tf
```diff
@@ -0,0 +1,72 @@
...
+}
+
+variable worker_count {
+  description = "Number of k8s worker nodes"
+  default     = 1
```
alexbrand: We don't want defaults on variables that we expect from the user.

terraform/providers/azure/variables.tf
```diff
@@ -0,0 +1,72 @@
...
+}
+
+variable ingress_count {
+  description = "Number of k8s ingress nodes"
+  default     = 1
```
alexbrand: We don't want defaults on variables that we expect from the user.

terraform/providers/azure/variables.tf
```diff
@@ -0,0 +1,72 @@
...
+}
+
+variable storage_count {
+  description = "Number of k8s storage nodes"
+  default     = 1
```
alexbrand: We don't want defaults on variables that we expect from the user.

terraform/providers/azure/variables.tf
```diff
@@ -0,0 +1,72 @@
+variable "region" {
+  default = "East US"
```
alexbrand: We don't want defaults on variables that we expect from the user.

pkg/install/plan_types.go
emmetthitz: I suppose so. That also feels kind of clunky though, don't you think?

pkg/provision/azure_types.go
```diff
@@ -0,0 +1,30 @@
...
+	Terraform
+	SubscriptionID string
+	ClientID       string
+	ClientSecret   string
+	TenantID       string
+}
+
+// AzureTerraformData provider for creating and destroying infrastructure on Azure
+type AzureTerraformData struct {
+	SubscriptionID    string `json:"sub_id"`
```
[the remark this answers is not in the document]
emmetthitz: It will, but there's something funny going on with the terraform when the azurerm provider is set.

terraform/providers/azure/arg.tf
```diff
@@ -0,0 +1,53 @@
+resource "azurerm_resource_group" "kismatic" {
+  name     = "${var.cluster_name}"
+  location = "${var.region}"
+  tags {
+    "Name"                  = "${var.cluster_name}-rg"
+    "kismatic.clusterName"  = "${var.cluster_name}"
```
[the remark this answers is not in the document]
emmetthitz: No. Forward slashes are a reserved character for azurerm tags.

terraform/providers/azure/arg.tf
```diff
@@ -0,0 +1,53 @@
...
+    "kismatic.clusterOwner" = "${var.cluster_owner}"
+    "kismatic.dateCreated"  = "${timestamp()}"
+    "kismatic.version"      = "${var.version}"
+    "kubernetes.io.cluster" = "${var.cluster_name}"
```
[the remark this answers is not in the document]
emmetthitz: If it could, I'd think 😉

Otherwise, I'd be in favor of removing it.

terraform/providers/azure/instances.tf
```diff
@@ -0,0 +1,361 @@
...
+  }
+}
+
+resource "azurerm_virtual_machine" "master" {
+  depends_on            = ["azurerm_resource_group.kismatic", "azurerm_network_interface.master"]  
+  count                 = "${var.master_count}"
+  name                  = "${var.cluster_name}-master-${count.index}"
+  location              = "${azurerm_resource_group.kismatic.location}"
+  resource_group_name   = "${azurerm_resource_group.kismatic.name}"
+  network_interface_ids = ["${azurerm_network_interface.master.*.id}"]
```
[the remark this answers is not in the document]
emmetthitz: Oh whoops.

terraform/providers/azure/main.tf
emmetthitz: If it actually did, yeah. I agree. Terraform seems to expect an azurecli login token as it's priority, so if one of those isn't set, it will just fail. Thankfully, anything in the hcl takes priority over environment vars, which is why the previous bit of this you commented on exists.

terraform/providers/azure/securitygroups.tf
emmetthitz: 👍

terraform/providers/azure/variables.tf
emmetthitz: Yes. I have this here as this is what we also do on aws. It should probably be removed there as well.

terraform/providers/azure/securitygroups.tf
emmetthitz: I believe so? I didn't manage to test without it, but I can do that now and see what happens.

terraform/providers/azure/lb.tf
```diff
@@ -0,0 +1,123 @@
+resource "azurerm_lb" "kismatic_master" {
```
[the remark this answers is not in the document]
emmetthitz: It should, but it doesn't. I need to pull it apart some more as well.

pkg/provision/azure_types.go
```diff
@@ -0,0 +1,30 @@
+package provision
+
+// Azure provisioner for creating and destroying infrastructure on Azure.
+type Azure struct {
+	Terraform
+	SubscriptionID string
```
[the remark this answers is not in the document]
emmetthitz: Oh whoops. This should've been flowed through from the CLI/server. Fixed.

pkg/install/plan_types.go
alexbrand: Yeah, disregard.

terraform/providers/azure/lb.tf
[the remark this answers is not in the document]
emmetthitz: Ended up needing availability sets anyway. Fixed.

pkg/install/plan_types.go
```diff
@@ -79,18 +79,29 @@ type Plan struct {
...
-	Region string `yaml:"region`
+	Region string `yaml:"region"`
+	//TODO: add AZ options, and make machine-type configurable.
+}
+
+// AazureProvisionerOptions contains specific options used when provisioning infrastructue Azure
+type AzureProvisionerOptions struct {
+	// The Azure region to deploy in
+	// +required
+	Location string `yaml:"region"`
```
alexbrand: Is the Azure term "Location" or "Region"? I think the struct field and the yaml tag should be consistent.

terraform/providers/azure/nic.tf
```diff
@@ -0,0 +1,169 @@
+resource "azurerm_network_interface" "bastion" {
+  internal_dns_name_label   = "${var.cluster_name}"
```
alexbrand: Should this have a `-bastion` suffix?

pkg/install/plan_types.go
emmetthitz: Azure uses "location" instead of region. Good catch.

terraform/providers/azure/nic.tf
emmetthitz: Yes!

PR #986, opened 2017-12-15
Add flag to server cmd to disable TLS

### LINKED ISSUE #985: Server mode: Add support for serving API over HTTP
Currently, the server can only be accessed via HTTPS. This makes it cumbersome to work with during dev and test.

Add support for accessing the API via HTTP. This should be disabled by default.

alexbrand:
Fixes #985

### DIFF
--- a/pkg/cli/server.go
+++ b/pkg/cli/server.go
@@ -23,17 +23,19 @@ import (
 )
 
 const (
-	loggerPrefix   = "[kismatic] "
-	defaultTimeout = 10 * time.Second
-	clustersBucket = "kismatic"
-	assetsFolder   = "assets"
+	defaultTimeout      = 10 * time.Second
+	clustersBucket      = "kismatic"
+	assetsFolder        = "assets"
+	defaultInsecurePort = "8080"
+	defaultSecurePort   = "8443"
 )
 
 type serverOptions struct {
-	port     string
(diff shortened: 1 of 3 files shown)


dkoshkin:
Do you think its ok to keep the default port `443` [here](https://github.com/apprenda/kismatic/blob/c6366285f6d7ef324a5c6b85ff6d211fdf71d286/pkg/cli/server.go#L60)?

alexbrand:
Thanks @dkoshkin. Changed the defaults.

pkg/cli/server.go
```diff
@@ -94,10 +96,15 @@ func doServer(stdout io.Writer, options serverOptions) error {
 	// create handlers
 	clusterAPI := handler.Clusters{Store: clusterStore, AssetsDir: assetsDir, Logger: logger}
 
+	port := "8443"
```
dkoshkin: Don't we also need to check if `options.port` is not empty and use that?

alexbrand: Whoops! Yes, I basically rendered that flag useless.

dkoshkin: :) I got you

dkoshkin:
LGTM!

emmetthitz:
LGTM

PR #1007, opened 2017-12-19
Persist provider-specific options in store

### LINKED ISSUE #984: Server mode: Persist provider specific options and return them over API
Part of the fix for #962 involved removing the plan from the struct that is being persisted in the store. The provider specific options that can be set through the API were being stored in the plan, so we must now store them in the actual cluster specification instead.

alexbrand:
Fixes #984 

In order to not get into deserialization and type assertion hell, I made the options field of the ClusterRequest a `map[string]string`. This is the initial step towards the long term plan of supporting providers without having to recompile KET.

With that said, we still have provider specific code as we do not have the required framework built yet. The structs are basically replaced by provider-specific constants that define the keys we expect in the options map. These constants are used in validation and when transforming the request into a cluster spec. More importantly, the cluster spec's provider field is still cloud-provider specific. This is to avoid unnecessary instability until we fully figure out our approach for "any-provider" provisioning.

### DIFF
--- a/pkg/store/cluster.go
+++ b/pkg/store/cluster.go
@@ -34,12 +34,21 @@ type ClusterStatus struct {
 type Provisioner struct {
 	Provider    string
 	Credentials ProvisionerCredentials
+	Options     ProvisionerOptions
 }
 
 // ProvisionerCredentials are the credentials necessary for connecting to the
 // infrastructure provisioner.
 type ProvisionerCredentials struct {
 	AWS AWSCredentials
+}
+
+type ProvisionerOptions struct {
+	AWS AWSProvisionerOptions
+}
+
+type AWSProvisionerOptions struct {
+	Region string
 }
 
 // AWSCredentials are the credentials for interacting with the AWS API.
--- a/integration/daemon_test.go
+++ b/integration/daemon_test.go
@@ -81,9 +81,10 @@ The error: %v
 					IngressCount: 1,
 					Provisioner: handler.Provisioner{
 						Provider: "aws",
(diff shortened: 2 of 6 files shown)


pkg/store/cluster.go
```diff
@@ -34,6 +34,7 @@ type ClusterStatus struct {
 type Provisioner struct {
```
dkoshkin: Do we need both the struct here and in `plan_types.go`? Is there anyway to reuse one?

alexbrand: We can't quite get rid of the Provisioner struct here. The credentials are not part of the plan struct in the install pkg, so I think we need this.

emmetthitz: What about the one from the provision package?

alexbrand: If we use the AWS provisioner options type in the `install` pkg, we'll need to add json tags to it. I don't think I can tell which is better at this point. 

The one in the provision package is not suitable for this, as it's used for a completely different purpose. Furthermore, doesn't make sense to me for the store pkg to depend on the provision pkg.

emmetthitz: I'm not sure I follow the reasoning for having it depend on the install package then either.

alexbrand: Dependency on install pkg doesn't quite make sense either, which is why I removed that dependency in this PR. However, that's where the "cluster config API" lives at this point, so the only reason I see for using the type defined in install is to avoid duplication.

dkoshkin:
Tested this manually and LGTM.

PR #1008, opened 2017-12-19
Remove defaults from AWS terraform vars

alexbrand:
Removed defaults from AWS terraform vars that we expect KET to set.

Also removed the secrets from the vars definition, so that they are never stored in the binary plan.


emmetthitz:
LGTM

PR #1009, opened 2017-12-19
Prevent terraform from attempting to read vars from stdin

alexbrand:
Set the `-input` flag to `false` to prevent terraform for asking for any unset variable.


emmetthitz:
This will cause failure at runtime if something isn't set the way we expect it to then, if I understand correctly, right?

alexbrand:
That is my understanding, yes.

emmetthitz:
LGTM then.

PR #1012, opened 2017-12-21
Send terraform logs to cluster log file, instead of stdout

alexbrand:
Both the provisioner and the installer (aka. executor) need to be able to write logs to the log file that the API server returns for the logs endpoint. For this reason, something other than the executor needs to create the cluster assets dir and the log file. 

Given that the multi-cluster controller is responsible for setting up the individual cluster controllers, it seems like it should also be responsible for setting up the cluster-specific assets directory and log file. Failure to create these results in the cluster controller not being created. Once created, the provisioner and installer both have access to these.

The overall outcome of this change is that the terraform logs are now visible through the API, so any provisioning errors are observable by the user.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -3,14 +3,15 @@ package controller
 import (
 	"context"
 	"fmt"
+	"io"
 	"log"
-	"os"
 	"path/filepath"
 	"time"
 
 	"github.com/apprenda/kismatic/pkg/install"
 	"github.com/apprenda/kismatic/pkg/provision"
 	"github.com/apprenda/kismatic/pkg/store"
+	"github.com/blang/semver"
 )
 
 // The ClusterController manages the lifecycle of clusters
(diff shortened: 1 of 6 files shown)


pkg/controller/controller.go
```diff
@@ -79,16 +73,20 @@ func DefaultExecutorCreator() ExecutorCreator {
...
-// DefaultProvisionerCreator uses terraform for provisioning infrastructure
+// TerraformProvisionerCreator uses terraform for provisioning infrastructure
 // on the clouds we support.
-func DefaultProvisionerCreator(terraform provision.Terraform) ProvisionerCreator {
-	return func(cluster store.Cluster) provision.Provisioner {
+func TerraformProvisionerCreator(binaryPath string, kismaticVersion semver.Version) ProvisionerCreator {
+	return func(cluster store.Cluster, output io.Writer) provision.Provisioner {
 		switch cluster.Spec.Provisioner.Provider {
 		case "aws":
 			p := provision.AWS{
```
emmetthitz: Why does it make sense to have the server depend on this work, but not the store? 
Referencing https://github.com/apprenda/kismatic/pull/1007/files#diff-8c8941fddd62bac74401c4ac526abbe9L34

dkoshkin:
LGTM

pkg/controller/controller.go
alexbrand: Discussed this on call. The controller needs to know how to install a cluster.

emmetthitz: Yeah, disregard.

PR #1018, opened 2017-12-24
Upgrading Calico node from 2.6.4 and 2.6.5.

### LINKED ISSUE #1017: Upgrade to Calico 2.6.5
Calico released 2.6.5 as it resolves an issue which may lead to a brief loss of connectivity while upgrading to v3.0.0 from v2.6.4 when using an etcd datastore. See [here](https://github.com/projectcalico/felix/pull/1676).

swade1987:
Closes #1017 

Based off the Calico 2.6.4 release notes (see below)

> Important: Due to a known issue in this release that can cause potential
> brief losses of connectivity while upgrading from v2.6.4, this release is
> deprecated. Use v2.6.5 instead. This issue does not affect those using the
> Kubernetes API datastore or running in policy-only mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ KET operational tools include:
 | Docker | v1.12.6 |
 | Etcd (for Kubernetes) | v3.1.10 |
 | Etcd (for Calico & Contiv) | v3.1.10 |
-| Calico | v2.6.2 |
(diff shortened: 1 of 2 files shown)


dkoshkin:
They need to stop releasing :)

PR #1019, opened 2017-12-26
Copy inspector to all nodes before upgrade

dkoshkin:
The `inspector` service needs to be updated and restarted before upgrading any of the individual nodes.


alexbrand:
LGTM!

PR #1021, opened 2017-12-26
Always lower hostname with kubernetes.io/hostname label

dkoshkin:
Fixes `hosts file modification feature` test.

Node label `kubernetes.io/hostname` will be the same as `nodeName`, and it's always lower cased.


alexbrand:
LGTM

PR #1022, opened 2017-12-27
Update source versions for upgrade tests for v1.7.0

alexbrand:
According to our versioning guidelines, we must test upgrades from the previous minor, latest patch.

Updated our upgrade integration tests.


dkoshkin:
LGTM!

PR #1024, opened 2017-12-27
Upgrade to using Golang (1.9.2) and Glide (v0.13.1)

swade1987:


emmetthitz:
What do we still need to merge this?

swade1987:
Lets merge this into the 1.8.0 release branch?

dkoshkin:
During my testing locally it was always getting stuck `Gathering Facts`, needs further investigation.

ansible/callback/json_lines.py
```diff
@@ -54,7 +54,7 @@ def _print_event(self, event):
 
     def __init__(self):
         named_pipe_file = os.environ["ANSIBLE_JSON_LINES_PIPE"]
-        self.named_pipe = open(named_pipe_file, 'w')
```
dkoshkin: Finally fixes a tricky intermittent bug we have always seen but has really broke with Go 1.9 on OSX.
The pipe is used by Ansible to send events to the Go process. The Go process reads the pipe and parses the data displaying it to the user. 
The Go code uses a buffered line reader. Before this change Python was also using a buffer to write. The size of that buffer is dependent on the system and that is why this issue was surfacing on OSX but not on Linux in our integration tests. The writes to the buffer would fill up and the Python would block, on the other side of the pipe Go would wait for a newline character to come, a deadlock.

To debug this issue I would run the installer and let it get "stuck" I then looked at the Goroutines, showing that both were waiting. I also read from the pipe with `cat` that would cause the program to get unstuck.

emmetthitz: the man, the myth, the dimitri koshkin

dkoshkin:
LGTM!

PR #1025, opened 2017-12-27
Removing implemented proposals.

swade1987:


alexbrand:
What's the motivation behind this? These are a record of design decisions and dev documentation that I think should be kept around.

swade1987:
I think the word proposals isn't an accurate reflection.

Happy to move them to another directory called `design-decisions` if necessary.

swade1987:
Removing as have re-worked this in PR #1029

PR #1030, opened 2017-12-29
Adding pod anti-affinity to the kube-dns deployment.

### LINKED ISSUE #490: kubedns Replica Set -> DaemonSet
I would advice to make de kubedns pod a Daemonset and let it scale on all nodes in the cluster, this to spread the load on deployments which rely heavy on internal DNS lookups.

swade1987:
Fixes #490 

cc @jarkomes

### DIFF
--- a/ansible/roles/kube-dns/templates/kubernetes-dns.yaml
+++ b/ansible/roles/kube-dns/templates/kubernetes-dns.yaml
@@ -73,6 +73,18 @@ spec:
       tolerations:
       - key: "CriticalAddonsOnly"
         operator: "Exists"
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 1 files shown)


ansible/roles/kube-dns/templates/kubernetes-dns.yaml
```diff
@@ -73,6 +73,16 @@ spec:
       tolerations:
       - key: "CriticalAddonsOnly"
         operator: "Exists"
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
```
dkoshkin: IMO this should be `preferredDuringSchedulingIgnoredDuringExecution` we validate that the pods come up and in certain situations this might not work.

dkoshkin:
LGTM!

Issue #770, opened 2017-09-05
Support the overlay2 docker storage driver


aquasheep20:
any plans for kismatic support of overlay2 for Redhat? rhel 7.4 docker 1.12.6 docker can now use overlay2

alexbrand:
Hey @aquasheep20,

We don't plan to in the short term, but we might bump it up in priority given that overlay2 is preferred over devicemapper. 

One thing that needs more thought is how to handle this during an upgrade. I haven't looked into it, but I imagine there is some migration that needs to happen from devicemapper to overlay2.

swade1987:
Also, you need to disable SELinux to use overlay and overlay2 on RHEL or CentOS.

aquasheep20:
i've got kismatic running on RH 7.4 with overlay2 on a new cluster.  remove docker-engine, install RH docker, disable selinux, disable container-storage-setup (echo STORAGE_DRIVER="" >> /etc/sysconfig/docker-storage-setup), add "storage-driver": "overlay2" to /etc/docker/daemon.json, add --cgroup-driver=systemd to /etc/systemd/system/kubelet.service.  make sure you push all your own images before you swap docker, & pull them back post swap.  All the k8s  images will be re pulled on their own.  This will also leave your cluster in a precarious situation for a kismatic upgrade.

mrgleeco:
+1 to overlay2 options- as a general thing.   Would like to use on xenial.  Separate Issue?

alexbrand:
Thanks @mrgleeco. I think we can keep both RHEL and Ubuntu in this issue.

swade1987:
@aquasheep20 what version of Docker are you using with overlay2 also how do you find the performance?

aquasheep20:
I'm using Ubuntu 16.04 with whatever docker kismatic is installing.  You can set overlay2 by creating  /etc/docker/daemon.json before installing kismatic.   

Mine for reference:  
{
  "storage-driver": "overlay2",
  "icc": false,
  "disable-legacy-registry": true,
  "live-restore": true,
  "userland-proxy": false
  }

swade1987:
@aquasheep20 you mentioned RHEL 7.4 up above, is this not the case?

Regardless, pretty cool it was that straightforward to configure.

aquasheep20:
[...]

PR #1032, opened 2018-01-01
Support the overlay2 docker storage driver

swade1987:
Implements: #770 

**Note: This is currently a work-in-progress.**

The story thus far:

- I have tested overlay2 for Ubuntu 16.04 and CentOS 7.4.1708

Things to complete:

- Comments in the plan file that overlay2 will be the default storage driver
- Comments in the plan file that directlvm only needs to be enabled for RHEL 7.3 and below
- Comments in the plan file that directlvm only needs to be enabled for CentOS 7.3 and below
- Add integration tests for the implementation on RHEL, CentOS and Ubuntu versions
- Add integration tests for upgrading from devicemapper (with directlvm) to overlay2

cc @aquasheep20

### DIFF
--- a/ansible/roles/docker/tasks/main.yaml
+++ b/ansible/roles/docker/tasks/main.yaml
@@ -11,23 +11,26 @@
 
   # The overlay2 docker storage driver will be used for the following operating systems:
   # - RHEL 7.4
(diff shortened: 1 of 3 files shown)


dkoshkin:
Closing until it is fully implemented.

PR #1037, opened 2018-01-03
Correctly set Content-Type Header to json

dkoshkin:
The header was not being set correctly after writing the response. 
Also changed json marshaling to be consistent for all methods.


emmetthitz:
Lgtm

PR #1040, opened 2018-01-04
Destruction prevention

### LINKED ISSUE #980: Provisioning: Destruction of infrastructure should only be possible through a destroy operation
As we add provisioning capabilities to KET, we need to be careful about any destruction operations that Terraform might decide to perform.

We need to find a way to prevent terraform from doing any destructive operations when creating or modifying existing infrastructure. The only acceptable time for destroying infrastructure is when the user has requested the destruction of their cluster.

This imposes a limitation on scaling, because we won't be able to reduce the number of nodes that make up the cluster. With that said, this is an acceptable limitation for the initial release of this feature.

emmetthitz:
Fixes #980

### DIFF
--- a/pkg/cli/provision.go
+++ b/pkg/cli/provision.go
@@ -15,6 +15,7 @@ import (
 
 // NewCmdProvision creates a new provision command
 func NewCmdProvision(in io.Reader, out io.Writer, opts *installOpts) *cobra.Command {
+	provisionOpts := provision.ProvisionOpts{}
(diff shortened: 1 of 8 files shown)


dkoshkin:
Can we modify the integration test to scale up without setting the flag making sure its working and then scale down and make sure it fails. 

Just thinking for the future if TF changes the output it would be good to catch that.

emmetthitz:
👍

PR #1042, opened 2018-01-05
Fixes for labels appearing in the plan file for etcd nodes.

emmetthitz:
Exactly what it looks like.

### DIFF
--- a/pkg/install/plan.go
+++ b/pkg/install/plan.go
@@ -186,11 +186,21 @@ func (fp *FilePlanner) Write(p *Plan) error {
 	scanner := bufio.NewScanner(bytes.NewReader(bytez))
 	prevIndent := -1
 	addNewLineBeforeComment := true
+	var etcdBlock bool
 	for scanner.Scan() {
 		text := scanner.Text()
 		matched := yamlKeyRE.FindStringSubmatch(text)
 		if matched != nil && len(matched) > 1 {
 			indent := strings.Count(matched[0], " ") / 2
+
+			// Figure out if we are in the etcd block
+			if indent == 0 {
+				etcdBlock = (text == "etcd:")
+			}
+			// Don't print labels: {} for etcd group
+			if etcdBlock && strings.Contains(text, "labels: {}") {
+				continue
+			}
 
 			// Add a new line if we are leaving a major indentation block
 			// (leaving a struct)..
--- a/pkg/install/test/plan-template-with-storage.golden.yaml
+++ b/pkg/install/test/plan-template-with-storage.golden.yaml
@@ -197,17 +197,14 @@ etcd:
   - host: ""
     ip: ""
     internalip: ""
-    labels: {}
 
   - host: ""
     ip: ""
     internalip: ""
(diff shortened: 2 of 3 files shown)


swade1987:
@emmetthitz legend, thanks dude!

dkoshkin:
Have you explored only modifying the representation layer?

A simpler approach would be to just not print the labels when we are writing out the plan file.
In `plan.go` 
```
// Write the plan to the file system
func (fp *FilePlanner) Write(p *Plan) error {
...
	// the stack keeps track of the object we are in
	// for example, when we are inside cluster.networking, looking at the key 'foo'
	// the stack will have [cluster, networking, foo]
	s := newStack()
	scanner := bufio.NewScanner(bytes.NewReader(bytez))
	prevIndent := -1
	addNewLineBeforeComment := true
	var etcdBlock bool
	for scanner.Scan() {
		text := scanner.Text()
		matched := yamlKeyRE.FindStringSubmatch(text)
		if matched != nil && len(matched) > 1 {
			indent := strings.Count(matched[0], " ") / 2

			// Figure out if we are in the etcd block
			if indent == 0 {
				etcdBlock = (text == "etcd:")
			}
			// Don't print labels: {} for etcd group
			if etcdBlock && strings.Contains(text, "labels: {}") {
				continue
			}
...
}
```

We are doing a similar thing with `kubelet.option_overrides: {}` but for all the nodes, it is basically a no-op if a user passes in these options for `etcd` as they are not used anywhere.

emmetthitz:
I didn't, but that makes more sense than doing it like this. I'll take a look.

emmetthitz:
Oh bless you dima

emmetthitz:
Feels like more of a hack, but otherwise LGTM. Probably indicative of an overdue refactor.

pkg/install/plan.go
```diff
@@ -186,12 +186,22 @@ func (fp *FilePlanner) Write(p *Plan) error {
...
 	addNewLineBeforeComment := true
+	var etcdBlock bool
 	for scanner.Scan() {
 		text := scanner.Text()
 		matched := yamlKeyRE.FindStringSubmatch(text)
 		if matched != nil && len(matched) > 1 {
 			indent := strings.Count(matched[0], " ") / 2
 
+			// Figure out if we are in the etcd block
+			if indent == 0 {
```
emmetthitz: Actually, the more I think about it... is this necessary?

dkoshkin: Not really in the current state, but its one of those things where if the plan file ever changes to have `etcd` somewhere else 100% we would forget to modify this.

dkoshkin:
>Feels like more of a hack, but otherwise LGTM. Probably indicative of an overdue refactor.

Agree, Im not a fan that the docs still contain the `labels` section but I dont think the effort to properly fix it is worth it with the uncertainty of how etcd is going to be run in the future.

pkg/install/plan.go
emmetthitz: Can we pull the indentation check then?

PR #1048, opened 2018-01-11
Decouple Kubernetes Version from the releases of KET

dkoshkin:
Implementation for https://github.com/apprenda/kismatic/pull/1038

### DIFF
--- a/docs/plan-file-reference.md
+++ b/docs/plan-file-reference.md
@@ -2,6 +2,7 @@
 ## Index
 * [cluster](#cluster)
   * [name](#clustername)
+  * [version](#clusterversion)
   * [admin_password _(deprecated)_](#clusteradmin_password-deprecated)
   * [disable_package_installation](#clusterdisable_package_installation)
   * [allow_package_installation _(deprecated)_](#clusterallow_package_installation-deprecated)
@@ -162,6 +163,16 @@
 | **Kind** |  string |
 | **Required** |  Yes |
 | **Default** | ` ` | 
+
+###  cluster.version
+
+ The Kubernetes version to install. If left blank will be set to the latest tested version. Only a single Minor version is supported with. 
+
+| | |
+|----------|-----------------|
+| **Kind** |  string |
+| **Required** |  No |
+| **Default** | `v1.9.2` | 
 
 ###  cluster.admin_password _(deprecated)_
 
--- a/pkg/inspector/cmd/client.go
+++ b/pkg/inspector/cmd/client.go
@@ -4,17 +4,19 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/apprenda/kismatic/pkg/inspector"
 	"github.com/spf13/cobra"
 )
 
 type clientOpts struct {
-	outputType         string
-	nodeRoles          string
-	rulesFile          string
-	targetNode         string
-	useUpgradeDefaults bool
+	outputType          string
+	nodeRoles           string
+	rulesFile           string
+	targetNode          string
+	useUpgradeDefaults  bool
+	additionalVariables map[string]string
 }
 
 var clientExample = `# Run the inspector against an etcd node
@@ -29,6 +31,7 @@ kismatic-inspector client 10.0.1.24:9090 -f inspector-rules.yaml --node-roles et
 // NewCmdClient returns the "client" command
 func NewCmdClient(out io.Writer) *cobra.Command {
 	opts := clientOpts{}
+	var additionalVars []string
 	cmd := &cobra.Command{
 		Use:     "client HOST:PORT",
 		Short:   "Run the inspector against a remote inspector server.",
@@ -39,6 +42,14 @@ func NewCmdClient(out io.Writer) *cobra.Command {
 			}
 			// Set the target node as the first argument
 			opts.targetNode = args[0]
+			opts.additionalVariables = make(map[string]string)
+			for _, v := range additionalVars {
+				kv := strings.Split(v, "=")
+				if len(kv) != 2 {
+					return fmt.Errorf("invalid key=value %q", v)
+				}
+				opts.additionalVariables[kv[0]] = kv[1]
+			}
 			return runClient(out, opts)
 		},
 	}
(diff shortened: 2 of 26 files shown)


swade1987:
@dkoshkin would be great to see some CLI output examples of the following:

- Attempting to go from 1.8.x to 1.9.x
- Attempting to go from 1.9.0 to 1.9.x
- Attempting to go from 1.9.x to 1.8.x
- Attempting to go from 1.9.x to 1.9.0

dkoshkin:
What do you mean `go from`, upgrades?

emmetthitz:
@swade1987 I don't think it would be any different than a typical upgrade. Should just be a `./kismatic upgrade`

swade1987:
@emmetthitz  agreed but just wanted to make sure. 

Also wanted to make sure we have a list of scenarios we are testing against.

Those are the ones I came up with off the top of my head.

pkg/inspector/cmd/client.go
```diff
@@ -39,13 +42,22 @@ func NewCmdClient(out io.Writer) *cobra.Command {
 			}
 			// Set the target node as the first argument
 			opts.targetNode = args[0]
+			opts.additionalVariables = make(map[string]string)
```
emmetthitz: I'm not sure how I feel about this. What was the motivation behind it?

pkg/install/plan_types.go
```diff
@@ -85,6 +85,10 @@ type Cluster struct {
 	// cluster name, such as kubeconfig files and certificates.
 	// +required
 	Name string
+	// The Kubernetes version to install.
+	// If left blank will be set to the latest tested version.
+	// +default=v1.9.1
+	Version string
```
emmetthitz: Does semver not serialize the way you'd expect?

pkg/install/version.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	latest := strings.Trim(string(body), " \t\n")
+	if !kubeReleaseRegex.MatchString(latest) {
+		return kubernetesVersion, kubernetesVersionString, fmt.Errorf("Invalid version format %q", latest)
+	}
+	parsedLatest, err := parseVersion(latest)
+	if err != nil {
+		return kubernetesVersion, kubernetesVersionString, fmt.Errorf("Could not parse version string %q", latest)
+	}
+	return parsedLatest, latest, nil
```
emmetthitz: Related to previous comment, feels weird to return a semver.Version and a string representation of the same thing.

docs/plan-file-reference.md
```diff
@@ -2,6 +2,7 @@
 ## Index
 * [cluster](#cluster)
   * [name](#clustername)
+  * [version](#clusterversion)
```
swade1987: I think should be `kubernetes_version` as it feels more accurate as you aren't versioning the cluster itself.

dkoshkin: This gets autogenerated and thats just the href link, its actually just `cluster.version` see [here](https://github.com/apprenda/kismatic/pull/1048/files/a7a7d8a7401419b1aa3982412837ec51065cad3e#diff-e93a0c402f6331148e311cfe5b9e94a0R91)

swade1987: I still hold it should really be `kubernetes_version`.

swade1987:
**1. Online Scenarios:** 

O1 - Performing `kismatic install plan` what is the default value set to?

O2 - Performing a `kismatic install apply` and setting a version (A.B.C) that doesn't exist yet.

O3 - Performing a `kismatic install apply` and setting a version of (A.B.C-alpha.D) 

O4 - Performing a `kismatic install upgrade` and setting a version (A.B.C) that doesn't exist yet.

O5 - Performing a `kismatic install upgrade` and setting a version of (A.B.C-alpha.D)

**2. Offline Scenarios:** 

OF1 - Performing `kismatic install plan` what is the default value set to?

OF2 - Performing a `kismatic install apply` and setting a version (A.B.C) that doesn't exist yet.

OF3 - Performing a `kismatic install apply` and setting a version of (A.B.C-alpha.D) 

OF4 - Performing a `kismatic install upgrade` and setting a version (A.B.C) that doesn't exist yet.

OF5 - Performing a `kismatic install upgrade` and setting a version of (A.B.C-alpha.D)


@willvrny not sure if you have any more scenarios to cover?

dkoshkin:
1. Online Scenarios:

O1 - Performing kismatic install plan what is the default value set to?
Plan file gets set to tested version
```
cluster:
  name: kubernetes

  # Kubernetes cluster version
  version: v1.9.1
```
O2 - Performing a kismatic install apply and setting a version (A.B.C) that doesn't exist yet.
```
➜  out git:(decouple-k8s) ./kismatic install validate

- Cluster version "v1.9.4" greater than the latest stable version "v1.9.2"
```
O3 - Performing a kismatic install apply and setting a version of (A.B.C-alpha.D)
Setting in plan file:
```
  # Kubernetes cluster version
  version: v1.9.3-beta.0
```
```
➜  out git:(decouple-k8s) ./kismatic install validate

- Cluster version "v1.9.3-beta.0" must be a valid version, ie "v1.9.1"
```
O4 - Performing a kismatic install upgrade and setting a version (A.B.C) that doesn't exist yet.
```
➜  out git:(decouple-k8s) ./kismatic upgrade online

Computing upgrade plan==============================================================
- Cluster version "v1.9.4" greater than the latest stable version "v1.9.2"
```
O5 - Performing a kismatic install upgrade and setting a version of (A.B.C-alpha.D)
```
➜  out git:(decouple-k8s) ./kismatic upgrade online

Computing upgrade plan==============================================================
- Cluster version "v1.9.3-beta.0" must be a valid version, ie "v1.9.1"
```
2. Offline Scenarios:
When `disconnected_installation` is set to `true`, the `cluster.version` string is not verified with `https://storage.googleapis.com/kubernetes-release/release`.

However the inspector will validate that the packages exist in the repos and fail if not.

pkg/install/plan_types.go
dkoshkin: semver does not support having the `v` prefix, we want to be able to properly set the k8s version that includes the `v` prefix

pkg/inspector/cmd/client.go
dkoshkin: We need to be able to pass in the version of k8s to test for(and other versions in the future). The inspector has different types of rules that a user can write, with this change the user is able to pass in arbitrary values and template them in the rules.

swade1987:
@dkoshkin like it!

What about if I set the version to 1.10.1 (which exists) if I am running 1.9.1.

What is the error message then?

dkoshkin:
In both cases when setting to 1.8.x or 1.10.x it would return 
```
➜  out git:(decouple-k8s) ./kismatic install validate

- Cluster version "v1.10.1" must be a valid version, ie "v1.9.1"
```

swade1987:
If I were a user I would find that error message misleading as it is valid.

It may be better to say something like ...

`Cluster version "v1.10.1" must be a valid version, ie "v1.9.x" this release doesn't support upgrading to a new minor release`

swade1987:
How about `Cluster version "v1.10.1" invalid, must be a valid 1.9.x version, ie "v1.9.1"`

swade1987:
How about the following command:

```
# Kubernetes cluster version (supported versions 1.9.x)
```

dkoshkin:
Modified the error messages as discussed
```
➜  kismatic install validate

- Cluster version "v1.9.5" invalid, the latest stable version is "v1.9.2"

➜  kismatic install validate

- Cluster version "v1.8.5" invalid, must be a valid "v1.9.x" version, ie "v1.9.2"
```

swade1987:
@dkoshkin LGTM!

PR #1049, opened 2018-01-11
Merge any provisioner

emmetthitz:
Picking up where @alexbrand left off.


alexbrand:
Let me know if I can help in any way!

emmetthitz:
Will squash on merge. Really not in the mood to fight with a rebase.

integration-tests/daemon_test.go
```diff
@@ -79,7 +80,7 @@ The error: %v
 					MasterCount:  1,
 					WorkerCount:  1,
 					IngressCount: 1,
-					Provisioner: handler.Provisioner{
```
dkoshkin: This isn't correct, we want to send the request from the handler struct not the store struct, something else must be wrong.

emmetthitz: I believe it is. The handler provision struct is only responsible for the response. The handler struct for the request uses the store struct.

dkoshkin: It should not, an API client should not know anything about the store.

pkg/server/http/handler/clusters.go
```diff
@@ -40,37 +33,36 @@ type Clusters struct {
...
-	WorkerCount  int         `json:"workerCount"`
-	IngressCount int         `json:"ingressCount"`
-	Provisioner  Provisioner `json:"provisioner"`
+	Name         string            `json:"name"`
+	DesiredState string            `json:"desiredState"`
+	EtcdCount    int               `json:"etcdCount"`
+	MasterCount  int               `json:"masterCount"`
+	WorkerCount  int               `json:"workerCount"`
+	IngressCount int               `json:"ingressCount"`
+	Provisioner  store.Provisioner `json:"provisioner"`
```
emmetthitz: @dkoshkin here

integration-tests/daemon_test.go
emmetthitz: I mean, they still don't. This is strictly for deserializing the json into our go native data types. This is ultimately the information that needs to end up in the store, so I think it makes sense (for now at least - until we do the root package refactor).

emmetthitz:
@dkoshkin what's the verdict?

pkg/server/http/handler/clusters.go
```diff
@@ -386,46 +374,25 @@ func buildStoreCluster(req ClusterRequest) store.Cluster {
...
-			Region: req.Provisioner.Options[awsOptionRegion],
-		}
+		Provisioner:  req.Provisioner,
 	}
 	return store.Cluster{
 		Spec: spec,
 	}
 }
 
 func buildResponse(name string, sc store.Cluster) ClusterResponse {
```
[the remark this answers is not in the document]
dkoshkin: After discussing further this is not needed.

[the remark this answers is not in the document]
emmetthitz: Thought I think it would be good to have this documented (what the requests should look like) so we don't have this concern moving forward.

pkg/cli/provision.go
```diff
@@ -103,40 +73,33 @@ func NewCmdDestroy(in io.Reader, out io.Writer, opts *installOpts) *cobra.Comman
...
 		},
 	}
 	return cmd
 }
+
+type environmentSecretsGetter struct{}
+
+// GetAsEnvironmentVariables returns a slice of the expected environment
+// variables sourcing them from the current process' environment.
+func (environmentSecretsGetter) GetAsEnvironmentVariables(clusterName string, expected map[string]string) ([]string, error) {
```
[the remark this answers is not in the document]
emmetthitz: Fixed

pkg/cli/provision.go
```diff
@@ -94,12 +94,18 @@ type environmentSecretsGetter struct{}
...
 	for _, expectedEnvVar := range expected {
 		val := os.Getenv(expectedEnvVar)
 		if val == "" {
-			return nil, fmt.Errorf("the %q environment variable is required", expectedEnvVar)
+			missing = true
+			missingVars = append(missingVars, expectedEnvVar)
 		}
 		vars = append(vars, fmt.Sprintf("%s=%s", expectedEnvVar, val))
 	}
+	if missing {
```
dkoshkin: Can we just check the length of the `missingVars` slice instead here.

emmetthitz: 👍 my C is showing

dkoshkin:
LGTM!

PR #1051, opened 2018-01-16
Nodes with docker already installed

### LINKED ISSUE #474: Support nodes with existing docker installation

### LINKED ISSUE #672: Suggestion : Validate should check for Docker on Node

dkoshkin:
Fixes #474, fixes #672 

Added a new option to `disable` docker installation, this allows operators to use nodes with any (supported by k8s) version/flavor of docker already installed.
```
# Docker daemon configuration of all cluster nodes
docker:

  # Set to true if docker is already installed and configured.
  disable: false
```
Note: when set to `true`, none of the `docker` or `docker_registry` options will have an effect on as all configuration is left to the user.


swade1987:
Will the pre-flight checks still validate if the version of docker on the node is supported?

dkoshkin:
No, as there is no real "supported" version just recommended. The only reason why a user would want to use this feature is if for some reason they found that the official, recommended version does not work for them and at that point we cannot put any constraints on them without knowing their motivation for using a different version/repo.

swade1987:
@dkoshkin valid, if the user sets `disable: true` what will the CLI render?

dkoshkin:
Nothing specific, Ansible will have 
```
...
Install Docker                                                                  [SKIPPED]
...
```

emmetthitz:
@swade1987 probably a skipped?

swade1987:
I would expect to see` Installing Docker [SKIPPED]`

dkoshkin:
@swade1987 I've updated the original comment.

swade1987:
@dkoshkin LGTM. do we have a regression test for having docker pre-installed?

dkoshkin:
Yep see [commit](https://github.com/apprenda/kismatic/pull/1051/commits/9f4cb84a9ea518c6d0e96d091355ab056a073f4e).

* Runs validation and expects a failure
* Installs docker manually
* Runs installation with `docker.disable = true`.

PR #1052, opened 2018-01-16
Prevent labels from being overwritten during mutations

### LINKED ISSUE #1039: Provisioner overrides labels when mutating clusters
If you spin up a cluster with the provisioner, it populates your plan file. If you add labels to nodes using the plan and apply the labels, then mutate the cluster (either by adding or removing nodes), your labels will be overwritten with nothing when provisioning completes.

emmetthitz:
Fixes #1039.

### DIFF
--- a/pkg/install/plan_types.go
+++ b/pkg/install/plan_types.go
@@ -505,9 +505,63 @@ type Node struct {
 	KubeletOptions KubeletOptions `yaml:"kubelet,omitempty"`
 }
 
-// Equal returns true of 2 nodes have the same host, IP and InternalIP
+// mapMerge merges two map[string]strings. If both maps have some identical key, the value from the first map provided will be used.
+func mapMerge(l, r map[string]string) map[string]string {
+	ret := make(map[string]string, 0)
+	for k, v := range l {
+		ret[k] = v
+	}
+	for k, v := range r {
+		if _, ok := ret[k]; ok {
+			continue
+		}
+		ret[k] = v
+	}
+	return ret
+}
+
+func Merge(l, r NodeGroup) NodeGroup {
+	var ret NodeGroup
+	lmin := len(l.Nodes)
+	rmin := len(r.Nodes)
+	if l.ExpectedCount <= len(l.Nodes) {
+		lmin = l.ExpectedCount
+	}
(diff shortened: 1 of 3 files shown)


dkoshkin:
Need to rebase this with just the commits for this PR

pkg/install/plan_types.go
```diff
@@ -505,9 +505,60 @@ type Node struct {
...
+				ret = append(ret, Node{
+					IP:         l[i].IP,
+					InternalIP: l[i].InternalIP,
+					Host:       l[i].Host,
+					Labels:     mapMerge(l[i].Labels, r[j].Labels),
+					KubeletOptions: KubeletOptions{
+						Overrides: mapMerge(l[i].KubeletOptions.Overrides, r[j].KubeletOptions.Overrides),
+					},
+				})
+				//significantly faster than just axing the single element
```
dkoshkin: I think preserving the order would be better in this case, especially since these slices will be pretty small.

pkg/install/plan_types.go
```diff
@@ -505,9 +505,60 @@ type Node struct {
...
+	for k, v := range r {
+		if _, ok := ret[k]; ok {
+			continue
+		}
+		ret[k] = v
+	}
+	return ret
+}
+
+func merge(l, r []Node) []Node {
```
dkoshkin: Can you add some unit tests, it would be good to test some edge cases and make sure everything works as expected.

emmetthitz: Sure thing.

pkg/install/plan_types.go
emmetthitz: slice r ultimately ends up empty no matter what. preserving its order shouldn't matter. are you thinking of some case in particular where this might be a problem?

dkoshkin: Maybe Im misunderstanding how this is going to work, but I expect when modifying the node count and TF adds a new node the new order of the nodes will remain the same with an additional one at the end.

emmetthitz: Yeah, that shouldn't be affected.

dkoshkin: Great, would be awesome to have a unit test showing that.

dkoshkin:
LGTM!

PR #1055, opened 2018-01-17
Upgrading to CoreDNS v1.0.3 

### LINKED ISSUE #1053: Upgrade to using CoreDNS v1.0.3
Title says it all.

swade1987:
Closes #1053 

I also removed the master toleration, for more info see https://github.com/coredns/deployment/issues/50

We are now also scraping via the pods directly rather than the service.

### DIFF
--- a/ansible/roles/coredns/templates/coredns.yaml
+++ b/ansible/roles/coredns/templates/coredns.yaml
@@ -79,6 +79,9 @@ spec:
       k8s-app: coredns
   template:
     metadata:
+      annotations:
+        prometheus.io/port: "9153"
+        prometheus.io/scrape: "true"
       labels:
         k8s-app: coredns
     spec:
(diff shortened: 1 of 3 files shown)


ansible/roles/coredns/templates/coredns.yaml
```diff
@@ -137,7 +137,7 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
-  name: kube-dns
+  name: coredns
```
willvrny: After some discussion I feel that leaving this as `kube-dns`, whilst slightly misleading is beneficial for two reasons:

- Reduce downtime of DNS during an upgrade when switching between the two DNS options.
- `kube-dns` can act as an interface that abstracts away the underlying implementation.

What do you guys think?

swade1987: The maintainers of core-dns themselves argued about the service naming and went with kube-dns to allow for a smooth upgrade from kube-dns to core-dns.

dkoshkin:
LGTM!

swade1987:
@dkoshkin want me to delete the branch?

PR #1064, opened 2018-01-19
Tiller namespace

dkoshkin:
Some work on #938

Added `namespace` option to allow setting the `--tiller-namespace` option when running `helm init`


swade1987:
@magowant please note that when performing an upgrade and setting this namespace to something different from `kube-system `then the tiller/helm history we be removed as its essentially a new deployment of Tiller.

PR #1067, opened 2018-01-19
Docker storage drivers

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/770

Add support to allow users to configure [docker storage drivers](https://docs.docker.com/engine/userguide/storagedriver/)

The values from the  `opts{}` field will be set in docker `daemon.json` `"storage-opts"` 
```
# Docker daemon configuration of all cluster nodes.
docker:

  # Set to true if docker is already installed and configured.
  disable: false
  logs:
    driver: json-file
    opts:
      max-file: "1"
      max-size: 50m

  storage:

    # Leave empty to have docker automatically select the driver.
    driver: ""
    opts: {}

    # Used for setting up Device Mapper storage driver in direct-lvm mode.
    direct_lvm_block_device:

      # Absolute path to the block device that will be used for direct-lvm mode.
      # This device will be wiped and used exclusively by docker.
      path: ""
      thinpool_percent: "95"
      thinpool_metapercent: "1"
      thinpool_autoextend_threshold: "80"
      thinpool_autoextend_percent: "20"
```

**NOTE IT IS NOT RECOMMENDED CHANGING THE STORAGE DRIVER OF A RUNNING CLUSTER**

---
To maintain backwards `docker.storage.direct_lvm` providing the `docker.storage.direct_lvm` field will still work as it previously has (it is deprecated and will be removed in a future version), during execution it will be updated to the latest schema.

The new schema allows the user to override any of the previously hardcoded options.
Old config:
```
docker:
  disable: false
  storage:
    direct_lvm:
      enabled: true
      block_device: "/dev/xvdb"
      enable_deferred_deletion: false
```
New Equivalent Config
```
  storage:
    driver: devicemapper
    opts:
      dm.thinpooldev: /dev/mapper/docker-thinpool
      dm.use_deferred_deletion: "false"
      dm.use_deferred_removal: "true"
    direct_lvm_block_device:
      path: /dev/xvdb
      thinpool_percent: "95"
      thinpool_metapercent: "1"
      thinpool_autoextend_threshold: "80"
      thinpool_autoextend_percent: "20"
```

### DIFF
--- a/ansible/_docker.yaml
+++ b/ansible/_docker.yaml
@@ -10,7 +10,6 @@
     roles:
       - role: packages-docker
         when: allow_package_installation|bool == true
-      - role: docker-storage
       - role: docker-registry-cert
         when: configure_docker_with_private_registry is defined and configure_docker_with_private_registry|bool == true
       - role: docker-proxy
--- a/ansible/group_vars/all.yaml
+++ b/ansible/group_vars/all.yaml
@@ -55,11 +55,6 @@ docker_certificates_cert_path: "{{ docker_install_dir }}/{{ docker_certificates_
 docker_certificates_key_path: "{{ docker_install_dir }}/{{ docker_certificates_key_file_name }}"
 #===============================================================================
 # docker configuration
-# docker.storage.directlvm.enabled: no default
-# docker.storage.directlvm.block_device: no default
-# docker.storage.directlvm.enable_deferred_deletion: no default
-docker_device_mapper_thin_pool_autoextend_threshold: 80
-docker_device_mapper_thin_pool_autoextend_percent: 20
 docker_system_d: /etc/systemd/system/docker.service.d
 #===============================================================================
 # calico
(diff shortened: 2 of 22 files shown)


swade1987:
@dkoshkin 

Not sure if having the below options is too much for the user.

```
 direct_lvm_block_device:
      path: /dev/xvdb
      thinpool_percent: "95"
      thinpool_metapercent: "1"
      thinpool_autoextend_threshold: "80"
      thinpool_autoextend_percent: "20"
```

Are we going to validate all of these?

dkoshkin:
Yes we can validate thats its a number and its `=>0` but not sure if we can do much more than that.

dkoshkin:
Also if any of the values are missing they are defaulted to the ones shown above.

swade1987:
Are we validating the value of `driver` if set?

dkoshkin:
No `driver` validation, the reasoning for the lack of validation is all these options are pretty critical in the cluster, the user is expected to understand and know the valid config options by reading the Docker documentation.

swade1987:
I am also assuming that the driver will remain mandatory as well?

dkoshkin:
No, an [empty value is valid](https://github.com/apprenda/kismatic/pull/1067/files#diff-f5f00b312fbf6dd1efbb751aa6e1f8b1R474).

PR #1069, opened 2018-01-19
Removing the prometheus scrape annotations from kube-proxy.

### LINKED ISSUE #1068: Remove prometheus scrape annotations from kube-proxy.
Currently the Prometheus target for the kube-proxy responds with the following:

`Get http://<ip>:10249/metrics: dial tcp <ip>:10249: getsockopt: connection refused`

I think it’s due to the fact the metrics bind address defaults to `127.0.0.1` whilst the other bind address defaults to `0.0.0.0`. Unfortunately, I think there is an open issue with kube-proxy as attempting to override via the command line flags fails validation.

When trying to set `- --metrics-bind-address=0.0.0.0:10249` you got the following error in the logs 

```
error: invalid argument "0.0.0.0:10249" for 
--metrics-bind-address=0.0.0.0:10249: "0.0.0.0:10249" is not a valid IP address
```

Therefore we should probably remove the scrape annotation until this has been fixed upstream.

swade1987:
Closes #1068

### DIFF
--- a/ansible/roles/kube-proxy/templates/kube-proxy.yaml
+++ b/ansible/roles/kube-proxy/templates/kube-proxy.yaml
@@ -46,8 +46,6 @@ spec:
         kismatic/version: "{{ kismatic_short_version }}"
         k8s-app: kube-proxy
         scheduler.alpha.kubernetes.io/critical-pod: ''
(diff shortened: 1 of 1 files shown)


dkoshkin:
LGTM

PR #1073, opened 2018-01-22
Remove configuration of Kismatic hem repo

dkoshkin:
The kismatic-charts repo was deprecated and will no longer be configured with the cluster.


swade1987:
LGTM!

Issue #1056, opened 2018-01-17
Provide the ability to add an ingress node.


swade1987:
Suggested command: `kismatic install add-ingress <hostname> <ip address>`

dkoshkin:
Closed in #1076

PR #1076, opened 2018-01-23
Add node roles

dkoshkin:
Closes https://github.com/apprenda/kismatic/issues/1056

Extends the existing `kismatic install add-node` with `--roles` flag.
```
./kismatic install add-node worker001 192.168.42.4 --roles worker,ingress
./kismatic install add-node worker002 192.168.42.5 --roles ingress
./kismatic install add-node worker003 192.168.42.6 --roles storage
./kismatic install add-node worker004 192.168.42.7 --roles worker,ingress,storage
```

When an incorrect role is provided
```
kismatic install add-node worker003 192.168.42.6 --roles foo
invalid role "foo", options [worker ingress storage]
```

This PR modifies an existing command `add-worker` renaming it to `add-node`. **NOTE** `add-worker` is now aliased to `add-node`.


emmetthitz:
Only gripe is the verbs used. ~`./kismatic install add-worker worker001 192.168.42.4 --roles worker,ingress
`~ -> `./kismatic install add-node worker001 192.168.42.4 --roles worker,ingress`?

Otherwise, LGTM.

swade1987:
I would agree with @emmetthitz here it should really be `add-node` now.

dkoshkin:
Renamed command to `add-node`.
@swade1987 we cant just add a new master as that requires restarting all `kube-apiservers`. That features requires more thought and testing and should be added later.

swade1987:
@dkoshkin understood we should probably call that out in the release notes for this feature and add it to a future issue.

dkoshkin:
@swade1987 yep for sure!

swade1987:
may want to change the name of the PR @dkoshkin

PR #1081, opened 2018-01-23
Upgrading GlusterFS to 3.12.5

### LINKED ISSUE #1080: Upgrade to GlusterFS 3.12
We should probably upgrade to the latest version of GlusterFS which seems to be 3.12

See [here](https://www.gluster.org/install/)

swade1987:
Closes #1080 

@vbellur are we using the correct repo URLs here?


swade1987:
@vbellur thanks for taking the time to review it's much appreciated.

dkoshkin:
The issue is that the systemd service `glusterfs-server.service` that we started in the previous version is no longer there, I haven't looked further on what it should be.

PR #1094, opened 2018-01-25
KET v1.8.0

dkoshkin:
PR to merge all KET v1.8.0 changes


emmetthitz:
LGTM!

Can we merge this into KET server too then?

dkoshkin:
Yep my plan is to rebase `ket-server` with the `master` after the release

PR #1095, opened 2018-01-26
Fix seed-registry command to support configurable cluster version

dkoshkin:
Integration tests were failing trying to seed-registry.

The `seed-registry` command needs to read the `container_images.yaml`([link](https://github.com/apprenda/kismatic/blob/seed-registry/ansible/group_vars/container_images.yaml#L15)) file that now contains version overrides,  the command needed modification to use default versions or try to read the versions from the plan file.

### DIFF
--- a/pkg/cli/seed.go
+++ b/pkg/cli/seed.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	yaml "gopkg.in/yaml.v2"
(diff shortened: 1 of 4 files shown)


dkoshkin:
Ignore the failed test, I ran the full test-suite and the reason why the tests failed was because this is an untagged build.
All tests are passing.

pkg/cli/seed.go
```diff
@@ -182,14 +210,58 @@ func seedImage(stdout, stderr io.Writer, img image, registry string, verbose boo
...
+	versions := make(map[string]string, 0)
+	versions["kube_proxy"] = install.KubernetesVersionString
+	versions["kube_controller_manager"] = install.KubernetesVersionString
+	versions["kube_scheduler"] = install.KubernetesVersionString
+	versions["kube_apiserver"] = install.KubernetesVersionString
+
+	return versions
+}
+
+func planVersions(versions map[string]string, plan install.Plan) map[string]string {
```
emmetthitz: I _think_ this should be a method of the `install.Plan` type, and should take no arguments. I'm not sure I understand why you'd even pass in `versions` in the first place.

pkg/cli/seed.go
```diff
@@ -182,14 +210,58 @@ func seedImage(stdout, stderr io.Writer, img image, registry string, verbose boo
...
+	if plan.Cluster.Version != "" {
```
emmetthitz: Likewise, I think we'd want to return `defaultVersions()` in the case where this isn't true.

pkg/cli/seed.go
dkoshkin: Thanks! That worked out much better.

Issue #1099, opened 2018-02-07
Support docker storage opts with numbers 


dkoshkin:
Thanks to @robb-foster-ck for reporting this.

**Is this a BUG REPORT or FEATURE REQUEST?**:
BUG REPORT

**What happened**:
When passing in:
```
docker:
  storage:
    opts:
      "overlay2.override_kernel_check": true
```
The installation failed with:
```
error installing: error recording plan file to runs/apply/2018-02-07-10-25-43/kismatic-cluster.yaml: Empty Stack
```

**What you expected to happen**:
The installation to succeed as this is a valid docker option and a valid `yaml` key.

**How to reproduce it (as minimally and precisely as possible)**:
Set the plan file described above ^ or any other `yaml` key with a number in it.

**Anything else we need to know?**:
This happens when the installer tries to parse the plan file because the regex for keys used did not include numbers `[^a-zA-Z]*([a-z_\-A-Z.\d]+)[ ]*:` 

**Environment**:
- Kismatic version (use `kismatic version`):
`v1.8.0`

PR #1100, opened 2018-02-07
Fix regex to support yaml keys with numbers

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/1099


robb-foster-ck:
Thanks.

PR #1102, opened 2018-02-08
Swagger docs

emmetthitz:
Annotations done. Would be good to get eyes on it to see if I missed something.

- [x] Fix integration test endpoints
- [x] Clean up some messes
- [x] Figure out why spec generation doesn't work
- [x] Embed swagger-ui

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,8 +107,12 @@ bare-test: vendor
 integration-test: dist just-integration-test
 
 .PHONY: vendor
-vendor: tools/glide-$(GLIDE_GOOS)-$(HOST_GOARCH)
+vendor: tools/glide-$(GLIDE_GOOS)-$(HOST_GOARCH) 
 	tools/glide-$(GLIDE_GOOS)-$(HOST_GOARCH) install
+
+
+tools/swagger-$(GLIDE_GOOS)-$(HOST_GOARCH):
+	curl -L https://github.com/go-swagger/go-swagger/releases/download/$(SWAGGER_VERSION)/swagger_$(GLIDE_GOOS)_$(HOST_GOARCH)
 
 tools/glide-$(GLIDE_GOOS)-$(HOST_GOARCH):
 	mkdir -p tools
@@ -206,6 +210,10 @@ docs/update-plan-file-reference.md:
 
 docs/generate-plan-file-reference.md:
 	@go run cmd/gen-kismatic-ref-docs/*.go -o markdown pkg/install/plan_types.go Plan
+
+docs/swagger: tools/swagger_$(GLIDE_GOOS)_$(HOST_GOARCH)
+	swagger generate spec -o ./swagger.json
+	
 
 version: FORCE
 	@echo VERSION=$(VERSION)
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ HOST_GOARCH = $(shell go env GOARCH)
 GLIDE_VERSION = v0.13.1
 ANSIBLE_VERSION = 2.3.0.0
 TERRAFORM_VERSION = 0.11.0
+SWAGGER_VERSION = 0.13.0
+SWAGGER-UI_VERSION = 3.10.0
 KUBERANG_VERSION = v1.2.2
 GO_VERSION = 1.9.2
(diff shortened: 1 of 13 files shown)


pkg/server/http/handler/clusters.go
```diff
@@ -211,6 +269,7 @@ func (api Clusters) GetAll(w http.ResponseWriter, r *http.Request, p httprouter.
 		return
 	}
 	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
```
emmetthitz: Woops

Makefile
```diff
@@ -17,7 +17,7 @@ GLIDE_VERSION = v0.13.1
 ANSIBLE_VERSION = 2.3.0.0
 TERRAFORM_VERSION = 0.11.0
 KUBERANG_VERSION = v1.2.2
-GO_VERSION = 1.9.2
+GO_VERSION = 1.9.3
```
dkoshkin: Is this required, it might be better to have this in a different PR. Also there was a security patch release with Go 1.9.4.

circle.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - checkout
       - run: 
           name: Install unzip
-          command: apt update && apt install unzip
+          command: sudo apt update && sudo apt install unzip
```
dkoshkin: Was this not working before?

pkg/server/http/handler/clusters.go
```diff
@@ -32,35 +32,93 @@ type Clusters struct {
...
+	//
+	// in: body
+	// required: true
+	Provisioner store.Provisioner `json:"provisioner"`
 }
 
 // The Provisioner defines the infrastructure provisioner that should be used
 // for hosting the cluster
 type sanitizedProvisioner struct {
+	// Options: aws, azure, digital ocean
```
dkoshkin: + any other provider user defined.

pkg/server/http/server.go
```diff
@@ -1,3 +1,21 @@
...
+//     BasePath: /api/v1
+//     Version: 0.0.1
+//     License: Apache 2.0 http://www.apache.org/licenses/
+//
+//     Consumes:
+//     - application/json
+//
+//     Produces:
+//     - application/json
+//     - application/x-yaml
```
dkoshkin: I still don't know how I feel about using this instead of `text/plain`

circle.yml
emmetthitz: It's related to the previous comment. The updated container doesn't give root by default.

Makefile
emmetthitz: Yeah, there's a handful of things that got tracked with this that I added by accident.

pkg/server/http/server.go
emmetthitz: The more I think about it, the more I think it's kind of misleading. I think plain text generally implies that there's no structure to it. I.E., a log file or some other kind of typical human writing.

integration-tests/daemon_test.go
```diff
@@ -95,7 +96,7 @@ The error: %v
 				pb, err := json.Marshal(payload)
 				Expect(err).ToNot(HaveOccurred())
 
-				clustersEndpoint := fmt.Sprintf("https://localhost:%d/clusters", daemonPort)
+				clustersEndpoint := fmt.Sprintf("https://localhost:%d%s/clusters", daemonPort, ketHttp.APIPath)
```
dkoshkin: Looks like missing a slash between `%d%s`

emmetthitz: APIPath begins with a `/`. I'll change that as well if you think it makes more sense stylistically.

dkoshkin: Its inconsistent with [this](https://github.com/apprenda/kismatic/pull/1102/files#diff-334a3bd3e1b27ff451cbce986a0b9623R144) so one of them will be wrong, I do think its should have `/`

emmetthitz: `httprouter` complains if `APIPath` doesn't begin with a `/`. I corrected the other version instead of changing this.

Issue #1116, opened 2018-02-14
Support metrics-server


benmathews:
FEATURE REQUEST

New versions (1.7+) of kubernetes provide metrics through a metrics server.
https://github.com/kubernetes-incubator/metrics-server
https://kubernetes.io/docs/tasks/debug-application-cluster/core-metrics-pipeline/

Please support automatic installation

dkoshkin:
[This is unfortunate](https://github.com/kubernetes/heapster/issues/1840). Those v1.8 release notes were not clear at all.

Thanks for bringing this up. We will work on adding this as it looks like HPA does not work without the metrics-server.

dkoshkin:
Ok looks like the HPA does in fact NOT function properly without having the metrics-server (although `kubectl top` does work with just `heapster`)
```
I0215 17:44:02.278877       1 event.go:218] Event(v1.ObjectReference{Kind:"HorizontalPodAutoscaler", Namespace:"default", Name:"nginx-deployment", UID:"87b68664-1277-11e8-b868-080027c30a85", APIVersion:"autoscaling/v2beta1", ResourceVersion:"5281", FieldPath:""}): type: 'Warning' reason: 'FailedComputeMetricsReplicas' failed to get cpu utilization: unable to get metrics for resource cpu: unable to fetch metrics from API: the server could not find the requested resource (get pods.metrics.k8s.io)
```

Trying to deploy the metrics-server results in an error:
```
F0215 17:46:20.335656       1 heapster.go:97] Could not create the API server: cluster doesn't provide requestheader-client-ca-file
```

Relevant issue https://github.com/Azure/acs-engine/issues/1390

PR #1120, opened 2018-02-20
Add metrics-server addon

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/1116

Add `metrics-server` addon.
```
add_ons:
  # Metrics Server is a cluster-wide aggregator of resource usage data.
  metrics_server:
    disable: false
```
For this to work properly, new api-server options 

For the deployment to work properly new `kube-apiserver` options are needed:
```
+  "requestheader-client-ca-file": "{{ kubernetes_certificates.ca }}"
+  "proxy-client-cert-file": "{{ kubernetes_certificates.proxy_client }}"
+  "proxy-client-key-file": "{{ kubernetes_certificates.proxy_client_key }}"
+  "requestheader-allowed-names": ""
+  "requestheader-extra-headers-prefix": "X-Remote-Extra-"
+  "requestheader-group-headers": "X-Remote-Group"
+  "requestheader-username-headers": "X-Remote-User"
```


dkoshkin:
Looking at docs and issues around this we need a new CA instead of using the root CA.

emmetthitz:
Ah right. Totally forgot.

pkg/install/pki.go
```diff
@@ -102,7 +108,7 @@ func (s certificateSpec) equal(other certificateSpec) bool {
 }
 
 // returns a list of specs for all the certs that are required for the node
-func certManifestForNode(plan Plan, node Node) ([]certificateSpec, error) {
+func certManifestForNode(plan Plan, node Node, ca *tls.CA) ([]certificateSpec, error) {
```
emmetthitz: I think it would make sense for this to be a method of `Node`.

pkg/install/pki.go
```diff
@@ -178,20 +190,21 @@ func certManifestForNode(plan Plan, node Node) ([]certificateSpec, error) {
...
+			ca:          ca,
 		})
 	}
 
 	return m, nil
 }
 
 // returns a list of cert specs for the cluster described in the plan file
-func certManifestForCluster(plan Plan) ([]certificateSpec, error) {
+func certManifestForCluster(plan Plan, clusterCA *tls.CA, proxyClientCA *tls.CA) ([]certificateSpec, error) {
```
emmetthitz: And for this to be a method of `Plan`

pkg/install/pki_test.go
```diff
@@ -529,7 +552,7 @@ func TestGenerateClusterCertificatesPlanFileExpirationIsRespected(t *testing.T)
 	}
 }
 
-func validateClientCertificateAndKey(certsDir, filename, expectedCommonName string, expectedOrganizations ...string) func(t *testing.T) {
+func validateClientCertificateAndKey(certsDir, filename, expectedIssuer string, expectedCommonName string, expectedOrganizations ...string) func(t *testing.T) {
```
emmetthitz: This feels a little clunky too, but I'm not sure what the answer is.

pkg/install/pki.go
dkoshkin: IMO it makes more sense to keep all the PKI/certs related code in its own folder as a function since these operations are not simple retrieval methods that return some fields from the node and plan but instead return a pki specific struct.

emmetthitz: I think that's still possible though, right? This is even still a part of the same package - so I can't find any gripes wrt that. It just kind of smells that you have to append `ForX` when that's the point of go's receivers I thought.

emmetthitz:
LGTM

PR #1126, opened 2018-02-21
Dockerize kismatic

emmetthitz:


dkoshkin:
What's happening with this PR? Also we need to merge the Makefile changes(and all other upstream changes) into `ket-server` first and then finish this up.

emmetthitz:
This PR includes the makefile changes - including those to publish to dockerhub and build a docker-compatible package.

pkg/cli/dashboard.go
```diff
@@ -72,6 +73,13 @@ func doDashboard(in io.Reader, out io.Writer, opts *dashboardOpts) error {
 	if generateErr == nil {
 		fmt.Fprintf(out, "Use the kubeconfig in %q\n", adminKubeconfig)
 	}
+	// Dashboard is accessible.. take action
+	// In URL mode or
+	// Running in a docker container, can't open a browser from here
+	if opts.dashboardURLMode || util.RunningInDocker() {
```
dkoshkin: We need another flag on this command, maybe `--no-ui` that will still start the proxy but not try to open the browser.

integration-tests/kismatic_platform_suite_test.go
```diff
@@ -145,7 +150,7 @@ func extractTarball(src, dst string) error {
 // extracts the current build of kismatic (the one being tested)
 func extractCurrentKismatic(dest string) error {
 	By(fmt.Sprintf("Extracting current kismatic to directory %q", dest))
-	if err := extractTarball(filepath.Join(currentKismaticDir, "kismatic.tar.gz"), dest); err != nil {
```
dkoshkin: CTRL+SHIFT+F `kismatic.tar.gz`, there are other places where this needs to change.

pkg/cli/dashboard.go
```diff
@@ -68,19 +69,27 @@ func doDashboard(in io.Reader, out io.Writer, opts *dashboardOpts) error {
 		}
 	}
 
+	cmd := exec.Command("./kubectl", "proxy", "--kubeconfig", kubeconfig)
```
dkoshkin: This only needs to happen when using the new `--no-ui` flag.

Makefile
```diff
@@ -406,6 +418,6 @@ trigger-ci-slow-tests:
 	  $(CIRCLE_ENDPOINT)
 trigger-ci-focused-tests:
 	@echo Triggering focused test
-	curl -u $(CIRCLE_CI_TOKEN): -X POST --header "Content-Type: application/json"    \
-	  -d "{\"build_parameters\": {\"FOCUS\": \"$(FOCUS)\"}}"                         \
-	  $(CIRCLE_ENDPOINT)
\ No newline at end of file
+	curl -u $(CIRCLE_CI_TOKEN): -X POST --header "Content-Type: application/json"     \
```
emmetthitz: whoops. will fix.

Makefile
```diff
@@ -406,6 +418,6 @@ trigger-ci-slow-tests:
 	  $(CIRCLE_ENDPOINT)
 trigger-ci-focused-tests:
 	@echo Triggering focused test
-	curl -u $(CIRCLE_CI_TOKEN): -X POST --header "Content-Type: application/json"    \
```
dkoshkin: Remove these changes.

pkg/cli/dashboard.go
emmetthitz: This should be fixed.

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:2
```
dkoshkin: As part of this PR we should add docs how to run the docker container, probably just in the main README

pkg/cli/dashboard.go
emmetthitz: Opted to just run the command if this check passes.

dkoshkin: There is still no way to start the proxy without opening the browser. The `util.RunningInDocker()` should move to above `if err := browser.OpenURL(url); err != nil {`

integration-tests/disconnected_install.go
```diff
@@ -62,7 +63,7 @@ func seedRegistry(repoNode NodeDeets, registryCAFile string, registryPort int, s
 	cmds = []string{
 		fmt.Sprintf("sudo docker login -u kismaticuser -p kismaticpassword %s", registry),
 		"sudo mkdir kismatic",
-		"sudo tar -xf /tmp/kismatic-linux.tar.gz -C kismatic",
+		"sudo tar -xf /tmp/kismatic-" + runtime.GOOS + ".tar.gz -C kismatic",
```
dkoshkin: Can you make these `runtime.GOOS` changes be its own commit so it can be cherrypicked into master

Makefile
```diff
@@ -243,8 +244,8 @@ copy-inspector:
 	cp -r bin/inspector/* $(BUILD_OUTPUT)/ansible/playbooks/inspector
 
 copy-playbooks:
-	mkdir -p $(BUILD_OUTPUT)/ansible
 	rm -rf $(filter-out $(BUILD_OUTPUT)/ansible/playbooks/inspector $(BUILD_OUTPUT)/ansible/playbooks/kuberang, $(wildcard $(BUILD_OUTPUT)/ansible/playbooks/*))
+	mkdir -p $(BUILD_OUTPUT)/ansible
```
dkoshkin: This snuck in again.

emmetthitz: Why isn't there a screeching emoji

integration-tests/disconnected_install.go
emmetthitz: Done

Dockerfile
emmetthitz: Done

pkg/cli/dashboard.go
```diff
@@ -38,6 +40,7 @@ func NewCmdDashboard(in io.Reader, out io.Writer) *cobra.Command {
 
 	cmd.Flags().StringVar(&opts.generatedAssetsDir, "generated-assets-dir", "generated", "path to the directory where assets generated during the installation process will be stored")
 	cmd.Flags().BoolVar(&opts.dashboardURLMode, "url", false, "Display the kubernetes dashboard URL instead of opening it in the default browser")
+	cmd.Flags().BoolVar(&opts.noUI, "no-ui", false, "Disables the kube proxy that allows dashboard use - proxy runs on 0.0.0.0:8001.")
```
dkoshkin: This flag doesn't make sense, when the command is run with this flag nothing happens.

The current behavior in master is running `kismatic dashboard`:
1. execute `kubectl proxy` with the default settings listening on 127.0.0.1:8001
2. open the browser on the users machine

This will not work when running in a docker container as there is no browser and the command will fail. To fix this I proposed slightly modifying this behavior, running `kismatic dashboard` would:
1. execute `kubectl proxy` with `--address=0.0.0.0` flag when `util.RunningInDocker()` returns `true`
2. Wrap the [lines](https://github.com/apprenda/kismatic/blob/e25de525a682ff1cbe49369434dfefa282707818/pkg/cli/dashboard.go#L95) with:
```
if !util.RunningInDocker() {
	if err := browser.OpenURL(url); err != nil {
		fmt.Fprintf(out, "Unexpected error opening the kubernetes dashboard: %v. You may access it at %q", err, url)
	}
}
```
3. Before even starting the proxy print the url where the user can reach the dashboard.

dkoshkin: No need for the new flag

dkoshkin:
Please try to keep the commit names meaningful,  `Requested dashboard changes` won't mean anything when looking back at the history, `Fix running kismatic dashboard command inside a docker container by not opening the browser` is much more representative of what is actually going on

emmetthitz:
I figured it would end up in a squash anyway. The commit history is only really going to reflect this PR, unless you dig into the commit details.

pkg/cli/dashboard.go
```diff
@@ -72,11 +73,17 @@ func doDashboard(in io.Reader, out io.Writer, opts *dashboardOpts) error {
 	if generateErr == nil {
 		fmt.Fprintf(out, "Use the kubeconfig in %q\n", adminKubeconfig)
 	}
-	if err := browser.OpenURL(url); err != nil {
-		fmt.Fprintf(out, "Unexpected error opening the kubernetes dashboard: %v. You may access it at %q", err, url)
+	// Dashboard is accessible.. take action
+	// In URL mode or
+	// Running in a docker container, can't open a browser from here
+	if opts.dashboardURLMode || util.RunningInDocker() {
```
dkoshkin: Let's not check `opts.dashboardURLMode` anymore since that already happens in the beginning of the function.

Needs to only print [this line](https://github.com/apprenda/kismatic/blob/5259cbea9d2bfa3754784617206a8b3404ff3595/pkg/cli/dashboard.go#L72) when opening the dashboard, can just move in the `else` block.

pkg/cli/dashboard.go
```diff
@@ -72,11 +73,17 @@ func doDashboard(in io.Reader, out io.Writer, opts *dashboardOpts) error {
...
+		fmt.Fprintln(out, url)
```
dkoshkin: Can we print a message something like `navigate to %s in your browser to access the dashboard`

pkg/cli/dashboard.go
```diff
@@ -72,11 +73,17 @@ func doDashboard(in io.Reader, out io.Writer, opts *dashboardOpts) error {
...
+	} else {
+		if err := browser.OpenURL(url); err != nil {
+			fmt.Fprintf(out, "Unexpected error opening the kubernetes dashboard: %v. You may access it at %q", err, url)
+		}
 	}
-
-	cmd := exec.Command("./kubectl", "proxy", "--kubeconfig", kubeconfig)
+	cmd := exec.Command("./kubectl", "proxy", "--kubeconfig", kubeconfig, "--address", "0.0.0.0")
```
dkoshkin: The `"--address", "0.0.0.0"` should only be set when running in a docker container, this is a security concern.

Dockerfile
dkoshkin: I just looked at the image size and its `400MBs` does the `2-alpine` base not work?

Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+
+WORKDIR /root/kismatic
+EXPOSE 8080 8443
```
dkoshkin: Need to add `8001`

Dockerfile
emmetthitz: It could, potentially. I'll add it as an issue. `alpine` doesn't ship with a couple of the system binaries we implicitly expect (`sed`, `ssh`). I can address this in a future PR.

README.md
```diff
@@ -82,15 +82,28 @@ The installation consists of three phases:
...
    1. The installer checks your provisioned infrastructure against your intent.
    2. If the installation plan is valid, Kismatic will build you a cluster.
    3. After installation, Kismatic performs a basic test of scaling and networking on the cluster
 
+## Using Kismatic with Docker.
+
+### Using the CLI
+
+The basic installation instructions still apply. The entrypoint to the container is the binary itself. Please note that the container itself does not have persistent storage, so you need a volume mount to preserve your plan file, ssh keys, terraform state, etc.
+`docker run -itv $(pwd):/root/kismatic/ apprenda/kismatic`
```
dkoshkin: Doesn't work with neither an existing cluster nor a new cluster.
```
➜  kettest ll
total 16
drwxr-xr-x  4 dkoshkin  1535981906   128B Mar  9 15:07 generated
-rw-r--r--  1 dkoshkin  wheel        6.8K Mar  9 15:16 kismatic-cluster.yaml
➜  kettest docker run -itv $(pwd):/root/kismatic/ apprenda/kismatic:dev-dockerize
docker: Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"/root/kismatic/kismatic\": stat /root/kismatic/kismatic: no such file or directory": unknown.
➜  kettest2 ll
total 0
drwxr-xr-x   2 dkoshkin  wheel    64B Mar  9 15:18 .
drwxrwxrwt  15 root      wheel   480B Mar  9 15:19 ..
➜  kettest2 docker run -itv $(pwd):/root/kismatic/ apprenda/kismatic:dev-dockerize
docker: Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"/root/kismatic/kismatic\": stat /root/kismatic/kismatic: no such file or directory": unknown.
```

pkg/cli/dashboard.go
emmetthitz: Should that condition in the beginning also drop the return?

dkoshkin: No because in `--url` mode all the user wants is for the url string and to exit.

emmetthitz: Sure. But that url won't point to anything then, right?

emmetthitz: Feels bad from UX perspective

dkoshkin: It will point to the full URL after a user runs `kubectl proxy`, its just a simple way to get the url instead of having to dig through the k8s docs to know that the dashboard can be reached at:
```
http://localhost:8001/api/v1/namespaces/kube-system/services/https:kubernetes-dashboard:/proxy/#!/login
```

This is also the same behavior as we have know, if we want to change it it needs to be its own PR not part of this.

emmetthitz:
The scope of this PR has been cut down drastically and relevant sections moved to other PRs.

Makefile
```diff
@@ -357,6 +359,16 @@ dist-common: vendor build-host build-inspector-host copy-all
...
 dist-host: shallow-clean dist-common
 
+dist-docker: 
+	@$(MAKE) GOOS=linux BUILD_OUTPUT=out-docker dist-common
+	docker build . --no-cache --tag apprenda/kismatic:dev-$(BRANCH) 
+	docker push apprenda/kismatic:dev-$(BRANCH)
+
+docker-release:
+	@$(MAKE) GOOS=linux BUILD_OUTPUT=out-docker dist-common
+	docker build . --no-cache --tag apprenda/kismatic
```
dkoshkin: Here we should release both `apprenda/kismatic:latest` and `apprenda/kismatic:v2.0.0`

README.md
```diff
@@ -82,15 +82,32 @@ The installation consists of three phases:
...
 
+## Using Kismatic with Docker.
+
+### Using the CLI
+
+The basic installation instructions still apply. The entrypoint to the container is the binary itself. Please note that the container itself does not have persistent storage, so you need a volume mount to preserve your plan file, ssh keys, terraform state, etc.
+`docker run -itv $(pwd):/root/kismatic/clusters apprenda/kismatic`
+If you're looking to provide your own terraform config, you need to add a volume for that as well.
+I.E.:
+`docker run -itv $(pwd):/root/kismatic/clusters -v path_to_dir_with_HCL:/root/kismatic/providers/your_provider_name_here apprenda/kismatic`
```
dkoshkin: What happens to existing providers when you use a volume, I would assume they still exist and available to the tool, but it would be good to check.

Makefile
emmetthitz: Untagged builds (IE anything missing a `:`) would be automatically pushed as `:latest`. I just created this as a macro for `:latest`. We can add a `:2.0.0` then when we actually release.

README.md
emmetthitz: If you mount it at the root, IE `/root/kismatic/providers/`, you overwrite any existing providers. If you have your own single or multiple HCL scripts, you'd have to mount them where the document says.

Makefile
dkoshkin: Ok thats fine.

README.md
dkoshkin: Ok, just wanted to confirm when mounting `/root/kismatic/providers/your_provider_name_here` that existing providers still work.

dkoshkin:
LGTM!

PR #1133, opened 2018-02-25
Kube bench

dkoshkin:
Fix some of the [kube-bench](https://github.com/aquasecurity/kube-bench) failures.


willvrny:
With the change 'Set cAdvisor port to 0' just need to make users aware that if they're not going through the API server for their cadvisor monitoring then they will need to. Or use the flags override in the plan file?

dkoshkin:
Yes @willvrny, my plan is to highlight all of these changes in the release notes.

ansible/group_vars/all.yaml
```diff
@@ -32,14 +32,14 @@ etcd_certificates:
   etcd_client_key: "{{ etcd_install_dir }}/etcd-client-key.pem"
   owner: root
   group: root
-  mode: 0660
+  mode: "0660"
```
emmetthitz: Does this actually fix it? 
Isn't this the point of duck typing?

dkoshkin: Yeah I was surprised to, but I noticed for etcd certs were being set to "0432" and was fixed by wrapping these in quotes.

PR #1146, opened 2018-02-28
fix merge error not providing correct plans

emmetthitz:
+being able to locally define vars instead of using the ones in the request


dkoshkin:
LGTM!

emmetthitz:
There's a mutation test that's still failing and I just want to double check and make sure it's not something super dumb.

Issue #803, opened 2017-09-22
Ability to apply Kismatic to a single node


swade1987:
**FEATURE REQUEST**:

I would like to be able to apply Kismatic to a given node.

For example `kismatic install apply -f kismatic-cluster.yaml -n <node>`

jaycoon:
Make that "given nodes" (default being "all nodes")

swade1987:
That'd be `-n <node-a> -n <node-b>`

swade1987:
This comes from a client request who wanted to make changes to the two ingress nodes and not wait for the whole cluster to be re-provisioned again.

emmetthitz:
I can look into this. It should be reasonably easy, I think. It's on the back burner for now though while we work on the KET daemon.

dkoshkin:
Removing from the `v1.8.0` milestone as it needs more discussion.

PR #1158, opened 2018-03-13
CLI --limit nodes

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/803

Add an option to `--limit` execution to a subset of nodes for:
```
kismatic install validate
kismatic install apply
kismatic install step
```

**NOTE** certain tasks will still run on other nodes when they are `delegate_to` in Ansible. The smoketest after an install is also unmodified and will run from the master even when a master is not provided in the `--limit` option.


dkoshkin:
The motivation here is not to add new nodes this way, more of a debugging thing. It is helpful to be able to try a different configurations or bring a single node back to a known state.

emmetthitz:
I see. 

I think that would still be related, though, right? 

You could potentially even really crank the differential installs up depending on the depth of the state we decide to track to. IE you could --limit somenode someothernode yetanothernode --step-to _docker.yaml and potentially just diff the current step those nodes were at with --step-to, and only apply those steps found in the diff to those nodes. This would be a pretty enormous undertaking though.

dkoshkin:
We aren't keeping any state at this point, this PR only addresses the current behavior of the tool. 
Ansible already gives us idempotency, the need for this PR arises when there are manual changes done to the cluster that the tool is not aware of. To limit the impact of running `install` or `step` commands on the cluster I've seen users try to modify the plan file to exclude certain nodes, however that is not always possible with the validation the tool performs, having a `--limit` option optimized for this exact usecase is a much better option.

emmetthitz:
LGTM then!

swade1987:
Is this limit on a per node or per role basis?

dkoshkin:
Per node, its a list so multiple nodes can be specified.

swade1987:
I guess comma separated?

dkoshkin:
Exactly.

PR #1166, opened 2018-03-23
added flexvolume support by specifiying mount path

### LINKED ISSUE #1157: Flexvolumes not working (missing mount path)
**BUG REPORT**:

**What happened**:
When using kubernetes flex volume plugins (ex. Rook, DigitalOcean Block Storage) they fail to be mounted by the kubelet because the flexvolume path is not set and mounted.

Flexvolumes should just work as they are a GA feature as of Kubernetes 1.8 (https://github.com/kubernetes/community/blob/master/contributors/devel/flexvolume.md#flexvolume). The required configuration does not seem to have any negative effects and could easily be made optional/opt-in or could be disabled with a flag.

- Setup a k8s cluster with any provider (ingress and storage are not required)
- Deploy a flexvolume plugin (for example rook)
- Create a Volume
- Attempt to create a pod which uses the volume
- Mount fails, `kubectl describe pod xxxxxx` returns mount failed/timeout.

**What has to be done**:
1) Add a `volume-plugin-dir` argument to `kubelet` for example via `ansible/playbooks/group_vars/all.yaml`:
```yaml
...
kubelet_defaults:
  ...
  "volume-plugin-dir": "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/"
...
```

2) Add a `hostPath` mount to `kube-controller-manager` in `ansible/playbooks/roles/kube-controller-manager/templates/kube-controller-manager.yaml`:
```yaml
...
volumeMounts:
    ...
    - mountPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec/
      name: flexvolume-mount
    ...
volumes:
    ...
    - hostPath:
          path: /usr/libexec/kubernetes/kubelet-plugins/volume/exec/
      name: flexvolume-mount
    ...
```

**Environment**:
- Kismatic version (use `kismatic version`): 1.9.0
- Plan file Gist (remove sensitive information if any): n/a 
- Cloud provider or hardware configuration: irrelevant
- OS (e.g. from /etc/os-release): debian, but should apply to all
- Kernel (e.g. `uname -a`): irrelevant

davidhiendl:
fixes #1157, all that was missing was a volume mount path and an argument for kubelet

### DIFF
--- a/ansible/group_vars/all.yaml
+++ b/ansible/group_vars/all.yaml
@@ -242,6 +242,7 @@ kubelet_defaults:
   "tls-cert-file": "{{ kubernetes_certificates.kubelet }}"
   "tls-private-key-file": "{{ kubernetes_certificates.kubelet_key }}"
   "v": "2"
+  "volume-plugin-dir": "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/"
 
 # etcd IPs
 etcd_networking_cluster_ip_list: "{% for host in groups['etcd'] %}https://{{ host }}:{{ etcd_networking_client_port }}{% if not loop.last %},{% endif %}{% endfor %}"
--- a/ansible/roles/kube-controller-manager/templates/kube-controller-manager.yaml
+++ b/ansible/roles/kube-controller-manager/templates/kube-controller-manager.yaml
@@ -46,6 +46,8 @@ spec:
     - name: usr-ca-certs-host
       mountPath: /usr/share/ca-certificates
       readOnly: true
+    - mountPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec/
+      name: flexvolume-mount
 {% if cloud_provider is defined and cloud_provider == 'aws' and ansible_os_family == 'RedHat' %}
     - mountPath: /etc/ssl/certs/ca-bundle.crt
       name: rhel-ca-bundle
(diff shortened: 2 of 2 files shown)


davidhiendl:
@dkoshkin Sorry for the delay. As I understand it (and I'm no expert on Kubernetes internals, so I might be wrong) this option controls if the "volume manager" (which is an "addon" you install on a cluster to provide/manage a new storage type) handles the attach/detatch operations instead of the kubelet itself.

See this for a description of the option in GO source: https://github.com/kubernetes/kubernetes/blob/ee4d90aaa61150139cdcd67a73e22da8cb226dc6/cmd/kubelet/app/options/options.go#L543
`Enables the Attach/Detach controller to manage attachment/detachment of volumes scheduled to this node, and disables kubelet from executing any attach/detach operations`

And this kubernetes changelog, which basically states it's there for backward compatibility and since you don't need to preserve that as the feature was missing entirely until now I think it should be safe to skip:
https://github.com/kubernetes/kubernetes/blob/ee3bf5bef7235f4b763e5189e6848d9a49a62daf/CHANGELOG-1.3.md#L617
`In order to remain backwards compatible, the new controller only manages attach/detach of volumes that are scheduled to nodes that opt-in to controller management. Nodes running v1.3 or higher of Kubernetes opt-in to controller management by default by setting the "volumes.kubernetes.io/controller-managed-attach-detach" annotation on the Node object on startup. This behavior is gated by a new kubelet flag, "enable-controller-attach-detach,” (default true).`

It also appears that the "old way" of kubelet doing this is might be deprecated or become so:
https://github.com/kubernetes/kubernetes/issues/55517

PR #1171, opened 2018-03-27
Add weave option password, used for traffic encryption

### LINKED ISSUE #1155: Support Weave Encryption
**FEATURE REQUEST**:

Currently Weave encryption is not officially supported. It can easily be enabled by setting a env variable for the weave containers or supply the password to it's configuration file.

This would allow running Kismatic in potentially hostile environments where a private network is not available and traffic may be read by other clients (ex.: various cloud providers).

Excerpt from `ansible/playbooks/roles/weave/templates/weave.yaml` line 115+:
```yaml
           env:
                - name: HOSTNAME
                  valueFrom:
                    fieldRef:
                      apiVersion: v1
                      fieldPath: spec.nodeName
                - name: IPALLOC_RANGE
                  value: "{{ kubernetes_pods_cidr }}"
                - name: WEAVE_PASSWORD
                  value: "{{ variable passed in from configuration ins kismatic-cluster.yaml }}"
```

**Environment**:
- Kismatic version: 1.9.0

dkoshkin:
Fixes #1155

### DIFF
--- a/pkg/install/test/plan-template-with-storage.golden.yaml
+++ b/pkg/install/test/plan-template-with-storage.golden.yaml
@@ -165,6 +165,11 @@ add_ons:
         # Used to detect the IPv4 address of the host.
         ip_autodetection_method: first-found
 
+      weave:
+
+        # Used by Weave for network traffic encryption. Should be reasonably strong, with at least 50 bits of entropy.
+        password: ""
+
   dns:
     disable: false
 
--- a/ansible/roles/weave/templates/weave.yaml
+++ b/ansible/roles/weave/templates/weave.yaml
@@ -120,6 +120,8 @@ items:
                       fieldPath: spec.nodeName
                 - name: IPALLOC_RANGE
                   value: "{{ kubernetes_pods_cidr }}"
+                - name: WEAVE_PASSWORD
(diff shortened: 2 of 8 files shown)


pkg/install/test/plan-template-with-storage.golden.yaml
```diff
@@ -165,6 +165,11 @@ add_ons:
         # Used to detect the IPv4 address of the host.
         ip_autodetection_method: first-found
 
+      weave:
+
+        # Used by Weave for network traffic encryption. Should be reasonably strong, with at least 50 bits of entropy.
```
emmetthitz: This feels a little strange. Would it be a bad idea to generate a password for the user automatically if they select weave as their CNI?

dkoshkin: The encryption adds overhead to the cluster and in many situations its really not needed, I dont think we can assume that the default case is to have the encryption on.

emmetthitz: I'm not sure about that. I think the enterprise use case cares very much about having things as locked down as possible. Though if you're adamant about not defaulting to using encryption, this LGTM.

PR #1175, opened 2018-03-28
Added clusters cmd

emmetthitz:
Following the workflow for the `/clusters` endpoint

### DIFF
--- a/pkg/cli/clusters.go
+++ b/pkg/cli/clusters.go
@@ -1,0 +1,55 @@
+package cli
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"strconv"
+	"text/tabwriter"
+
+	"github.com/spf13/cobra"
+)
+
+type clustersOpts struct {
+	verbose bool
+}
+
+// NewCmdClusters creates a new cluster command
+func NewCmdClusters(out io.Writer) *cobra.Command {
+	opts := &clustersOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "clusters",
+		Short: "list the names of the clusters currently being managed",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return doClusters(out, *opts)
+		},
+	}
+	cmd.Flags().BoolVarP(&opts.verbose, "verbose", "v", false, "print the verbose details of the clusters dir.")
+	return cmd
+}
+
+func doClusters(out io.Writer, opts clustersOpts) error {
+	clusters, err := ioutil.ReadDir("clusters")
+	if err != nil {
+		return err
+	}
+	header := "Cluster Name:\t"
+	if opts.verbose {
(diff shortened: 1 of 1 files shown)


pkg/cli/ls.go
```diff
@@ -0,0 +1,55 @@
...
+type clustersOpts struct {
+	verbose bool
+}
+
+// NewCmdLs creates a new ls command
+func NewCmdLs(out io.Writer) *cobra.Command {
+	opts := &clustersOpts{}
+
+	cmd := &cobra.Command{
+		Use:   "ls",
```
dkoshkin: I though we agreed on `list` can you modify the command to be `list` and then have an alias of `ls`

circle.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - run:
           name: Verify reference API documentation is up to date
           command: 
-            "diff -u <(cat docs/swagger.json) <(GOROOT=$(go env GOROOT) tools/swagger_$(go env GOOS) generate spec -b ./pkg/server)"
+            "diff -u <(cat docs/swagger.json) <(GOROOT=$(go env GOROOT) tools/swagger-$(go env GOOS)-$(go env GOARCH) generate spec -b ./pkg/server)"
```
dkoshkin: How are all other tests not failing when using the wrong path?

pkg/cli/ls.go
dkoshkin: What are your thoughts on this? Using `list` instead of `ls`?

circle.yml
emmetthitz: That's a great question.

pkg/cli/ls.go
emmetthitz: I don't really feel strongly one way or the other. I think it's more unix-y to have it be `ls`, but if you think it's clearer to use `list`, I can change it and add `ls` as an alias.

dkoshkin: Yeah it makes more sense with the other command verbs.

circle.yml
emmetthitz: It's possible it snuck through when we rebased with the makefile changes. I don't think anything has been merged in since then, and I couldn't get tests to pass without it.

pkg/cli/ls.go
```diff
@@ -0,0 +1,56 @@
...
+
+	cmd := &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "list the names of the clusters currently being managed",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return doList(out, *opts)
+		},
+	}
+	cmd.Flags().BoolVarP(&opts.verbose, "verbose", "v", false, "print the verbose details of the clusters dir.")
```
dkoshkin: Can this be `"print the verbose details"`

dkoshkin:
LGTM otherwise.

Issue #1057, opened 2018-01-17
Provide the ability to add taints to nodes


swade1987:
It would be useful to allow for taints to be added to all nodes (excluding etcd).

**Suggested layout:**

```
taints:
      NoSchedule:
        key: value
        key: value
      NoExecute:  {}
      PreferNoSchedule: {}
```

@dkoshkin @willvrny thoughts?

willvrny:
My only thought is that Kismatic would need to be updated when a new effect is added. The upside of this is that we can validate that only correct effects are used.

swade1987:
I think this method does allow for a more strict set of preflight checks though.

PR #1186, opened 2018-04-03
Add in the plan file the ability to taint nodes

dkoshkin:
Fixes https://github.com/apprenda/kismatic/issues/1057

The taints are a flat list, similar to how it is defined for pods in Kubernetes. This allows the most flexibility when new effects are added. 

Plan file changes:
```
...
worker:
  expected_count: 3
  nodes:
  - host: "worker1"
    ip: "1.2.3.1"
    labels: {}
    taints:
    - key: "foo"
      value: "bar"
      effect: "NoSchedule"

  - host: "worker2"
    ip: "1.2.3.2"
    labels: {}
    taints:
    - key: "xyz"
      value: "zyx"
      effect: "PreferNoSchedule"

  - host: "worker3"
    ip: "1.2.3.3"
    labels: {}
    taints: []
...
```

### DIFF
--- a/ansible/_label-nodes.yaml
+++ b/ansible/_label-nodes.yaml
@@ -1,7 +1,7 @@
 ---
   - hosts: master:worker:ingress:storage
     any_errors_fatal: true
-    name: Label Kubernetes Nodes 
+    name: Label and Taint Kubernetes Nodes 
     serial: "{{ serial_count | default('100%') }}"
     become: yes
     vars_files:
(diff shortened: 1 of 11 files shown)


dkoshkin:
Yeah that comes from the [upstream](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/), my guess is that to be consistent with everything else and provide more flexibility.

swade1987:
@dkoshkin love it!

PR #1193, opened 2018-04-09
Don't verify memory swap on etcd nodes

dkoshkin:


dkoshkin:
Disabling memory swap is required by the kubelet, not by docker or etcd itself.

PR #1213, opened 2018-04-26
Fix typo of node taint configuration

sullivanchan:
Find this issue by deploy with Taints, but find labels are empty, it should be a typo that must be fixed.


dkoshkin:
Thanks!

PR #1230, opened 2018-05-07
add Docker & K8s APT keys before updating APT cache

jashandeep-sohi:
While upgrading a cluster, I ran into an issue where the the pre-flight checks were failing because the K8s PGP key had changed. I was getting stuck at `install apt-transport-https package` task because that also tries to update the APT cache before installing the package.

I was able to resolve the issue by adding the keys before trying to install `apt-transport-https`.


dkoshkin:
LGTM thanks!

PR #1242, opened 2018-05-24
update calico netowrk plugin to 3.1

wenwenxiong:


dkoshkin:
Unfortunately this is not that simple, there is an upgrade process that needs to occur https://docs.projectcalico.org/v3.0/getting-started/kubernetes/upgrade/. 

Calico v3.x does not yet [support route reflectors clustering](https://github.com/projectcalico/calico/issues/1887) which is required for larger clusters. At this point we are not yet ready to upgrade.

